PR #942, opened 2014-08-27
Add var for different space monitoring path (old way)

gmelikov:
If download dir is in another mounted disk, we'll need to write path to it for right space monitoring.
Added new var for path to folder, which is needed to be monitored.


gmelikov:
Novik, i checked it with $topDirectory looking at rutorrent directory, it also worked.
I made new pull request with more elegant way https://github.com/Novik/ruTorrent/pull/989 , please look at it.

Novik:
https://github.com/Novik/ruTorrent/pull/1004

PR #980, opened 2014-12-19
Kill child process before task process.

ArthurJam:
When you killed a task, you first closed the task before killing its children and therefore their were not closed.
Now I act in the other way, by killing the children before.


ArthurJam:
Hi,

Actually the command to kill a task is in 3 steps :
1 Kill the parent process (kill -9 pid)
2 Get the children process from parent (pgrep pid)
3 Kill children process kill -9 `pgrep pid`

But in practice (in my platform), the children are not killed by this processus.
I think you cant pgrep a pid are already killed, so you can't get the children process list (Unless the first kill command takes a certain delay).

So, i propose to inverse the scenario.

Novik:
OK, thanks, i change order of kill requests.

PR #985, opened 2015-01-01
fix again link to dl *.torrent file on cpasbien website

t0mab:


t0mab:
I don't know why cpasbien url for *.torrent files is always changing...

PR #989, opened 2015-01-11
check rtorrent working directory free space in diskspace plugin 

gmelikov:
Before - we check $topDirectory (rutorrent main directory)
After - we check rtorrent working directory free space

This change will display valid data in situations when rtorrent working directory is on another drive.

If there is an option to get rtorrent working directory in plugin on server side - please tell me, it would be more elegant.

### DIFF
--- a/plugins/diskspace/action.php
+++ b/plugins/diskspace/action.php
@@ -1,3 +1,3 @@
 <?php
 	require_once( '../../php/util.php' );
-	cachedEcho('{ "total": '.disk_total_space($topDirectory).', "free": '.disk_free_space($topDirectory).' }',"application/json");
+	cachedEcho('{ "total": '.disk_total_space($_GET["p"]).', "free": '.disk_free_space($_GET["p"]).' }',"application/json");
--- a/plugins/diskspace/init.js
+++ b/plugins/diskspace/init.js
@@ -48,7 +48,7 @@ plugin.init = function()
 				timeout: theWebUI.settings["webui.reqtimeout"],
 			        async : true,
 			        cache: false,
-				url : "plugins/diskspace/action.php",
+				url : "plugins/diskspace/action.php?p="+plugin.folderToScan,
 				dataType : "json",
 				cache: false,
 				success : function(data)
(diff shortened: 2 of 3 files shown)


gmelikov:
This pull request is similar to https://github.com/Novik/ruTorrent/pull/942 , but it doesn't add any variables which need to be modified manually, it's fully automatical.

Novik:
Sorry, but i can't understand - what is you doing in this request.
You pass settings->directory to the client side (BTW, what if rtorrent is not started and this directory is unknown?) after that pass it backward without any usage...

> we check $topDirectory (rutorrent main directory)

$topDirectory is a not "rutorrent main directory". This is a upper top directory, which available for ruTorrent user. By default '/'. If you want to check space at the settings->directory part - simply set $topDirectory to the corresponding value.

gmelikov:
For example:
/dev/sda1 mounted as /
rutorrent directory: /var/www/rutorrent on /dev/sda1
rtorrent download directory: /data/download on /dev/sdb1 (you can get it by rTorrentSettings::get()->directory)
Permissions for these directories are the same (for ex. www-data 0755).

If we check free space on /, it'l give us free space on /dev/sda1.

Yes, we can use $topDirectory=/data/download, but if we use it somewhere else, it would be wrong.

I think it would be more effective to check free space on drive where rtorrent downloads files.

Can you help me - how can i get rTorrentSettings::get()->directory value in plugin's action.php? I understand this pull request with it's get request isn't perfect.

Novik:
First of all, you don't need to use rTorrentSettings::get()->directory value in the plugin's action.php. You should use it in the plugins conf.php. And only in case, if rtorrent is started (otherwise rTorrentSettings::get()->directory will have value '/tmp' or something like that).

gmelikov:
>First of all, you don't need to use rTorrentSettings::get()->directory value in the plugin's action.php. You should use it in the plugins conf.php

It gives me empty result, if i use it in conf.php. ("Fatal error: Class 'rTorrentSettings' not found", if check conf.php through browser directly).

Novik:
https://github.com/Novik/ruTorrent/pull/1004

PR #1004, opened 2015-02-05
plugins/diskspace: possibility to set mount point of a torrents partition

indvd00m:


Novik:
1) You need to use line 
`eval( getPluginConf( 'pluginname' ) );`
instead
`include 'conf.php';`
for including conf.php, because your variand disable per-user plugin configuration.
2) I think, this is unusable thing, check comments to https://github.com/Novik/ruTorrent/pull/942
3) But OK, if we have 2 different pull requests for this, i will add it.

PR #1011, opened 2015-02-20
Erasedata including unpacked files

m-hume:
Adds a new menu item 
Remove and... -> Delete Data (All)
to delete unpacked files in multifile torrents

### DIFF
--- a/plugins/erasedata/fin.sh
+++ b/plugins/erasedata/fin.sh
@@ -5,7 +5,9 @@
 # $3 - hash
 # $4 - base_path
 # $5 - is_multi_file
+# $6 - custom_5 (force delete)
 
 echo "$4" >> "$1"/"$2".tmp
 echo "$5" >> "$1"/"$2".tmp
+echo "$6" >> "$1"/"$2".tmp
 mv -f "$1"/"$2".tmp "$1"/"$3".list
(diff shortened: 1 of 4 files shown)


m-hume:
Sorry one more small change needed
plugins/erasedata/update.php line 21
`if($cnt>2)` needs to change to `if($cnt>3)` as 3 lines of non file location data now exist.

Novik:
https://github.com/Novik/ruTorrent/commit/256ca747e3929a85ab222fac947ba67842baf0c1

ne0ark:
Didn't Delete Data deleted All files in the torrent anyways? Why are there two options now?

m-hume:
Scenario 
You download a torrent that consists of an archive file (or multiple archive files) within a folder.
You have auto extract enabled (or manually extract) to that folder.
You select Remove and... -> Delete Data
Only the original torrents files are deleted i.e. the extracted files persist and the folder remains
Delete Data (All) removes extracted files also (and anything in that folder) also the folder itself.

Note. this will **not** delete extracted files from a torrent containing only a file (i.e. not a file in a folder) as the delete would then occur in the root of your downloads directory removing everything. Not something we wish!

m-hume:
@ne0ark ratio plugin has now been updated
https://github.com/Novik/ruTorrent/pull/1023

PR #1039, opened 2015-04-01
Created piedown plugin

davorb:
Select a torrent in the torrent list, and this plugin will turn the favicon into a piechart that will display the torrent's progress.


davorb:
Actually, wait a bit with merging this in. I found a bug with the latest nightly.

Novik:
What with this?

PR #1050, opened 2015-04-15
Option to delete archive files after successful auto unpacking

xombiemp:
Option to delete archive files after successful auto unpacking.  Set $deleteAutoArchives variable to true in conf.php to enable.  This should only be enable with AutoTools AutoMove enabled with an operation of Copy, Soft Link or Hard Link. Otherwise this will delete the files you are seeding from.  This will only delete archives when auto unpacking, not manual unpacking.

I tried to make the commands in the .sh files as portable as possible.  It should be fully POSIX compatible and is not using any extensions to the 'find' command.

### DIFF
--- a/plugins/unpack/conf.php
+++ b/plugins/unpack/conf.php
@@ -7,4 +7,4 @@ $pathToExternals['unzip'] = '';		// Something like /usr/bin/unzip. If empty, wil
 $pathToExternals['unrar'] = '';		// Something like /usr/bin/unrar. If empty, will be found in PATH.
 
 $cleanupAutoTasks = false;		// Remove autounpack tasks parameters after finish, otherwise will be shown in the 'Tasks' tab
-$deleteAutoArchives = false;	// Delete archive files after successful auto unpack. Do not enable if you are seeding from the complete directory.
+$deleteAutoArchives = false;	// Delete archive files after successful auto unpack. Will not remove archives unless AutoMove is enabled and the operation type is not Move.
--- a/plugins/unpack/conf.php
+++ b/plugins/unpack/conf.php
@@ -7,3 +7,4 @@ $pathToExternals['unzip'] = '';		// Something like /usr/bin/unzip. If empty, wil
 $pathToExternals['unrar'] = '';		// Something like /usr/bin/unrar. If empty, will be found in PATH.
 
 $cleanupAutoTasks = false;		// Remove autounpack tasks parameters after finish, otherwise will be shown in the 'Tasks' tab
+$deleteAutoArchives = false;	// Delete archive files after successful auto unpack. Do not enable if you are seeding from the complete directory.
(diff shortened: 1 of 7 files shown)


xombiemp:
Any chance of this getting merged?  Is there something wrong with it that is preventing it?

Novik:
Sorry, no. Because 
1) This is potentially dangerous. User can use this option with a 'Move' type of operation. As result, he will delete part of torrent.
2) Really, user doesn't need this IMHO. He should use unpack to constant directory with a 'Add Torrent Label/Name' option.

But I will merge this, i you will solve point one.

xombiemp:
Ok, I made it so that the archive deletion will not take place unless AutoMove is enabled and the operation type is not Move.  If you want me to squash the commits into one before you merge let me know.

Novik:
IMHO this is not good variant. You simply need to check before deletion - is this file is a part of live torrent. And do nothing if it is a part.

xombiemp:
So would this check be done in the .sh files? Do you have a suggestion on how to check if a file is part of a live torrent?

Novik:
> So would this check be done in the .sh files?

Sorry, but i can't understand - for which reason you wish to do this in .sh files. 

> Do you have a suggestion on how to check if a file is part of a live torrent?

Parse source .torrent file and try to search archive name in it, for example. 
Anyway, you may do what is you want, see my first comment.

xombiemp:
I'm trying to compare the file paths in the torrent to the file paths of the destination directory.  If they match then I will know that the files are being seeded and to not delete them.  I'm testing with operation set to Move in Autotools and I'm getting a list of files from the torrent using
`$req = new rXMLRPCRequest( 
			new rXMLRPCCommand( "f.multicall", array($hash,"",getCmd("f.get_frozen_path=")) ));`
inside the startSilentTask function in unpack.php.  I get the list of file paths, but they are all set to the download directory /torrent/download, instead of the move to directory /torrent/complete.  I thought maybe I needed to wait for the move to complete first so I added a sleep(20), as a test, before the rXMLRPCRequest but that didn't help.  It seemed to wait to perform the move until that sleep timer expired.  So my question is how can I grab the file list from the torrent, but be sure that it's after the move has taken place?

Novik:
First of all, you can't use rXMLRPCRequest inside startSilentTask. Because in old versions of rtorrent (0.8.6, for example) this will not work. startSilentTask is launched from onFinished handler, i.e. rtorrent is freeze at this point and can't response to your requests.
IMHO you should open corresponding .torrent file from session directory and parse it instead. See plugin retrackers as example.
Second, your should compare relative path, not absolute.

xombiemp:
Parsing the .torrent file would just give me files names.  Those will always match what's in the destination directory, so I'm not sure how that would help.  I need to determine if the archive files in the destination directory are being seeding from.  I tried a new approach in #1057.  Check it out.

PR #1057, opened 2015-04-30
Option to delete archive files after successful auto unpacking.

xombiemp:
Set $deleteAutoArchives variable to true in conf.php to enable. This will only delete the archives if AutoTools AutoMove is enabled with an operation of Copy, Soft Link or Hard Link and when auto unpacking, not manual unpacking.

This uses some logic to determine what type of AutoMove operation was used, if any.  If the download location and complete location are the same then it wasn't moved and it won't delete files.  Otherwise, if the download location doesn't exist anymore then the operation was Move and we won't delete files.  Otherwise, if it's a soft link or hard link it's ok to delete files and the last possibility is Copy and it's ok to delete files.

The rar and zip files are passed to the .sh files and then parsed so that only those files are delete, which ensures there are no accidental deletions.

### DIFF
--- a/plugins/unpack/conf.php
+++ b/plugins/unpack/conf.php
@@ -7,3 +7,6 @@ $pathToExternals['unzip'] = '';		// Something like /usr/bin/unzip. If empty, wil
 $pathToExternals['unrar'] = '';		// Something like /usr/bin/unrar. If empty, will be found in PATH.
 
 $cleanupAutoTasks = false;		// Remove autounpack tasks parameters after finish, otherwise will be shown in the 'Tasks' tab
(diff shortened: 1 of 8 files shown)


xombiemp:
You said I couldn't do a rXMLRPCRequest in unpack.php, but I took that code from the same file a little further down so I'm hoping it's ok.  If not, could you please explain why it won't work there and it does work further down in the same file? Thanks for your suggestions.

xombiemp:
I removed the rXMLRPCRequest in startSilentTask.  Please review this and let me know if it's acceptable.

Novik:
> If not, could you please explain why it won't work there and it does work further down in the same file?

Old versions of rtorrent (<=0.8.6) work with only one thread. And its can't process xmlrpc requests inside torrent hooks like on_finshed.

PR #1058, opened 2015-04-30
Add Peer DL and Peer Downloaded columns in Peers tab

Micdu70:
Peer DL = Download speed of the peer as reported by the peer.
Peer Downloaded = Downloaded amount of the peer, since the beginning of the communication with the peer, as reported by the peer.

Screenshot:


Micdu70:
"Port" column info can be added too with "p.get_port" command, but I think it's not important.
The "Peer Downloaded" info could be wrong, depending of the BitTorrent client.

PR #1061, opened 2015-05-02
Update to detect iOS

sj660:
iOS security blocks AJAX downloads of PDF files via hidden frames like datafrm. By detecting if iOS is running and then changing the target, we can allow PDF files to download in a new tab, allowing direct downloading and reading or importation to iBooks or any other reader.

Another way might be a function to create a new _blank with a direct link to action.php, but this works.

### DIFF
--- a/plugins/data/init.js
+++ b/plugins/data/init.js
@@ -82,11 +82,19 @@ plugin.onLangLoaded = function()
 		if(d && (d.location.href != "about:blank"))
 			try { eval(d.body.textContent ? d.body.textContent : d.body.innerText); } catch(e) {}
 	}));
-	$(document.body).append(
-		$('<form action="plugins/data/action.php" id="getdata" method="get" target="datafrm">'+
-			'<input type="hidden" name="hash" id="datahash" value="">'+
-			'<input type="hidden" name="no" id="datano" value="">'+
-		'</form>').width(0).height(0));
+	var iOS = /(iPad|iPhone|iPod)/g.test(navigator.userAgent);
+ 	if (iOS) {
+        	$(document.body).append(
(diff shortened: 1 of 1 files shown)


JohnFlowerful:
Good find, but do we need all of that duplicate code just to change `target="datafrm"` to `target="_blank"`?

sj660:
Probably. Not detecting iOS creates junk tabs on other systems when you use _blank. There's probably a better way than what I proposed to change that string for iOS, but this works and has been tested on iPad and iPhone. Ideas?

JohnFlowerful:
Perhaps something like this:
```javascript
var iOS = /(iPad|iPhone|iPod)/g.test(navigator.userAgent);
var derp = ""
 if (iOS) {
		derp = "_blank"; }
	else {
		derp = "datafrm"; }
$(document.body).append(
	$('<form action="plugins/data/action.php" id="getdata" method="get">'+
			'<input type="hidden" name="hash" id="datahash" value="">'+
			'<input type="hidden" name="no" id="datano" value="">'+
	'</form>').width(0).height(0));
document.getElementById('getdata').setAttribute('target', derp)
```
Disclaimer: I don't know javascript. This might not be the best way either. It may even be completely wrong.

sj660:
Javascript is not my strongest language either. If Novik prefers something that drops in the script, I'll leave that up to him. I couldn't get anything that dropped in just _blank or datafrm in it, but that's probably just my limits being exposed.

My code works and I'm not familiar with the whole stack going on there to be sure what limits there are on me playing with that string. To net a line or two of code just not to repeat doesn't seem to me like a big deal, especially since I don't think it adds or subtracts much to the performance.

I'm completely agnostic about all of this; I just would like to be able to load PDFs on my iPhone and iPad from this command.

PR #1063, opened 2015-05-02
Use target=_blank to work around iOS security

sj660:
iOS security prevents PDF downloads (just AJAX downloads, I think) going into hidden frames like datafrm. This blocks "get file" of PDFs on iOS. This workaround detects iOS and changes the target to a _blank frame. Tested on iOS 8 and iPhone and iPad.


Novik:
Your code is incorrect, but I understood - what is you want.
https://github.com/Novik/ruTorrent/commit/4f241d82c1ec9baac2d171abf311f9ee71197b8f

Issue #1068, opened 2015-05-09
Cannot import empty feed in RSS plugin


Daiyousei:
When I try to import a valide but empty feed (like this one https://aperture-sc.net/test.rss) I have the error : "Error loading feed : https://aperture-sc.net/test.rss".

I use rutorrent 3.7

After some searchs, I found it's because the plugin search for some "\<entry>\</entry>", but as the feed is empty (for now), there isn't any. The test at line 170 of the file "ruTorrent/plugins/rss/rss.php" fails, returning false, and trigering the "Error loading feed".

a solution could be to returne true if the "$ret" value is equal to 0, but not equal to false.

Could you please fix this ?

Edit : version

Edit 2 : I think the commit 7db9423b86d92d2c4b7f7f72d983c68a0a19579c broke the import of empty feeds. since this commit, we check if there is elements before importing ("!empty($this->items)" line 299 of "ruTorrent/plugins/rss/rss.php")

Daiyousei:
Fixed by pull request #1070

jumperalex:
I may be wrong but this seems to be back again in some form. I could not add an RSS bookmark feed that was empty. 

I removed the "bookmarks=on" portion of the url and I was able to add the feed which then listed the tracker's torrents, not just my autobookmarks. I put "bookmarks=on" back in the feed, it failed again, and removed the feed completely. 

Then I added a bookmarked torrent on my tracker's site. I was then able to add the feed to ruTorrent and it worked as expected by displaying that single bookmarked torrent. 

Finally I removed the torrent from my bookmarks and waited for the feed to update. On next update it popped an "error loading" message but at least it didn't delete the RSS feed. I actually noticed this behavior a while back but didn't much care because my feed was working when there were things to download. This time I had to regenerate my entire ruTorrent install (a docker) and the above all happened.

I'm using 3.7 and the link is in the form of https://revolutiontt.me/rss.php?feed=dl&bookmarks=on&passkey=

Novik:
Try to see the source of feed.

jumperalex:
I'm not sure what you mean. Do you mean view the feed url in my browser? Did that works fine. Or do you mean the source html returned? If so, here is a blank one:

`<?xml version="1.0" encoding="windows-1251" ?>`
`<rss version="2.0">`
`<channel>`
`<title>Revolution</title>`
`<link>https://revolutiontt.me</link>`
`<description>RSS Feeds - RevolutionTT</description>`
[...]

PR #1069, opened 2015-05-10
Update rss.php

Micdu70:
Fix Issue #1068


Novik:
This require presence of simplexml extension. Can we avoid it's usage?

Micdu70:
I don't know..., but my code put many error messages in the web server log for some URL so I will delete this pull request. ;)

PR #1088, opened 2015-07-21
Remove maxlength on loginmgr fields

mallegonian:


mallegonian:
@alexcoe Your change only set the limit to 64 characters. Some trackers have no limit (others have bug reports open). master is still broken, and this PR still fixes it.

PR #1097, opened 2015-08-14
Fix disk space : use download directory by default to calculate free space

Alex131089:


Novik:
For example, rtorrent is not started and rTorrentSettings::get()->directory is undefined. And?

Alex131089:
Right, I'll try to fix that, I guess that if rTorrent isn't running, free space shouldn't be displayed, as we don't know where the destination is ?
Or display the $topDirectory space instead (which is unrelated) ?

Novik:
Don't know. 
Main reason for usage of $topDirectory instead rtorrent download directory now is:
1) We can't show free space when we haven't data from rtorrent
2) But I'm want to see free space in this case, because this is comfortably for me (for example, I like to use filemanager plugin, it doesn't require rtorrent)
Of course we can make a special check (if rtorrent running use one value, otherwise - another), but, IMHO, this is a several strange logic.

Alex131089:
Maybe it's the filemanager plugin that should manage to display free space for other location than the download directory ?
Anyway, I expect to have the freespace of the download directory (which is a NFS mount in my case, so totally unrelated to the local fs) when available, as it's the primary role of the client.
If not available, display nothing or "not available" will help not misinform the user, on the other hand displaying the current fs space can be handy (but can be mistaken) .. so I don't mind.
Another thing is : can ruTorrent be on another host from rTorrent ? If so, "directory" shouldn't be used as disk space cannot be retreived remotly.

Novik:
> Another thing is : can ruTorrent be on another host from rTorrent ? 

Yes, it can.

Alex131089:
What do you think of this one ?
If availlable, display actual disk space of the download directory, else fallback to the old display of $topDirectory. Might be confusing, but no more than currently actually, and should works out-of-the-box for most users.

Alex131089:
Done, didn't mean to rewrite existing code ; the isLocalMode isn't IPv6 compliant though (::1 and/or [::1] should be added).

Novik:
> the isLocalMode isn't IPv6 compliant though

This is configurable via $localhosts array in the ruTorrent/config.php.

Alex131089:
Yup, but the default array https://github.com/Novik/ruTorrent/blob/master/php/util.php#L28 is not complete.

PR #1101, opened 2015-08-16
Update fr.js

Karlingue:
frFR update


Micdu70:
Bad and useless translation update, it was already good...

Frizlab:
I’m French.

The original version was not correct grammatically. However, the replacement does not have the exact same meaning.
The original version meant: “Check for a new version”
The new version means: “Checking for a new version…”
To have the same meaning, it should be: “_Rechercher les mises à jour…_”

As a side note (but it’s the same in English), the ellipsis shouldn’t be made with three points in a row. It should be the character “**…**” (an ellipsis).

PR #1114, opened 2015-09-15
Added russian labels

byte916:
It is now possible to use images for multilingual labels. Added russian
labels and a little english labels.


Novik:
mbstring is a non-default extension. Where you check it's presence?

byte916:
Я добавил проверку)

PR #1118, opened 2015-09-24
Update jquery.js

Machou:
update jQuery to 1.11.3

These releases include a hot-fix for a rare bug in iOS 8.2 and 8.3. This is the only change. As with 1.11.2 and 2.1.3, we do not anticipate any issues when upgrading. However, if you do encounter bugs in upgrading from the previous versions, please let us know.

Maybe go to jQuery to 2.x ?


Novik:
> Maybe go to jQuery to 2.x ?

And what we will do with IE8 in this case?

bhavicp:
Personally I don't think there are enough users on IE8 to justify supporting it.

Machou:
Who use IE8 ? o_O

PR #1138, opened 2015-11-02
Several improvements: config.php renaming & rutracker_check refactoring

hapylestat:
1) conf/config.php was renamed to conf/config.php.sample to avoid replacements of already existed user configs during update of code. As second option, this should force user to rename & edit file first time and tune setting to environment settings

2) rutracker_check refactoring. Move out tracker-specific logic from rutracker_check to separate files. By doing this we achieve:
 - support more easy and  clean adding of new torrent trackers
 - keeping plugin core clean and compact.


JohnFlowerful:
:+1: for the config.php change. You should split the rutracker_check refactoring into a separate PR though.

hapylestat:
@JohnFlowerful ok, will create two separate pull requests

PR #1139, opened 2015-11-03
config.php renaming

hapylestat:
conf/config.php was renamed to conf/config.php.sample to avoid replacements of already existed user configs during update of code. As second option, this should force user to rename & edit file first time and tune setting to environment settings


Novik:
From one side - this is a normal approach. From another - we will break a lot of existing third-party install scripts...

hapylestat:
well, git repository shouldn't be used for installation and distribution. Moreover - master branch, he is always supposed only for dev/testing and not production things(coz master branch have a lot of chances to be unstable). Changes in the master branch should warn users about the change and active third-party will survive. 

Another question is, that they need to be notified via documentation about upcoming changes, for example "What's new" section or another visible place.

As some more soft solution, we can use "deprecation" mechanic. i.e. create both files: config.php and config.php.sample and make notes in the config.php that file will be removed after N releases, please use another one.

What do You think?

Novik:
> well, git repository shouldn't be used for installation and distribution.

In theory - yes. But in practice I have plan to create new release here - https://github.com/Novik/ruTorrent/releases
from a master branch. 

> What do You think?

I think, for usage this (proper) technique we must have an own installation script or something like this. Without it we will have a lot of troubles.

PR #1145, opened 2015-11-06
Adding total torrents size on labels

crazy-max:
Have a quick view of total torrents size by label :

image3


.gitignore
```diff
@@ -26,4 +34,4 @@ plugins/bandwidth
 plugins/plimits
 plugins/localnews
 plugins/hints
-plugins/dmca
+plugins/dmca
```
hapylestat: no new line at the end of the file

hapylestat:
@Micdu70 just synced my instance with this change and....it's possibly to make this feature optional (i.e. possibility to switch on/off in the options)?

crazy-max:
@hapylestat I've not added option to disable this feature.
Not convenient but you can remove `theConverter.bytes(s[lbl], 2)` from `webui.js` on line 1846 and 1853 but i can implement a new webui setting like `webui.show_labelsize` ?

hapylestat:
@crazy-max I can be not only person who doesn't need this info, same like i can't say that i will not need this in the future. I think the best way to add new webui setting, like you proposed.  Thanks!

crazy-max:
@hapylestat Yes that's why i suggest a new webui setting ;)
Btw is there someone to notify for translations?

hapylestat:
@crazy-max idk, you could check by commits, who contributed translations for other plugins, create separate issue and mention them there.....but this is only my suggestion. 

p.s. can help with RU and UA translations, if needed offcource

crazy-max:
I will make a pull request, it's implemented, and i will try to "notify" commiters.

image2

ChriZathens:
I will take care of Greek...

PR #1151, opened 2015-11-10
Unpack: Don't delete files if there was a warning or error

xombiemp:
Added some checks before deleting files that the previous command returned a 0.  For the _dir files it seems to me that the comparative operators for the return values were wrong.  It seems like it should be -ge 1 to assign the $ret value to anything that is not 0.  Let me know if I'm missing something here.

### DIFF
--- a/plugins/unpack/unzip_dir.sh
+++ b/plugins/unpack/unzip_dir.sh
@@ -15,7 +15,7 @@ process_directory()
 	        	mkdir -p "$3"
 			"$1" -o "${fn}" -d "$3"
 			last=$?
-			[ $last -gt 1 ] && ret=$last
+			[ $last -ge 1 ] && ret=$last
 		fi
 	done
 	for fn in "$2"* ; do
@@ -23,14 +23,13 @@ process_directory()
 			name=$(basename "${fn}")
 			process_directory "$1" "${fn}/" "$3${name}/"
 			last=$?
-			[ $last -gt 1 ] && ret=$last
+			[ $last -ge 1 ] && ret=$last
 		fi
 	done
 	return $ret
 }
 
 if [ "$6" != '' ] ; then
-	mkdir -p "$6"
 	process_directory "$1" "$2" "$6"
 	ret=$?
 else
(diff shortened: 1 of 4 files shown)


plugins/unpack/unzip_dir.sh
```diff
@@ -23,14 +23,13 @@ process_directory()
...
 			last=$?
-			[ $last -gt 1 ] && ret=$last
+			[ $last -ge 1 ] && ret=$last
 		fi
 	done
 	return $ret
 }
 
 if [ "$6" != '' ] ; then
-	mkdir -p "$6"
```
xombiemp: This mkdir wasn't necessary because it will do it in the process_directory function

xombiemp:
Any feedback on this?

Novik:
I'm very busy now, sorry. I remember that.

Novik:
Several fragments seems strange. For example
```
[ $ret -le 1 ] && echo 'All OK'
if [ $ret -eq 0 ] && [ "$5" != '' ] ; then
```
I.e. when we have result code 1 we print 'all ok' and... not delete source files.
IMHO, we should choice one of two options here
1) We interpret result code = 1 as minor warning (AFAIR this is correct approach for zip). Or
2) We interpret result code = 1 as error. 
But you has both options at same time...

xombiemp:
I agree that we should choose.  From the unzip man page it says return code 1 means:

```
one  or  more  warning errors were encountered, but processing completed success-  
fully anyway.  This includes zipfiles where one or more files was skipped due to   
unsupported compression method or encryption with an unknown password.
```
So, one example would be if a zip is password protected we would get a return code of 1 and the zip would not be unarchived.  We would not want to delete the archive in this situation.  When checking if we can delete the archive, the only return code we want is 0.

So I think we should interpret 1 as error.

xombiemp:
I made the change to the existing commit so that it only echoes All OK with return code 0
```
[ $ret -eq 0 ] && echo 'All OK'
```
Check it out and let me know what you think.

Issue #1128, opened 2015-10-04
PHP Warning: uisettings.json


bStyler:
Greetings,

One of my ruTorrent clients keeps generating this error:

``` javascript
PHP Warning:  rename(/var/www/rutorrent/share/users/user/settings/uisettings.json.tmp,/var/www/rutorrent/share/users/user/settings/uisettings.json): No such file or directory in /var/www/rutorrent/php/setsettings.php on line 20
```

While I am just filtering my active torrents by DL column or so.

Please advise how I can fix this recurring warning.

Many thanks!

Micdu70:
Hello,

I have this warning sometimes too... I really don't understand why...

PR #1243, opened 2016-04-10
Fix PHP Warning: uisettings.json

GoodOmens83:
This seems to fix the numerous warnings I'm seeing in my error.log:

`PHP Warning:  rename(/var/www/rutorrent/share/users/user/settings/uisettings.json.tmp,/var/www/rutorrent/share/users/user/settings/uisettings.json): No such file or directory in /var/www/rutorrent/php/setsettings.php on line 20`

My guess is sometimes the update happens so fast that sometimes ruTorrent tries to rename a file that has already been renamed

Issue: https://github.com/Novik/ruTorrent/issues/1128

### DIFF
--- a/php/setsettings.php
+++ b/php/setsettings.php
@@ -16,8 +16,10 @@ if(isset($_REQUEST['v']))
 			{
 	        		
 				flock( $fp, LOCK_UN );
-				if( fclose( $fp ) !== false )
-	       				rename( $name.'.tmp', $name );
+				if( fclose( $fp ) !== false ){
+					if (file_exists ($name.'.tmp'))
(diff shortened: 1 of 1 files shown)


Micdu70:
Your pull request doesn't 100% fix this little issue... I think there is no way to avoid that. xD

Novik:
You may simple add '@' character before 'rename'. This will mask this warning. But this will not fix it's reason.

GoodOmens83:
Yea I'm not worried about the reason just OCD about full logs of meaningless text xD

Novik:
I fix this by another way.

PR #1258, opened 2016-04-28
Added access control swapping

osteth:
A simple way to open and close the access configurations.

access-swap.sh swaps out the contents of access.ini with the contents of
access_yes or access_no

This is use full for multi user situations were the admin needs to maintain control over the configuration settings.

### DIFF
--- a/conf/access-swap.sh
+++ b/conf/access-swap.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# This script will quickly and easily swap between two different configurations so that access to configuration
+# options can be easily turned on or off, this can help to protect against unauthorised setting changes.
+
+STR=$(cat access.ini | head -n1| cut -c4)
+GOOD=$(cat access_yes)
+BAD=$(cat access_no)
+
+echo $STR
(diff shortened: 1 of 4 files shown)


Novik:
conf/access.ini
+;; All flags are assumed to be yes by default.
 
 [settings]
+showDownloadsPage = no
+showConnectionPage = no
+showBittorentPage = no
+showAdvancedPage = no

Very strange situation... 
1) Comment and contents are contradictory
2) For which reason you change default settings?

osteth:
oops,  I did not mean to change the defaults. I keep mine closed most of the time to keep other users of my server from accessing the setting and it got copied over. 

I allow multiple friends to use my server to download via an rutorrent pannel.  I use this script to open and close the settings so that I can open them when I need to change them and close them when I am finished so other people don't go in and break things.

Novik:
I understand you reasons, but, please, correct your pull request. Default settings in the access.ini must be  'yes'.

osteth:
ok, the defaults are set back to yes.

PR #1264, opened 2016-05-10
autotools: introduce skip_move on file patterns

isaactolstoi:
The idea was to either auto unpack or auto move the content of a torrent.
It looked easier to test for archives inside a torrent than to check if unpack found nothing and the chain-call auto move, as unpack runs asynchronous. So:
- ui: added list of file patterns separated with ;
- if any pattern matches a file inside the torrent, move is skipped
- fnmatch is used, so shell patterns are possible

### DIFF
--- a/plugins/autotools/autotools.php
+++ b/plugins/autotools/autotools.php
@@ -10,6 +10,7 @@ class rAutoTools
 	public $label_template = "{DIR}";
 	public $enable_move = 0;
 	public $path_to_finished = "";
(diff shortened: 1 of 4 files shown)


Novik:
I.e. all other UI operate with a PCRE patterns, but UI for this feature - with a shell patterns. IMHO this seems strange.

isaactolstoi:
Finally changed the code to match regex instead of file pattern

PR #1273, opened 2016-05-23
Add 6 day interval to default intervals in RSS feed plugin

mariotacke:
I added a 6 day (144h) RSS interval with translations to the RSS plugin. I think users who need weekly refreshes will benefit from this. The built in 1 week interval was too large for me to capture the feed I wanted.

### DIFF
--- a/plugins/rss/init.js
+++ b/plugins/rss/init.js
@@ -1553,6 +1553,7 @@ plugin.onLangLoaded = function()
 					        "<option value='48'>"+theUILang.rssInterval2d+"</option>"+
 					        "<option value='72'>"+theUILang.rssInterval3d+"</option>"+
 					        "<option value='96'>"+theUILang.rssInterval4d+"</option>"+
+						"<option value='144'>"+theUILang.rssInterval6d+"</option>"+
 					        "<option value='168'>"+theUILang.rssInterval1w+"</option>"+
 					        "<option value='336'>"+theUILang.rssInterval2w+"</option>"+
 					        "<option value='504'>"+theUILang.rssInterval3w+"</option>"+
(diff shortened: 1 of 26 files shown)


Novik:
Sorry, but I can't understand reason of this. What about 5 days? 8 days? 11 or 13 days?

mariotacke:
Let me elaborate:
When the RSS matcher matches a specific download, it pauses for the given interval until it matches it again. If a download is released weekly (very common) at lets say Friday 5pm, setting the interval to 7 days will pause the matcher past the release time of the next download => ~5:01pm. Over time, this will push the window forward and eventually exclude certain feeds. By using a value slightly less than a week (6 days), we give ourselves a 24 hour window where a download could be released early or on time or late. The reason I am not using 5 days, is because this gives an additional day to falsely identify a download twice before a new one is released which resets the timer.

An Example:

```
Releases weekly on Fridays around 5pm, 4 day interval scenario
1. Show S01E01, matched at 5pm, pausing for 4 days before matching again
2. Show S01E01 (again), matched at 5:15pm on Tuesday, four days later, pausing for 4 days <= false positive
3. Show S01E02, not matched on Friday at 5:05pm because RSS match is paused, must reset timer

Releases weekly on Fridays around 5pm, 6 day interval scenario
1. Show S01E01, matched at 5pm, pausing for 6 days before matching again
2. Show S01E01 (again), not-matched at 5:15pm on Tuesday, not pausing <= intended behavior
3. Show S01E02, matched on Friday at 5:05pm, pausing 6 days before matching again
```

I hope this makes sense. What it comes down to is that we want to be able to close the window as tightly as possible until the next release all while giving it a buffer of a day or a few hours to identify early releases and exclude late re-releases from before.

Issue #1252, opened 2016-04-18
Add option to display sizes in KiB/MiB/GiB


JamesTheAwesomeDude:
It would be extremely useful if there were an option to display sizes in powers-of-1024 instead of powers-of-1000.

 * `1 KiB` = `1024 B`
 * `1 MiB` = `1024 KiB` (= `1048576 B`)
 * `1 GiB` = `1024 MiB` (= `1073741824 B`)
 * `1 TiB` = `1024 GiB` (= `1099511627776 B`)

Novik:
All sizes are already in MiB etc.

JamesTheAwesomeDude:
If that's the case, could the UI be updated to reflect this?

It's displaying `*B` sans-i currently… So it lists an 860MiB torrent as "860 MB"

Novik:
May be.

bakkra:
Since this commit save as path is truncated, only the base directory is seen not the whole path.

![2016-09-06 14_36_04- 5 3 kib_s rutorrent v3 7](https://cloud.githubusercontent.com/assets/10657324/18272229/64561454-743f-11e6-9ee5-afda5354e83c.png)

Micdu70:
Hello @bakkra, I did this modification because I think it's better to display in this way:
- If torrents are created from a directory => Do not need to display the torrent folder's name for "Save as" info.
  '--> Except if the torrent folder's name torrent's name is different than the torrent's name.
- If torrents are created from a file only => Do not display the filename for "Save as" info. (That was before my commit)

PR #1316, opened 2016-08-31
Add a column with the save path of each torrent

Micdu70:
Request from rutorrent forum: http://forums.rutorrent.org/index.php?topic=6758.0


bakkra:
I think this commit affects the way it's displaying the information, I've posted a screenshot here -> https://github.com/Novik/ruTorrent/issues/1252

Micdu70:
@bakkra It can easily be reverted, with editing the line 2166 in 'js/webui.js', renaming "d.save_path" to "d.base_path". Like uTorrent, you are right. =)

PR #1361, opened 2016-12-07
Snoopy: Add port to Host header

### LINKED ISSUE #1247: Error loading feeds RSS problems with https
Version rtorrent 0.9.6 rutorrent 3.7
cURL can get the right information from HTTPS RSS link.
However, the RSS plugins failed to load RSS link with https.
I can't add torrent link with https download link either.
But the RSS plugins can get the feeds from HTTP RSS link.
I don't know what's wrong with it.
Is there any specific error logs about the RSS plugins so I can check what is the problem?

kaso17:
Fixed #1247/#1196 for me

### DIFF
--- a/php/Snoopy.class.inc
+++ b/php/Snoopy.class.inc
@@ -317,7 +317,7 @@ class Snoopy
 		if(!empty($this->agent))
 			$headers .= "User-Agent: ".$this->agent."\r\n";
 		if(!empty($this->host) && !isset($this->rawheaders['Host']))
-			$headers .= "Host: ".$this->host."\r\n";
+			$headers .= "Host: ".$this->host.":".$this->port."\r\n";
 		if(!empty($this->accept))
 			$headers .= "Accept: ".$this->accept."\r\n";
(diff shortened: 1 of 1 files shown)


Novik:
Thanks, but we have two bad points here.
1) You changed http request, but we also have https.
2) Port number should not be included, when we use default port (80 for http and 443 for https). RFC doesn't require this directly, but most of browsers uses this approach.
Please, correct your pull request.

kaso17:
The latest change should take care of the two points.

Nodens-:
This last commit breaks SSL on RSS feeds for me. Reverting the commit fixes the issue.
Let me know if you need more information for debugging.

Running on Fedora 25 with:
apache 2.4.23
php 7.0.13
curl/libcurl 7.51.0

kaso17:
can you share the feed URL?
If not please let us know the server port and the server software.

kaso17:
Never mind, @gromez already fixed my typo with dac8676322d9a896741059f95f5118a4410e2001

Nodens-:
Indeed, I can verify it is working properly now :)

PR #1378, opened 2016-12-26
Fill out license

jkdf2:
The current license does not have the copyright info filled out. I filled it out the best that I could but feel free to correct it 😃 

Also there is no need to include the full description of GLP Version 3, only need to state that you are using GPL-3.


krichter722:
> Also there is no need to include the full description of GLP Version 3, only need to state that you are using GPL-3.

But it's State of the Art to do that. Leave it there.

jkdf2:
Either way - at least @Novik's copyright information should be added, though.

PR #1408, opened 2017-02-18
Modification of the "Add a torrent" dialog

Evililim:
Create better labels and code reduction in "Add a torrent" dialog".
Delete lang="ru" in _getdir.


Evililim:
Hello, can you delete this  pull request ? 
I have to work on it, then I will post it when finished.

PR #1420, opened 2017-02-28
Plugin Upload ETA

Windkracht8:
A plugin to add upload target and ETA columns to torrent list.
The upload target is a parameter in the settings screen.


enoch85:
@Novik Please merge. Thanks!

PR #1438, opened 2017-04-02
workaround for nginx

rtorrentuser:
New versions of cURL use http/2 by default (e.g. cURL v.7.52 in Debian 9.0), and this can cause problems with certain versions of Nginx (https://github.com/curl/curl/issues/1040). Though it's a Nginx bug, we need a workaround (for example, Rutracker uses Nginx version with this bug). A straightforward solution is to force cURL to always use http/1.0 .

### DIFF
--- a/php/Snoopy.class.inc
+++ b/php/Snoopy.class.inc
@@ -480,7 +480,7 @@ class Snoopy
 		if(!empty($this->user) || !empty($this->pass))	
 			$headers[] = "Authorization: Basic ".base64_encode($this->user.":".$this->pass);
 
-		$cmdline_params = '';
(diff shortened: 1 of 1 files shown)


Novik:
Only on qeuestion - what about *old* versions of curl, before 7.52? Is this flag will be safe for them?

rtorrentuser:
Absolutely safe... Curl changelog says that this option was added in 2001..
---
Fixed in 7.9.1 - November 4 2001
Changes:
...
-0/--http1.0 can now be used to set HTTP 1.0 operations
...
---
https://curl.haxx.se/changes.html

PR #1440, opened 2017-04-07
Added magnet register

Windkracht8:
Use registerProtocolHandler to register magnet links to rutorrent.
This change will make the browser ask the user to use rutorrent to handle magnet links on other pages. The response from addtorrent.php should be different when this happens, I will work on this some other time.

### DIFF
--- a/js/webui.js
+++ b/js/webui.js
@@ -225,8 +225,14 @@ var theWebUI =
 			this.update();
 		}
 		
-		var url = window.location.href.substr(0,window.location.href.lastIndexOf("/"));
-		navigator.registerProtocolHandler("magnet", url + "/php/addtorrent.php?url=%s", "RuTorrent");
+		if(typeof navigator.registerProtocolHandler == 'function'){
+			var url = window.location.href.substr(0,window.location.href.lastIndexOf("/")) + "/php/addtorrent.php?url=%s";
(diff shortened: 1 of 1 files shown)


Novik:
http://caniuse.com/#search=registerProtocolHandler

Windkracht8:
I have added a check if the function is available before using it.

Novik:
Sorry, but this is very strange check...
```
if(typeof navigator.registerProtocolHandler == 'function'){
...
			if((typeof navigator.isProtocolHandlerRegistered != 'function')  // How this may be here ???
}
```

enoch85:
@Windkracht8 Great fix, but this only opens up a new Chrome browser window for me. Is there any way I can hardcode the path, or fix the code so that it opens up ruTorrent at the location I'm using it?

Windkracht8:
@enoch85 I do not think so. The call only excepts protocol, url and title. The rest is up to the browser to do what they think is best.

enoch85:
@Windkracht8 Thanks, will check my browser defaults.

enoch85:
@Windkracht8 Hmm, couldn't find any. :( Could you provide any guidance maybe?

Windkracht8:
@enoch85 I do not think you can change this behaviour, at least not without altering the browser code itself. In FF it will open the link in the same window/tab.

enoch85:
Too bad. :/

PR #1442, opened 2017-04-09
Proxy settings in 'config.php'

rtorrentuser:


Novik:
This is a very strange technique. If you need proxy settings, then you should add them to the Snoopy __construct.

rtorrentuser:
May be...  I'm not a great specialist..)) But it works perfectly for me and it seems to me more obvious and convenient to configure everything configurable in conf file...

PR #1447, opened 2017-04-24
Create NyaaSukebe's search engine

ShinonomeHana:
Create NyaaSukebe's search engine by simply rewriting Nyaa.php


Novik:
What about corresponding icon?

ShinonomeHana:
The site icon of NyaaSukebe is same as Nyaa.
I think we can use the same icon.

PR #1458, opened 2017-05-16
Correct wrong use of correctDirectory, fixes #1072

### LINKED ISSUE #1072: Seeding torrent deleted when tracker edited
Hello,

I currently have an issue with ruTorrent...
When I try do edit a tracker from a torrent i'm seeding (like to change the private key) the files from the torrent are wiped and the download is started again from scratch...

It seems to happens only with torrent added with ruTorrent...

I have a watch folder directly in rTorrent configuration, if I edit the tracker with ruTorrent on it, no issue, files are still here and tracker is well edited... I just need to force a file check and it is OK

The bug seems to happen only on torrents I added with ruTorrent...

I tried to "stop" torrent before editing it, but it wipes the files... when I try to force the check, it stays at "pause" and on the folder, files are empty...

Do you know why ?

happens on 3.6 and 3.7 too...

Using rTorrent: 0.9.4/0.13.4

Thank you

jcharaoui:
If sendTorrent is called with a non-empty $directory parameter, it will always fail because correctDirectory returns a comparison with $topDirectory which is part of the app's URL, and not at all related for the filesystem. This commit fixes the problem so that if a $directory parameter is passed, the function will not fail.

### DIFF
--- a/php/rtorrent.php
+++ b/php/rtorrent.php
@@ -46,8 +46,7 @@ class rTorrent
 			$req = new rXMLRPCRequest();				
 			if($directory && (strlen($directory)>0))
 			{
-				if(!rTorrentSettings::get()->correctDirectory($directory))
-					return(false);
+				rTorrentSettings::get()->correctDirectory($directory);
 				$req->addCommand( new rXMLRPCCommand( 'execute', array('mkdir','-p',$directory) ) );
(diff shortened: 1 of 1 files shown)


Novik:
> $topDirectory which is part of the app's URL, and not at all related for the filesystem.

$topDirectory is NOT a part of the app's URL. $topDirectory has relation to the filesystem.
This is a top directory on the local filesystem, which available to the access from user.
Try to correct your configuration.

PR #1475, opened 2017-06-21
Have check_port bind to the same interface as rtorrent

jcharaoui:
This patch allows the check_port plugin to function correctly when rtorrent is bound to a non-default network interface, such as in scenarios where a VPN is used. Requires the curl PHP module.

### DIFF
--- a/php/settings.php
+++ b/php/settings.php
@@ -23,6 +23,7 @@ class rTorrentSettings
 	public $server = '';
 	public $portRange = '6890-6999';
 	public $port = '6890';
+	public $bind = '0.0.0.0';
 	public $idNotFound = false;
 	public $home = '';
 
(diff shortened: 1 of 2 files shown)


Novik:
Please, don't use php-curl interface in your code. Use Snoopy class. Because ruTorrent has not dependencies from php-curl, only from curl binary.
Of course, now this have mostly historical reasons and, may be, will be changed in future.

Novik:
You may use class variable Snoopy::$IP for that (set it after object construction).

Novik:
And yet one moment - old versions of rtorrent may return empty string, when your send to them 'get_bind' request. Check this in your code, please.

jcharaoui:
OK, I'll look at all this when I can, thanks for the review!

PR #1480, opened 2017-07-05
Peer ID detection update

Micdu70:
- Removed extra whitespace and tab space
- Added detection for the last BitTorrent version (7.10.0) [ Don't know if it will work for 7.10.1 ]
- Added detection for a new android torrent app: Flud [ version >= 0.9.3 use -FLxxxx- as peer id ]
- Added detection for a new android torrent app: tTorrent
- Added detection for a new software: Zona

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -78,7 +78,7 @@ function browserDetect()
 }
 var browser = new browserDetect();
 
-$(document).ready(function() 
+$(document).ready(function()
 {
 	var i = document.createElement('p');
 	i.style.width = '100%';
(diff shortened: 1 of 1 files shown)


Novik:
> Added detection for the last BitTorrent version (7.10.0)

AFAIK, BitTorrent have standard 'mainline-type' peer-id encoding. And it should be catched here:
https://github.com/Novik/ruTorrent/blob/master/js/common.js#L1197

They changed peer encoding? Where I may to see new specification?

Micdu70:
> They changed peer encoding?

The peer ID looks like this for version 7.10.0: **-BT71000%8B...**
Screenshot: [here](http://puu.sh/wFjvT/e9bd5f83a4.png)

> Where I may to see new specification?

Nowhere... or should ask to the BitTorrent devs...

PR #1482, opened 2017-07-05
Fix #1478

### LINKED ISSUE #1478: "setsettings.php" flood request when scrolling in "Files" tab (and others) containing 100 and more lines/rows (IE and Chrome, no issue in Firefox)
Hello,

Everything was said in the title, except that this flood can freeze all other requests and errors like "The request to rTorrent has timed out" pops.

EDIT: Happening when there is 100 and more lines/rows.

Screenshots (Chrome Browser):

During scroll:
setsettings-flood-during-scroll

After some seconds:
setsettings-flood-after

And if too much "setsettings.php" request:
setsettings-flood-timed-out

Micdu70:
Load max 1000 rows instead of 100 in one request should be okay, no ? =)

### DIFF
--- a/js/stable.js
+++ b/js/stable.js
@@ -83,7 +83,7 @@ var dxSTable = function()
 	this.prgEndColor = new RGBackground(".meter-value-end-color");
 	this.mni = 0;
 	this.mxi = 0;
-	this.maxViewRows = 100;
+	this.maxViewRows = 1000;
 }
 
 dxSTable.prototype.setPaletteByURL = function(url) 


Novik:
> Load max 1000 rows instead of 100 in one request

1) maxViewRows is a not size of 'load in one request' portion. This is a count of rows in tables with torrents, files etc. I.e. we may have 1000+ torrents, but corresponding table with torrents may contain only 100 rows.
2) Sorry, but this is not a fix. Proper fix should contain answer to question - 'for which reason we have  flood request in this case'. Instead we from one situation (100+ elements in table) have another (1000+ elements in table).

PR #1488, opened 2017-07-11
Make the text field wider

enoch85:
...to be able to fit 200% and not just 20...

screenshot from 2017-07-11 03 20 35


enoch85:
cc @Windkracht8

Windkracht8:
For me it was already large enough in FF, good change if needed for other browsers.

enoch85:
@Windkracht8 

> for other browsers.

Thanks! Yeah this is on Chromium Ubuntu 17.04. 

cc @Novik

PR #1496, opened 2017-07-18
Add YggTorrent as a new search engine

Arakmar:
Yggtorrent (https://yggtorrent.com) is a new french private tracker which claims to be a T411 successor (former biggest french private tracker).

All meta data are properly parsed.
But the engine currently lacks category filtering support because it's not possible (for now) to use any kind of filters during searches on this tracker.

### DIFF
--- a/plugins/extsearch/engines/YggTorrent.php
+++ b/plugins/extsearch/engines/YggTorrent.php
@@ -1,0 +1,103 @@
+<?php
+
+class YggTorrentEngine extends commonEngine
+{
+    const URL = 'yggtorrent.com';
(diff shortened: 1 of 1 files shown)


Novik:
Thanks, but what about tracker icon?

Micdu70:
The website (and tracker) of yggtorrent.com has no icon set... so I used this one from their twitter account : https://twitter.com/yggtorrent_com

Arakmar:
Yes, I think it's fine to use their Twitter icon, thanks Micdu70

SckyzO:
How to configure this plugin ?

PR #1516, opened 2017-07-31
Fix AutoWatch and AutoMove with add label duplicate path

### LINKED ISSUE #1513: AutoWatch, AutoMove, AutoLabel combined bug
Hello,

So I have AutoWatch enabled which moves a torrent to downloads/{SubDirTree}. Then I have AutoLabel which will set the torrent's label to {SubDirTree).

When the torrent is completed AutoMove will then move based on the Label. The problem is that when the torrents are added by AutoWatch in a sub dir, the torrent will be moved to completed/{SubDirTree}/{SubDirTree} (yes it is repeated). It seems that AutoMove will pick the path in the downloads/{SubDirTree} and also append the label (when that option is enabled).

I can fix it by disabling add label to path but I'd like to add either by auto watch or web UI.

I think by removing the duplicated section in AutoMove path should solve the issue.

barricas:
Fix duplicate path from combining autowatch and automove. If the folder already contains the label in the path don't add it again.

Fixes #1513

### DIFF
--- a/plugins/autotools/move.php
+++ b/plugins/autotools/move.php
@@ -121,7 +121,8 @@ if( $at->enable_move && (@preg_match($at->automove_filter.'u',$label)==1) )
 				{
 					if( $rel_path == './' ) $rel_path = '';
 					$dest_path = rtAddTailSlash( $path_to_finished.$rel_path );
-					if($at->addLabel && ($label!=''))
+					// last condition avoids appending duplicate path from combining folder and label (eg autowatch and autolabel)
+					if($at->addLabel && ($label!='') && strpos($dest_path, $label) === false)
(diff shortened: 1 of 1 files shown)


Novik:
For example, we have label named "/". And?

barricas:
Good observation. Since $dest_path already contains an ending slash ("/"), it shouldn't make a different in not adding another one.

But now it go me thinking that if someone has a label in which it is contained in another part of the path that is not in at the ending there may also be a false positive.

I'm going to think of a better way then.

barricas:
Okay, I added other commit, didn't notice before we already have $rel_path. So I just compare both and if they are not equal then we add the label.

PR #1520, opened 2017-08-04
Fix "Log" tab text display

Micdu70:
- Avoid text overlay


Novik:
1) div.tab is not only for Log tab. It used in many cases of tabs. 
2) For which reason you make this?

Micdu70:
1. Fixed.
2. Example: We want to see "URL info" from extsearch plugin results:

Before


After

PR #1522, opened 2017-08-06
Add welcome message translation

Micdu70:


mariotacke:
Maybe we should use Google translator to pre-fill the new message in all locales? Adding it as is might create more pull requests for individual languages. "started" is simple enough to just translate I think.

PR #1584, opened 2017-11-19
Update yggtorrent loginmgr

Arakmar:
The string for "not logged errors" in the torrent download page has been added somewhere in the homepage so we have to do a better check for that

### DIFF
--- a/plugins/loginmgr/accounts/YggTorrent.php
+++ b/plugins/loginmgr/accounts/YggTorrent.php
@@ -8,7 +8,7 @@ class YggTorrentAccount extends commonAccount
     {
         return (
             strpos($client->results, 'Ces identifiants sont invalides') === false
-            && strpos($client->results, 'Vous devez vous connecter pour télécharger un torrent') === false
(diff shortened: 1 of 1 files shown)


plugins/loginmgr/accounts/YggTorrent.php
```diff
@@ -8,19 +8,19 @@ protected function isOK($client)
     {
         return (
             strpos($client->results, 'Ces identifiants sont invalides') === false
-            && strpos($client->results, 'Vous devez vous connecter pour télécharger un torrent') === false
+            && $client->results !== 'Vous devez vous connecter pour télécharger un torrent'
```
Novik: Result contain this phrase and anything more? Are you sure?

Arakmar: Yes, the URL https://yggtorrent.com/engine/download_torrent?id=<torrent_id> returns exactly this string if not logged (and with a valid torrent id).

Issue #1603, opened 2017-12-14
Labels URI encoding when passing to rTorrent


Hawk84r:
When i stopped using AutoMove because of bug #1602 and switched to .rtorrent.rc method described [here](https://github.com/rakshasa/rtorrent/wiki/Common-Tasks-in-rTorrent#move-completed-torrents-to-a-fixed-location), i discovered that labels with spaces in them (like "some good stuff") would be passed to rtorrent as "some%20good%20stuff", thus moving them to correspondingly named directory. This happens because of encodeURIComponent(lbl) on line 2033 of webui.js.

deed02392:
Yup, same issue here. #metoo

Novik:
This is a normal behavior.
Because we can't send non-UTF characters via xmlrpc. But labels may contains such characters.

Hawk84r:
How is normal space non-UTF? And how is it actually working for me without encoding? This should not be normal behavior.

Novik:
> How is normal space non-UTF?

This is a strange question. Correct question sholud be "For which reason we need to use encoding?"
See comment above for answer.

> And how is it actually working for me without encoding?

Sorry, but I'm not write WebUI specially and only for you. As result, I must to see on the common case.

Hawk84r:
> Sorry, but I'm not write WebUI specially and only for you. As result, I must to see on the common case.
So, i am the only one in the world using spaces, ruTorrent and intended config-like downloads moving? Lucky me.
 
> This is a strange question. Correct question sholud be "For which reason we need to use encoding?"
How about because in this particular case we don't? But yeah, easy way does it.

deed02392:
@Novik The encoding for XMLRPC is not the problem. The problem is we can't use non-URI characters as a label (e.g. / and space) if we set them via ruTorrent. Is this a limitation of the rTorrent XMLRPC then? Because of course you can use non-URI characters if you set them in rTorrent directly.

Novik:
> So, i am the only one in the world using spaces

rtorrent's variable 'custom_1' is not specially intended for storing labels. ruTorrent uses it by such way. And yes, ruTorrent correctly works with a labels with spaces etc.
If you want to use this variable outside ruTorrent - then this is *your* problem. Not mine.

> The problem is we can't use non-URI characters as a label (e.g. / and space) if we set them via ruTorrent. 

Really? See screenshot. Corresponding label contains '/' and space and ruTorrent works with it correctly.
[...]

PR #1608, opened 2017-12-22
Fix double URL encoding of torrent labels

deed02392:
`webui.js` encodes the label text before dispatching the web request, where the `setlabel()` function in the httprpc plugin`init.js` file also does a URL encode of the label text. The end result is the label looks and acts fine in the web-ui, but it is actually setting a URL encoded string of the label in rTorrent. This can cause unexpected behaviour.

### DIFF
--- a/js/webui.js
+++ b/js/webui.js
@@ -2033,7 +2033,7 @@ var theWebUI =
    		for(var k in sr) 
    		{
       			if(sr[k] && (this.torrents[k].label != lbl))
-      				req += ("&hash=" + k + "&s=label&v=" + encodeURIComponent(lbl));
+      				req += ("&hash=" + k + "&s=label&v=" + lbl);
 		}
 		if(req.length>0)
 			this.request("?action=setlabel"+req+"&list=1",[this.addTorrents, this]);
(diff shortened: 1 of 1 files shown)


deed02392:
This PR fixes: #1603

Hawk84r:
Based on conversation from before there is a very slim chance that is happening :(

deed02392:
@Hawk84r Hopefully @Novik will check the code locations I pointed out and agree that URL encoding is performed twice and that is unnecessary.

deed02392:
@Novik Have you reviewed this perchance?

deed02392:
@Novik Any update?

Novik:
1) Yes, we have double encoding in labels now.
2) Your code is also incorrect. Try to think, what we will have in such situation:
- plugins httprpc or rpc are not installed
- labels contains special symbols like '<', '>' etc. 
See here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L661 for details. Label in the prm.value in this case.
3) In any case, what we will do with already existing labels with spaces etc?

PR #1611, opened 2017-12-24
Use PSR-2 style for code

Hyleus:
I've reformatted the code, so it uses PSR-2 as the coding style over the whole codebase. Which is the default style for PHP. It can be found here: http://www.php-fig.org/psr/psr-2/
No behaviour was changed, only the formatting was corrected.


Novik:
Sorry. But I don't like this code formatting.

PR #1684, opened 2018-04-15
Add source field to torrents created externally

itzaname:
The source field was not being set when torrents were created with an external tool.


plugins/create/correct.php
```diff
@@ -68,6 +68,12 @@
 		{
 			$torrent->is_private(true);
 		}
+		if(isset($request['source']))
```
Novik: But who will set this?

PR #1694, opened 2018-05-01
fetching all RSS items !

aminomaboul:
see 
9f2300c


Novik:
Try to RTFM about preg_match_all.
What we have in the zero element, what - in the first etc.

aminomaboul:
loool okey i was wrong, sorry bro

PR #1695, opened 2018-05-04
Update RARbgTorrentAPI.php

Machou:


Novik:
Replacing 'if' operator to the triple operator - is a not good pool request.

PR #1696, opened 2018-05-04
Update Demonoid.php

Machou:
demonoid.me to demonoid.pw


Novik:
Action for www.demonoid.pw, but cookies for www.demonoid.com? Why?

PR #1718, opened 2018-06-20
Replace the custom javascript json_encode with JSON.stringify

anthonyryan1:
JSON.stringify is supported in all browsers updated at any point
in the past 10 years. This allows us to avoid providing a custom
function for encoding json.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -1570,42 +1570,6 @@ function getCRC( str, crc )
 	return(crc);
 }
 
-function json_encode(obj)
-{
-	switch($type(obj))
-	{
-		case "number":
-			return(String(obj));
-		case "boolean":
-			return(obj ? "1" : "0");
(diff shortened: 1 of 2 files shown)


Novik:
With this replacement we will have, for example, false/true values instead 0/1 for booleans. This may have matter in existing code. Are you checked and correct all such cases?

anthonyryan1:
Correct. The only case where we're comparing to an integer is in [webui.js line 1555](https://github.com/Novik/ruTorrent/blob/master/js/webui.js#L1555), and that's a lose comparison.

(1 == true) and (0 == false), so it would continue to work correctly, the only case where anything might break is a === comparison which doesn't occur anywhere settings are used.

I have corrected it for obviousness however, because I think the code is clearer after simplifying it.

PR #1723, opened 2018-06-26
Display number of selected torrent(s) in context menu

Micdu70:
Info: Click on this will deselect all torrents.

How it will display:

### DIFF
--- a/js/webui.js
+++ b/js/webui.js
@@ -1341,12 +1341,23 @@ var theWebUI =
       		}
    	},
 
+        trtDeselect: function() 
+	{
+		var table = this.getTable("trt");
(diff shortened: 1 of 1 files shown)


Novik:
Good idea, but sorry, I have question to the implementation.
Why you calculate selection count instead simple usage of table.selCount ?

Micdu70:
If I remember correctly, in Chrome, it gave me sometimes wrong result like:
2 selected torrents = 4 (table.selCount value)

I try to reproduce this bug now but I can't... so I removed calculate of selection count.

Micdu70:
@ChriZathens thanks!
Don't know about parenthesis... personnaly I like more without.

Issue #1663, opened 2018-03-09
New version of rTorrent api prevents proper identification of listening ports


liaralabs:
The rTorrent dev has pushed a new API (10) that deprecates the command `network.listen.port` and replaces it with `network.port`. If using random port assignment, ruTorrent will simply assume the port opened by ruTorrent is the first port of the range, rather than the actual port itself. This causes ruTorrent to assume the port isn't open and also causes the check_port plugin to show a closed status (because it's checking the wrong port)

My solution was to add a check to php/settings.php for the new API v10 and assign a new port variable if it's found. Everything appears to be functioning correctly, though I haven't tested for backwards compatibility yet.

[network.port.txt](https://github.com/Novik/ruTorrent/files/1798250/network.port.txt)

Novik:
Corresponding code was added, check, please.

liaralabs:
All in all looks good so far; however I am noticing that in Settings > Connection > "Open Listening Port" doesn't seem to be checked, despite you adding the RPC call for it. Perhaps another variable conflict somewhere?

Novik:
I hope fixed, check now please.

liaralabs:
It seems that checkbox still isn't getting ticked -- have restarted php-fpm/rtorrent to ensure that there isn't a caching issue somewhere.

Micdu70:
I think httprpc plugin requires an update... @Novik check my PR

PR #1726, opened 2018-06-26
HTTPRPC plugin update for rtorrent API version 10

Micdu70:
This PR requires review ( #1663 related )

Not tested yet!


Novik:
IMHO this is unusable.
Commands get_port_open, get_port_random and get_port_range already present in the aliases list, see 
https://github.com/Novik/ruTorrent/blob/master/php/settings.php#L162
for details.

PR #1748, opened 2018-08-24
Fix IPTorrents Search

Flupster:


Novik:
Thanks, but sorry, what about errors checking? 
You doesn't check result of call of loadHTML, preg_match, getElementsByTagName etc.
Please, correct this.

Flupster:
Keep in mind getElementsByTagName or getElementById return null if it can't be found

This should never error out in theory, but as IPT change their table layouts in the future it will require changing again, but this is the same for every provider

Novik:
Sorry, but where you use $id_matches?

skiski456:
The error handling commit broke this. 

//if (count($tds) !== 9)
This always returns 1, never 9.

Commenting this line makes it work but obviously doesn't do the intended error handling.

Flupster:
@skiski456 if $tds count is only 1 then $tds[3] for example would be null which is called on later

Unless there is differences in the extensions? what version of php-xml are you running? in testing php7.2-xml package worked fine for me

skiski456:
This explains it:
From php DOMNodeList class changelog:

7.2.0 | The Countable interface is implemented and returns the value of the length property.

skiski456:
danke

PR #1753, opened 2018-09-07
WIP: Add Jackett engine for extsearch

Flupster:
I need information about where to store variables or the best way to provide a custom url/apikey for an search engine?

Temporarily it's located in the plugins/extsearch/conf.php with two variables:

```
$jackett_url = 'http://127.0.0.1:9117';
$jackett_api_key = 'abcdef123456';
```

But if there's a more interactive way of adding it i'm all ears


binhex:
hi @Flupster ive been after jackett support for rutorrent for a while and see youve made good progress with it so im going to try your patch out. 

in relation to your question i think the correct location for the two variables is with the loginmgr plugin, so you would need to create a loginmgr file for jackett by creating a new file under plugins/loginmgr/accounts/ called something like 'Jackett.php' and obviously then modify it to allow the user to enter in api key and url for jackett.

ronajon:
i've tried the code from @Flupster but searching with only Jackett gives no results
any progress on this ?

Novik:
See conversation above.
In the current state PR can't be merged, because it add configuration parameters by incorrect way.

Flupster:
I could probably work on this later this week, I never had time to check the loginmgr and I honestly forgot :laughing:

binhex:
I would love to see this working 😁

Flupster:
Worked on this some today

Is there a "good" way to interact with the loginmgr plugin?

My current way is to use the "login" as the URL and the "password" as the apikey for Jackett

then I added this to the commonEngine class:

```
	public function getAccount($name)
	{
		$accounts = accountManager::load();
		if (array_key_exists($name, $accounts->accounts)) {
			return $accounts->accounts[$name];
		} else {
			return false;
		}
	}
```

If this is not the way to go about it please let me know otherwise I'll finish up tonight or tomorrow.

Novik:
> then I added this to the commonEngine class:

Implementation of this function is incorrect. Because in theory accountManager::load may returns null.
See here - https://github.com/Novik/ruTorrent/blob/master/php/Snoopy.class.inc#L225
for correct way (of course, you should correct parameter for 'obtain' method).

mR12KL:
@Flupster did you get this to work? Jackett integration with ruTorrent would be huge.

PR #1768, opened 2018-10-09
Update YggTorrent tracker URL for 'extsearch' plugin

Micdu70:
_- Use "get_headers" function to obtain the current working URL for 'loginmgr' plugin (tired from these URL modifications...)_ (not good => https://github.com/Novik/ruTorrent/pull/1768#discussion_r224352012)
- Change YggTorrent.php extsearch engine to display the current working URL for .torrent download links

### DIFF
--- a/plugins/extsearch/engines/YggTorrent.php
+++ b/plugins/extsearch/engines/YggTorrent.php
@@ -131,8 +131,11 @@ class YggTorrentEngine extends commonEngine
             );
 
             if ($res) {
+                // Get current URL
+                preg_match('`.+?(?=/torrent)`', $matches["desc"][0], $url);
+
                 for ($i = 0; $i < $res; $i++) {
-                    $link = self::URL . "/engine/download_torrent?id=" . $matches["id"][$i];
+                    $link = $url[0] . "/engine/download_torrent?id=" . $matches["id"][$i];
                     if (!array_key_exists($link, $ret)) {
                         $item = $this->getNewEntry();
                         $item["desc"] = $matches["desc"][$i];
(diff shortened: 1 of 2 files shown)


plugins/loginmgr/accounts/YggTorrent.php
```diff
@@ -4,6 +4,16 @@ class YggTorrentAccount extends commonAccount
 {
     public $url = "https://yggtorrent.to";
 
+    // Check and change URL automatically if necessary
+    public function __construct()
+    {
+        $headers = @get_headers($this->url, 1);
```
Novik: As result we will have HTTP request to the yggtorrent on the each object construction.
This is not acceptable, because we iterate over all objects in the function commonAccount::getAccount.
I.e. we will have request to the yggtorrent on the *each* call of Snoopy::fetchComplex.

plugins/extsearch/engines/YggTorrent.php
```diff
@@ -131,8 +131,11 @@ public function action($what, $cat, &$ret, $limit, $useGlobalCats)
             );
 
             if ($res) {
+                // Get current URL
+                preg_match('`.+?(?=/torrent)`', $matches["desc"][0], $url);
```
Novik: We always have absolute URL in  $matches["desc"][0]? Are you sure?

Novik:
Sorry, but your changes in the loginmgr are unacceptable.

plugins/extsearch/engines/YggTorrent.php
Micdu70: Looks like it is. ;)

plugins/loginmgr/accounts/YggTorrent.php
Micdu70: Hm... okay so I will add the new URL directly. ;)

PR #1769, opened 2018-10-11
Update 'extsearch' plugin

Micdu70:
Change and fix for extsearch results table:

- Fix: JS Error if we want to open context menu² for one (1) already downloaded .torrent from extsearch
- Change: Extsearch table always display the same context menu²
-> Now we can "Open in browser" (in context menu² or double click on torrent) already downloaded .torrent file(s) or get URLs (website page/download link)
-> "Download" in context menu² is disabled if one (1) torrent was already downloaded in the selection

² = left click menu

### DIFF
--- a/plugins/extsearch/init.js
+++ b/plugins/extsearch/init.js
@@ -378,10 +378,11 @@ theWebUI.extTegDelete = function()
 plugin.createExtTegMenu = function(e, id)
 {
 	var trtArray = new Array();
+	var tegArray = new Array();
 	plugin.tegArray = new Array();
 
 	var sr = theWebUI.getTable("teg").rowSel;
-	for(var k in sr) 
+	for(var k in sr)
 	{
 		if(sr[k] == true)
 		{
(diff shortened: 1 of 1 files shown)


Novik:
> Fix: JS Error if we want to open context menu

What about this error? Case?

> ² = left click menu

For do you mean here? For which reason we need context menu functionality on the left click?

>  Now we can "Open in browser" (in context menu² or double click on torrent) already downloaded 

IMHO we don't need this for already downloaded torrents in the most of cases.

Micdu70:
> What about this error? Case?

I tried to understand this problem but I don't...

> For do you mean here? For which reason we need context menu functionality on the left click?

There was already a custom context menu for the extsearch table (teg) but the context menu for torrent table (trt) is used for already downloaded .torrent, to me it's not logical...

> IMHO we don't need this for already downloaded torrents in the most of cases.

It's the extsearch table so all the results should have same features.

Novik:
> I tried to understand this problem but I don't...

You don't understand this problem, but tried to fix it?

> It's the extsearch table so all the results should have same features.

I don't think so.

PR #1770, opened 2018-10-11
Update YggTorrent.php to follow the http 301

mickkael:
Calculate the latest URL based on an old one during the construction, and put this url as the base value of the object. Hope that  the code is nice enough

### DIFF
--- a/plugins/loginmgr/accounts/YggTorrent.php
+++ b/plugins/loginmgr/accounts/YggTorrent.php
@@ -2,8 +2,12 @@
 
 class YggTorrentAccount extends commonAccount
 {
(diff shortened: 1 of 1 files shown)


Novik:
Your code will make call to the yggtorrent on the each usage of Snoopy::fetchComplex (regardless to url value). Because we iterate over all accounts for this call.
Sorry, but this is not acceptable.
See for details:

https://github.com/Novik/ruTorrent/blob/master/plugins/loginmgr/accounts.php#L206
https://github.com/Novik/ruTorrent/blob/master/php/Snoopy.class.inc#L223

mickkael:
Make sense; I'll try another way

PR #1781, opened 2018-10-23
Update YggTorrent.php

KDTheory:
Update Yggtorrent current url


KDTheory:
In order to automate things is it possible to push https://yggtorrent.to instead of https://www6.yggtorrent.to ??

PR #1791, opened 2018-11-06
Update YggTorrent.php

KDTheory:
Skip wwwX.yggtorrent.to for yggtorrent.to


Micdu70:
This is not working, new subdomain 'www9' is required.

PR #1794, opened 2018-11-11
Update YggTorrent.php

KDTheory:
Fix wrong url

### DIFF
--- a/plugins/extsearch/engines/YggTorrent.php
+++ b/plugins/extsearch/engines/YggTorrent.php
@@ -2,7 +2,7 @@
 
 class YggTorrentEngine extends commonEngine
 {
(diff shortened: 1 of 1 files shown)


Micdu70:
Do you really know what are you doing?

KDTheory:
@Micdu70 yggtorrent.to does not work that way so we need to put the exact link in order to make it work. Do you have an other idea ?

Micdu70:
If you get torrents search result from Ygg but you can't download them, changing extsearch engine file is not needed.
To fix download from extsearch, you only need to fix loginmgr account file.

KDTheory:
Oh yeah you right i change the wrong one i close this one

PR #1795, opened 2018-11-11
Update YggTorrent.php

KDTheory:
Fix wrong update and apply exact url

### DIFF
--- a/plugins/loginmgr/accounts/YggTorrent.php
+++ b/plugins/loginmgr/accounts/YggTorrent.php
@@ -2,7 +2,7 @@
 
 class YggTorrentAccount extends commonAccount
 {
-    public $url = "https://yggtorrent.to";
+    public $url = "https://www9.yggtorrent.to";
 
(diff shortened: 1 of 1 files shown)


Micdu70:
subdomain is now "www8"... kill them! xD

Novik:
Please, stop changing url and write a normal code.
For example, you may try to overwrite method commonAccount::test and use reqexp instead comparison. Or something like this.

KDTheory:
Actually i don't know how to do That @Novik maybe @Micdu70 as the skill to perform this changes

Micdu70:
I don't know how you can get/guess current website URL with reqexp only.
I had proposed a fix which uses `get_headers` PHP function, but not "good" => #1768 (commit was removed)
I'm currently using this fix on my server: https://github.com/Micdu70/ruTorrent/commit/05a562cc0788e95ad10b26380e6c1a0319bb06ce 

That's all I can do...

PR #1831, opened 2019-02-03
Yggtorrent change the domain name again

mickkael:


KDTheory:
I just check on their website it’s still the same domain name. 
Careful do not go on another fake website. 
According to the yggtorrent website and their social networks page the correct adress is https://www.yggtorrent.gg/

mickkael:
Well, that's exactly what I have changed. From .to to .gg

KDTheory:
My bad i miss read ^^

Novik:
Is current release (with https://www.yggtorrent.to) works or not?
If yes - then why we need these changes?

mickkael:
For some reason extsearch does not work with .to

Micdu70:
doesn't work with .gg for me too... because of CloudFlare DDoS protection I think...

Alabrir:
Hi,
Indeed it doesn't work.
Someone got an idea maybe ? 
Currently trying to figure it out but nothig happened.
Thanks

Micdu70:
@Alabrir It requires to bypass CloudFlare DDoS protection...

mickkael:
I am thinking about integrating cloudflare-scrape
(Arabic/cloudflare-scrape), like did here:
https://github.com/Aranov/cloudflare-scrape/issues/87
It means we add a dependency with python, bit there is no similar on PHP, and as the cloudflare changes quite often, it is better to rely on a maintained script

mickkael:
@Micdu70 @Alabrir @KDTheory 
I have developed a plugin cloudflare scrape and merged. You can try.

Micdu70:
@mickkael YggTorrent disabled the CloudFlare DDoS protection... so your plugin is "useless" for now... :(
BTW if your plugin works, it's a very good job! 👍

PR #1854, opened 2019-03-30
Use qBittorrent-style torrent filtering

midgetspy:
With many torrents in my rtorrent I find I often can't filter my torrents down how I'd like.

This PR makes the State, Labels, Search, and Trackers sections independent so that each one can have an active selection. The torrent list will be filtered by AND'ing the selection of those 4 sections together. Each section now has an "All" item at the top which is the default selection.

This is the way qBittorrent does its filtering and is much more useful for large libraries.


midgetspy:
@Novik Are you interested in this change of functionality for ruTorrent? If not what if it was an option? I think it's too fundamental to do as a plugin.

Novik:
This is interesting functionality, but it require a lot of changes. If you can make them and will support result - you are welcome.

midgetspy:
I've been testing this for months now and am pretty comfortable that it's working as intended and doesn't have any unwanted side effects. I'm willing to support any problems that come up and I think this would be a good addition to rutorrent @Novik . Anything else I need to do to get it merged?

js/webui.js
```diff
@@ -2125,46 +2127,69 @@ var theWebUI =
...
+	resetLabels: function() {
+		var allLbls = $('.-_-_-all-_-_-');
+		for(var i = 0; i < allLbls.length; i++)
+			this.switchLabel(allLbls.get(i));
+	},
+
 	switchLabel: function(obj)
 	{
-		if(obj.id != this.actLbl)
+		var panelCont = obj.closest(".catpanel_cont");
```
Novik: AFAIK, method 'closest' is not supported by IE11. Can you replace it with a jQuery analog?

Novik:
We have as minimum, 3 plugins, that works with a categories panel - extsearch, tracklabels and rss.
You added changes to the 'tracklabels' plugin only.
What about other plugins? As I may see, they will not work correctly with your changes.
For example, you was delete variable theWebUI.actLbl, but plugins 'extsearch' and 'rss' use it. Etc.

midgetspy:
OK I've addressed your comments and rebased it onto the tip of master.

Novik:
@midgetspy  I added you to collaborators.

PR #1858, opened 2019-04-02
Add context menu item to remove all search tags

midgetspy:
More often I'm done with my searches and want to clear them all than do I need to remove a particular one so I have added "Remove all tags" to the context menu. I wasn't sure what to do with the languages so I pasted the English to all of them.

### DIFF
--- a/lang/no.js
+++ b/lang/no.js
@@ -196,6 +196,7 @@ var theUILang =
  doFastResume			: "Hurtig fortsettelse",
  innerSearch			: "Lokale torrents",
  removeTeg			: "Fjern tagg",
+ removeAllTegs			: "Remove all tags",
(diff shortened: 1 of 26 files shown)


jakobbg:
'removeAllTegs'? :-) Perhaps rename to 'removeAllTags'?

lang/no.js
```diff
@@ -196,6 +196,7 @@ var theUILang =
  doFastResume			: "Hurtig fortsettelse",
  innerSearch			: "Lokale torrents",
  removeTeg			: "Fjern tagg",
+ removeAllTegs			: "Remove all tags",
```
jakobbg: Norwegian translation can be set to 'Fjern alle tagger'

midgetspy:
@jakobbg That’s how the existing code is spelled so I just kept it consistent.

PR #1871, opened 2019-04-07
 unpack plugin: add 7z support

whytedwarf:
I was able to test and confirm that this works.

### DIFF
--- a/plugins/autotools/autotools.php
+++ b/plugins/autotools/autotools.php
@@ -49,7 +49,7 @@ class rAutoTools
 			$this->enable_move = 0;
 			$this->fileop_type = "Move";
 			$this->path_to_finished = "";
-			$this->skip_move_for_files= "/\.rar|\.zip/";
+			$this->skip_move_for_files= "/^[^\/]+(\.rar)|(\.7z)|(\.zip)/";
 			$this->enable_watch = 0;
 			$this->path_to_watch = "";
 			$this->watch_start = 0;
@@ -129,7 +129,7 @@ class rAutoTools
 		$ret .= ", SkipMoveForFiles: '" . addslashes( $this->skip_move_for_files ). "'";
 		$ret .= ", EnableWatch: ".$this->enable_watch;
 		$ret .= ", PathToWatch: '".addslashes( $this->path_to_watch )."'";
-		$ret .= ", MoveFilter: '".addslashes( $this->automove_filter )."'";		
+		$ret .= ", MoveFilter: '".addslashes( $this->automove_filter )."'";
 		$ret .= ", WatchStart: ".$this->watch_start;
 		$ret .= ", AddLabel: ".$this->addLabel;
 		$ret .= ", AddName: ".$this->addName;
--- a/plugins/unpack/7z_dir.sh
+++ b/plugins/unpack/7z_dir.sh
@@ -1,0 +1,58 @@
+#!/bin/sh
+#
+# $1 - unzip
+# $2 - input directory with tail slash
+# $3 - output directory with tail slash
+# $5 - archive files to delete
+# $6 - unpack temp dir
+
+ret=0
+
+process_directory()
+{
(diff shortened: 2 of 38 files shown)


plugins/unpack/7z_dir.sh
```diff
@@ -0,0 +1,58 @@
...
+# $2 - input directory with tail slash
+# $3 - output directory with tail slash
+# $5 - archive files to delete
+# $6 - unpack temp dir
+
+ret=0
+
+process_directory()
+{
+        for fn in "$2"*.zip ; do
```
Novik: May be we should to have .7z extension instead .zip here?

whytedwarf: Yes I must agree I forgot to mention that I had to change that

plugins/unpack/lang/en.js
```diff
@@ -10,6 +10,7 @@
  theUILang.unpackPath		= "Unpack to (leave blank for the torrent's current directory)";
  theUILang.unzipNotFound	= "Unpack plugin: rTorrent user can't access 'unzip' program.";
  theUILang.unrarNotFound	= "Unpack plugin: rTorrent user can't access 'unrar' program.";
+ theUILang.7zNotFound		= "Unpack plugin: rTorrent user can't access '7z' program.";
```
Novik: This is an error. JS identifier can't start with digit.
We should to have theUILang["7zNotFound"] form here.

plugins/unpack/unpack.php
```diff
@@ -433,20 +538,27 @@ public function startTask( $hash, $outPath, $mode, $fileno )
...
 							$rarPresent = true;
 						else
 						if(USE_UNZIP && (preg_match("'.*\.zip$'si", $name)==1))
 							$zipPresent = true;
+						else
+						if(USE_7Z && (preg_match("'.*\.7z$'si", $name)==1))
+							$un7zPresent = true;
 					}
-					$mode = ($rarPresent && $zipPresent) ? 'all' : ($rarPresent ? 'rar' : ($zipPresent ? 'zip' : null));
+					$mode = ($rarPresent && $zipPresent && $un7zPrenset) ? 'all' : ($rarPresent ? 'rar' : ($zipPresent ? 'zip' : ($un7zPresent ? '7z' : null)));
```
Novik: What, if we will have 2 types of archives available in torrent? For example, only zip and rar?
As I may see, in this case only rar files will be unarchived. This is not a correct behavior, IMHO.

whytedwarf: I appreciate you taking the time to look further into this. As I said I just did a quick test and was able to unzip a 7z archive with it by changing the first issue you pointed out. I haven't looked that far into it other than that.

Novik: Sorry, but 'quick tests' are not required.
If you make pull request, then this request must contain correct code. And you should be ready to support this code.

makefu:
I know it's been almost half a year but i would love to see 7z support in rutorrent :+1:

plugins/autotools/autotools.php
```diff
@@ -49,7 +49,7 @@ public function set()
 			$this->enable_move = 0;
 			$this->fileop_type = "Move";
 			$this->path_to_finished = "";
-			$this->skip_move_for_files= "/\.rar|\.zip/";
+			$this->skip_move_for_files= "/^[^\/]+(\.rar)|(\.7z)|(\.zip)/";
```
Micdu70: False positive (e.g.): precious.and.rare.book.pdf
Regex should check if filename ends by ".7z" , ".rar" (and forgotten variants like .r01,.r02,... and .001,.002,...) or ".zip".
So here is my solution:
`$this->skip_move_for_files = "/(^[^\/]+\.(7z|rar|r\d\d|\d\d\d|zip))$/";`

makefu:
still relevant for me!

PR #1888, opened 2019-05-07
Update YGG tracker logo

Micdu70:


Novik:
As I may see, you delete image from extsearch and add two images with a same name to tracklabels. Why?

PR #1892, opened 2019-05-24
fix sorting order

Micdu70:
Use JavaScript String `localeCompare()` Method to sort by AlphaNumeric order.

### DIFF
--- a/js/stable.js
+++ b/js/stable.js
@@ -730,6 +730,7 @@ dxSTable.prototype.Sort = function(e)
 		case TYPE_PEERS :
 		case TYPE_SEEDS :
 			d.sort(function(x, y) { return self.sortPeers(x, y); });
+			break;
       		default : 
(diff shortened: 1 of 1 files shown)


js/stable.js
```diff
@@ -808,7 +809,7 @@ var theSort =
 		if(y==null) y = "";
 		var a = (x + "").toLowerCase();
 		var b = (y + "").toLowerCase();
-		return((a < b) ? -1 : (a > b) ? 1 : 0);
+		return((a.localeCompare(b) < b.localeCompare(a)) ? -1 : (a.localeCompare(b) > b.localeCompare(a)) ? 1 : 0);
```
Novik: Why you need such strange constuction? Try to read about function localeCompare, please. About its return value.

Micdu70: Oh yeah my bad... ;)

Issue #1895, opened 2019-06-03
$XMLRPCMountPoint have no effect


SSoft7:
Hi,

Here is my config.php. As you can see I am using SCGI port to connect to rtorrent. Here rutorrent is working with or without the variable **$XMLRPCMountPoint** mentioned.
is the variable valid anymore? I have also tried removing RPC mount from webserver config but it still works.
The only thing that have working effect is the **$scgi_port** and **$scgi_host** variable.

Am i doing anything wrong?

```
<?php
  @define('HTTP_USER_AGENT', 'Mozilla/5.0 (Windows NT 6.0; WOW64; rv:12.0) Gecko/20100101 Firefox/12.0', true);
  @define('HTTP_TIME_OUT', 30, true);
  @define('HTTP_USE_GZIP', true, true);
  $httpIP = null;
  @define('RPC_TIME_OUT', 5, true);
  @define('LOG_RPC_CALLS', false, true);
  @define('LOG_RPC_FAULTS', true, true);
  @define('PHP_USE_GZIP', false, true);
  @define('PHP_GZIP_LEVEL', 2, true);
  $schedule_rand = 10;
  $do_diagnostic = true;
  $log_file = '/tmp/errors.log';
  $saveUploadedTorrents = true;
  $overwriteUploadedTorrents = false;
  $topDirectory = '/home/testuser/';
  $forbidUserSettings = false;
  $scgi_port = 42513;
  $scgi_host = "localhost";
  $XMLRPCMountPoint = "/sdsdsdsRPC2s";
$pathToExternals = array(
"php" => '/usr/bin/php', // Something like /usr/bin/php. If empty, will be found in PATH.
"curl" => '/usr/bin/curl', // Something like /usr/bin/curl. If empty, will be found in PATH.
"gzip" => '/bin/gzip', // Something like /usr/bin/gzip. If empty, will be found in PATH.
"id" => '/usr/bin/id', // Something like /usr/bin/id. If empty, will be found in PATH.
"stat" => '/usr/bin/stat', // Something like /usr/bin/stat. If empty, will be found in PATH.
"pgrep" => '/usr/bin/pgrep', // Something like /usr/bin/pgrep. If empty, will be found in PATH.
"python" => '/usr/bin/python', // Something like /usr/bin/pgrep. If empty, will be found in PATH.
);
  $localhosts = array("127.0.0.1", "localhost",);
  $profilePath = '../share';
  $profileMask = 0777;
  $autodlPort = 23139;
[...]

PR #1896, opened 2019-06-04
Allow "$XMLRPCMountPoint" variable to be optional in config.php

Micdu70:
#1895
- Remove useless "/RPC2" default value for XMLRPCMountPoint (`js/common.js`) => Already set in `conf/config.php`
- RPC and HTTPRPC plugins are disabled by default (`conf/plugins.ini`)
- New comment line for $XMLRPCMountPoint variable (`conf/config.php`)
- Do not get PHP Notice log messages if $XMLRPCMountPoint variable is not set (`php/getplugins.php`)

### DIFF
--- a/conf/config.php
+++ b/conf/config.php
@@ -44,7 +44,7 @@
 	// $scgi_port = 0;
 	// $scgi_host = "unix:///tmp/rpc.socket";
 
-	$XMLRPCMountPoint = "/RPC2";		// DO NOT DELETE THIS LINE!!! DO NOT COMMENT THIS LINE!!!
+	$XMLRPCMountPoint = "/RPC2";		// If you enable RPC plugin -OR- HTTPRPC plugin in "plugins.ini", you can delete or comment this line.
 
 	$pathToExternals = array(
 		"php" 	=> '',			// Something like /usr/bin/php. If empty, will be found in PATH.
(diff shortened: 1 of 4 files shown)


Novik:
> Remove useless "/RPC2" default value for XMLRPCMountPoint (js/common.js)

I.e. without any rpc-type plugins here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L183 we will have an empty variable?
And here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L1170 an ajax call to the empty url?

What is a purpose of this PR? Why we need it at all?

Micdu70:
> I.e. without any rpc-type plugins here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L183 we will have an empty variable?

`conf/config.php` still have "/RPC2" by default.

> And here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L1170 an ajax call to the empty url?

If empty, Ajax is calling the current browser url.

> What is a purpose of this PR? Why we need it at all?

Having the possibility to remove / not set $XMLRPCMountPoint variable in `config.php` when enabling a rpc-type plugin / not setting a SCGI mount point on the webserver.

Novik:
> Having the possibility to remove / not set $XMLRPCMountPoint variable in config.php when enabling a rpc-type plugin / not setting a SCGI mount point on the webserver.

I think, we don't need this.

PR #1898, opened 2019-06-14
screenshots: Wait for ffmpeg processes to finish

nyuszika7h:
Without this, when generating 1080p PNG screenshots some of the images appear as broken as it tries to load them before they're generated.


Novik:
This can't be a reason.
Because request to the every image is made only after the call of concrete ffmpeg was finished.

PR #1912, opened 2019-07-17
Check if 'rtorrent.dat' cache file exists/readable before using it

### LINKED ISSUE #1911: Caching issue with XMLRPC
Hi,

I am using ruTorrent with rtorrent (connected using Unix socket).
Sometimes I get the following error when trying to open the 'Settings' menu ( even though Unix socket is working properly.)

> Bad response from server: (500 [error,getsettings]) Warning: XMLRPC call is failed.

This problem can be fixed by clearing the browser cache and reloading the rutorrent page. **But I need to find the exact file which is causing this issue.**

## Additional Info
Browser: Chrome Version 75.0.3770.100 (Official Build) (64-bit)
ruTorrent: v3.9 (release version)

1. Request
```
curl 'http://rtorrenthost.com/rutorrent/plugins/httprpc/action.php' \
-H 'Authorization: Basic Z2lnYXVzZXIxOjEyMzQ1Ng==' \
-H 'Origin: http://rtorrenthost.com' \
-H 'Accept-Encoding: gzip, deflate' \
-H 'Accept-Language: en-US,en;q=0.9' \
-H 'Content-Type: application/x-www-form-urlencoded' \
-H 'Accept: application/json, text/javascript, */*; q=0.01' \
-H 'Referer: http://rtorrenthost.com/rutorrent/' \
-H 'X-Requested-With: XMLHttpRequest' \
-H 'Cookie: SID=v60SlwbaFSrEL2vQRbaIqj8oMPiwXzJ2; PHPSESSID=1tkcl0psmc5ufffpjlb23rjd9h;' \
-H 'Connection: keep-alive' \
-H 'DNT: 1' \
--data 'mode=stg' \
--compressed --insecure
```

2. Response
`Warning: XMLRPC call is failed.`

3. Content of /tmp/errors.log (These are the exact errors right after the above request)
https://pastebin.com/raw/DqCb5JhV
So, If someone can point out which file is causing this, I can disable caching for that on the webserver.

**UPDATE:  TL;dr - I believe rTorrentSettings::get() is a little buggy, it does not check if cache file exists and tries to load it which is the cause of this issue. That's all.**

Micdu70:
Close #1911

### DIFF
--- a/php/getplugins.php
+++ b/php/getplugins.php
@@ -242,7 +242,7 @@ if($handle = opendir('../plugins'))
 	else
 	{
 		$remoteRequests = array();
-		$theSettings = rTorrentSettings::get(true);
+		$theSettings = rTorrentSettings::get();
 		if(!$theSettings->linkExist)
 		{
 			$jResult.="noty(theUILang.badLinkTorTorrent,'error');";
--- a/php/initplugins.php
+++ b/php/initplugins.php
@@ -128,7 +128,7 @@ $tmp = getTempDirectory();
 if($tmp!='/tmp/')
 	makeDirectory($tmp);
 
-$theSettings = rTorrentSettings::get(true);
+$theSettings = rTorrentSettings::get();
 if( $theSettings->linkExist && ($handle = opendir('../plugins')))
 {
 	$plg = getConfFile('plugins.ini');
--- a/php/settings.php
+++ b/php/settings.php
@@ -127,18 +127,18 @@ class rTorrentSettings
 		$cache = new rCache();
 		return($cache->set($this));
 	}
-	static public function get( $create = false )
+	static public function get()
 	{
 		if(is_null(self::$theSettings))
 		{
 			self::$theSettings = new rTorrentSettings();
-			if($create)
-				self::$theSettings->obtain();
-			else
+			if(@file_get_contents(getSettingsPath().$name.'/'.self::$theSettings->hash))
 			{
 				$cache = new rCache();
 				$cache->get(self::$theSettings);
 			}
+			else
+				self::$theSettings->obtain();
 		}
 		return(self::$theSettings);
 	}


Novik:
Reason of this?

Micdu70:
The only reason is if the "share" folder has wrong permissions ; ruTorrent can still be loaded ( and we can even add a warning message about permissions issue, maybe? ).

Micdu70:
But... yeah... there is no "real" reason for this, there is already an error message if "settings" folder is not readable by webserver user. So I will close this...

Novik:
If share folder has wrong permission, then we must generate an error.
We can't retrieve all properties from rtorrent on the each call to ruTorrent (they is used in a lot of places), this is not good for performance.

Novik:
#1911 has relation to the wrong permission of share folder, right? If yes - then why user can't see an error message? He is a blind, or we have an error in diagnostic?

Micdu70:
@Novik: This user removed the rtorrent.dat file and I still don't really understand why he did that... He tried to find a "bug" when it's normally impossible to reproduce...

Novik:
But if he delete this file, then ruTorrent must to create it again when user reload interface (on getplugins.php call). Why this wasn't happening?
Or he delete this file every time? >:()

Micdu70:
With some "weird" PHP caching, it doesn't recreate the cache file. I really don't know how it is possible.

SSoft7:
> But if he delete this file, then ruTorrent must to create it again when user reload interface (on getplugins.php call). Why this wasn't happening?
> Or he delete this file every time? >:()
Hello,
just wanted to clarify, this is not a permission issue, its a cache issue, I think I have a proper explanation for this.
image
my web server is configured to cache the mime-type `application/javascript` so getplugins.php is flagged as javascript type and and it does not refresh with a normal reload.

Novik:
I understood, thanks.
With such aggressive caching you will have undefined behavior, when
1) you will change plugins set at runtime
2) you will use some plugins (several from them pass current state in the js variables)
3) you will make manual changes in the user's share directory.

SSoft7:
Thanks for explaining this, I was really interested in knowing what's going wrong. I can see a lot of php files are giving js output so I will just exclude PHP extension from caching.

PR #1931, opened 2019-08-05
Add Hostname plugin: adds hostname column to peers tab, resolve ip to hostname

rozhuk-im:
Old version and screenshot: http://netlab.dhis.org/wiki/ru:software:rtorrent:rutorrent_hostname

### DIFF
--- a/plugins/hostname/README
+++ b/plugins/hostname/README
@@ -1,0 +1,59 @@
+HostName Plugin for RUTorrent
+==========================
+
+The purpose of this plugin is to show hostnames of peers for selected torrent.
+Based on GeoIP plugin.
+
+Показывает имена хостов на вкладке с пирами.
+Основан на коде плагина GeoIP.
+
+
+
+Принцип работы
+--------------
+
+Запросы на преобразование имёт посылаются асинхронно на сервер rTorrent,
+преоборазованием занимается PHP скрипт.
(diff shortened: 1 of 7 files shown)


Novik:
Plugin GeoIP already has this functionality.

rozhuk-im:
It more heavy and require more deps to work.

Novik:
> It more heavy

JS code of GeoIP plugin - 6.1K, of your plugin - 4.8K. 
IMHO 1.3K of difference is not "more heavy".

> and require more deps to work.

This is not true. In minimal configuration (only $retrieveHost = true) plugin GeoIP has only one dependence - php extension 'json'. Your plugin has this dependence too.

Sorry, but I don't need your plugin in the ruTorrent repo.

rozhuk-im:
GeoIP replace IP address to hostname, this add host name, so user can see both IP and HOSTNAME.

Micdu70:
@rozhuk-im Can you please create a new git repo with this plugin only?
(name's example: hostname-rutorrent)
I will update the ruTorrent wiki after. ;)

rozhuk-im:
Ok, I need rework readme.txt to readme.md in this case :)

rozhuk-im:
https://github.com/rozhuk-im/rutorrent-hostname

PR #1933, opened 2019-08-08
Update 'geoip' plugin

Micdu70:
- 'ctype' PHP extension may be not installed by default (See Micdu70/geoip2-rutorrent#2)


Novik:
IMHO we must to use ```is_numeric``` instead ```ctype_digit```.

Micdu70:
I agree, PR updated. =)

PR #1955, opened 2019-10-16
stat plugin fix when csv is empty or incomplete

aureliendavid:
Hi, 

This is a small fix for cases I've encountered where the stats csv file is empty or incomplete. 

In these cases `fgetcsv()` returns `null`. Then when the class member is used there's an error. 

The fix just keeps the default value for the member in this case.

### DIFF
--- a/plugins/trafic/stat.php
+++ b/plugins/trafic/stat.php
@@ -21,16 +21,26 @@ class rStat
 		$this->fname = getSettingsPath().'/trafic/'.$prefix;
 		if($file=@fopen($this->fname,"r"))
 		{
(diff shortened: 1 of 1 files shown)


Novik:
https://github.com/Novik/ruTorrent/pull/1954 - ???

aureliendavid:
sorry I closed the previous one just before you merged it - I made a mistake in the commit message/id so I reopened this one with a better commit

aureliendavid:
if you can git reset the previous + git push -f and then merge this one it's better

if it's a hassle you can just close this one and keep the other

sorry about the mix-up

PR #1960, opened 2019-10-25
Update style.css

### LINKED ISSUE #1959: Oblivion theme: Add torrent window  broken
I have two problems with Oblivion theme
1. Just noticed that in Chrome, Add torrent menu displays wrong
Capture
In IE or Firefox it is almost OK (a bit trimmed at the bottom, please notice button "Add URL")
Capture11

2. Even in IE or Firefox which have almost no problem, when I switch the language to Greek, a lot of the window is trimmed
Capture2
(Notice that no Add URL option is visible at all, it is all trimmed)
The Greek translation is longer than the English one, thus a line break is added and the window itself does not stretch, so a part of the bottom stuff is trimmed
This only happens to Oblivion theme, however.. a couple of other themes don't pose this issue
Do you believe it can be fixed?
Cheers!

Micdu70:
Fix #1959


Novik:
We have an error with Oblivion theme, but you changed common styles instead styles of theme. This is a strange for me...
Do you check other existing themes? Is all correct with them after your changes?

Micdu70:
What's the need of fixing width and height values? Browser should and must size this window itselft, no?

PR #1991, opened 2020-01-19
extsearch plugin: update JPopsuki

koblack:


plugins/extsearch/engines/JPopsuki.php
```diff
@@ -13,16 +14,17 @@ class jpopsukiEngine extends commonEngine
...
 		$added = 0;
-		$url = 'http://jpopsuki.eu';
+		$url = 'https://jpopsuki.eu';
 		if($useGlobalCats)
-			$categories = array( 'all'=>'0', 
-				'tv'=>"&filter_cat[5]=1&filter_cat[6]=1&filter_cat[7]=1", 
-				'pictures'=>"&filter_cat[9]=1", 
+			$categories = array(
+				'all'=>'',
+				'music'=>''
```
Novik: Why global category 'tv' is absent? It present in local categories. Why we have same 'all' and 'music' global categories?

koblack: Yes you are right, fixed.

PR #1994, opened 2020-01-21
For IOS compatibility

Ozneb:
I remove the filter ",application/x-bittorrent" to avoid issue on IOS (iphone, ipad, ...).
The issue is : we can't add files because all files are disabled in the IOS Browser.
I just let ".torrent" in the filter.

So, it's work on IOS and other systems.

### DIFF
--- a/js/content.js
+++ b/js/content.js
@@ -105,7 +105,7 @@ function makeContent()
 				'</span>'+
 				'<label>'+theUILang.Label+':</label><input type="text" id="tadd_label" name="tadd_label" class="TextboxLarge" /><select id="tadd_label_select"></select><br/>'+
 				'<hr/>'+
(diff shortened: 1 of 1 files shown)


Novik:
Are you sure? On my iPad with iOS 12.4.4 I can see all files ('accept' attribute of control isn't checked by OS) in corresponding dialog and they are not disabled or something like that.

Ozneb:
I'm not able to add torrent files from my iphone 6 (IOS 13.3) and my iPad Pro (IPadOS 13), all files are gray.
I've just modify the accept attribute in the input type file from  ".torrent,application/x-bittorrent" to ".torrent" and now it works.

I'm not the only one, there is other guy with the same problem (see here : https://www.reddit.com/r/seedboxes/comments/d98wsz/rutorrent_web_client_and_ios_131/)

It's probably a recent issue due to new update 13 OS.

PR #2038, opened 2020-05-03
Fix ruTracker extsearch engine

Micdu70:
Currently, we can't retrieve torrents date from returned results.


Micdu70:
Info: Currently can't retrieve torrent date info in code result... :'(

PR #2050, opened 2020-05-14
Update check_port plugin

Micdu70:


Novik:
Why we need this?

Micdu70:
Because it doesn't return results anymore... (get "File not found" from portchecker website)
And I added PHP server IP value if we can't read "bind" value form rtorrent or it is set up to 0.0.0.0

Novik:
In common case this will not helps.
For example, in my case $_SERVER['SERVER_ADDR'] is 192.168.1.1...
We need to know default interface. For that, if we have empty $bind, then we must retrieve url https://portchecker.co/, parse result for 'data-ip' value and use it for secondary request to portchecker.co. 
And on normal way we must to do this (retrieve our ip) on the plugin initialization only. And store it in cache up to the next plugin initialization.

mikl0s:
portchecker.co changed their website the current implementation doesn't work anymore - however this works just fine:

```@$client->fetch("https://canyouseeme.org", "POST", "application/x-www-form-urlencoded", "port=".$port);
if($client->status==200)
{
        if(strpos($client->results,"Error")!==false)
                $ret = 1;
        else
        if(strpos($client->results,"Success")!==false)
                $ret = 2;
}```

Micdu70:
> In common case this will not helps.
> For example, in my case $_SERVER['SERVER_ADDR'] is 192.168.1.1...
> We need to know default interface. For that, if we have empty $bind, then we must retrieve url https://portchecker.co/, parse result for 'data-ip' value and use it for secondary request to portchecker.co.
> And on normal way we must to do this (retrieve our ip) on the plugin initialization only. And store it in cache up to the next plugin initialization.

Done. ;)

sir-hashpipe:
> portchecker.co changed their website the current implementation doesn't work anymore - however this works just fine:
> 
> ```
> @$client->fetch("https://canyouseeme.org", "POST", "application/x-www-form-urlencoded", "port=".$port);
> if($client->status==200)
> {
>         if(strpos($client->results,"Error")!==false)
>                 $ret = 1;
>         else
>         if(strpos($client->results,"Success")!==false)
>                 $ret = 2;
> }
> ```

This worked great for me thanks.
When trying the committed code I get an error from session start line 50, no clue how to solve that after trying a couple of hours, looks like some PHP related stuff:

2020/09/29 02:14:09 [error] 3152#3152: *1 FastCGI sent in stderr: "PHP message: PHP Warning:  session_start(): Write of lock failed in /var/www/rutorrent/plugins/check_port/action.php on line 50
PHP message: PHP Warning:  session_start(): Unable to clear session lock record in /var/www/rutorrent/plugins/check_port/action.php on line 50" while reading response header from upstream, client: xxx.xxx.xxx.xxx, server: xxxxxxx.xxx, request: "POST /rutorrent/plugins/check_port/action.php?init HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.0-fpm.sock:", host: "xxxxx.xxx", referrer: "https://xxxxxx.xxx/"
2020/09/29 02:14:09 [error] 3152#3152: *1 FastCGI sent in stderr: "PHP message: PHP Warning:  Unknown: Failed to write session data (memcached). Please verify that the current setting of session.save_path is correct (unix:/tmp/memcached.sock) in Unknown on line 0" while reading upstream, client: xxx.xxx.xxx.xxx, server: xxxxx.xxx, request: "POST /rutorrent/plugins/check_port/action.php?init HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.0-fpm.sock:", host: "xxxxx.xxx", referrer: "https://xxxxx.xxx/"

PR #2052, opened 2020-05-16
Add support for recaptcha 3rd party in CloudScraper

mickkael:


Novik:
1) If you need configuration, then you should use plugin's config file, not global one.
2) Use something like "if($recaptcha_solving_enabled)" instead "if(isset($cloudscraper_recaptcha))". $cloudscraper_recaptcha must be defined in the plugin's conf.php always.

PR #2064, opened 2020-05-31
check_port take global ip from settings, if set

UAReMMeR:
check_port take global ip from settings, if set

### DIFF
--- a/php/settings.php
+++ b/php/settings.php
@@ -27,6 +27,7 @@ class rTorrentSettings
 	public $idNotFound = false;
 	public $home = '';
 	public $tz = null;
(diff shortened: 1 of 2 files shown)


Novik:
"get_ip" ias a alias for "network.local_address".
Why we must check at external service our *local* address?

UAReMMeR:
I have in my config
# The ip address reported to the tracker.
ip = x.x.x.x

Is it wrong now?

My IP, reported to the tracker, is not my nat IP.
I use port tunneling to remote server.

So In my case to check the port i use rpc request get_ip and the open port checked right.

PR #2092, opened 2020-08-18
Add 'mTeam' extsearch engine

Micdu70:


Micdu70:
@Novik Hello, I have a question: This private tracker requires to have a "referer" url link (with at least its domain name) to be able to download .torrent files.
I found how to configure that for the 'extsearch' plugin but for "php/addtorrent.php" I don't know how to do that (except to modify addtorrent.php file itself => https://github.com/Micdu70/ruTorrent/commit/dc8175ccf94ef013bca76027be9f9cf71206c5b4 [line 77 to 82]), any better solution?

Novik:
1) What about magnet links? Maybe we may use them here instead direct links to .torrent?
2) You may try to use loginmgr for that. Add download url pattern to the 'test' function, override 'fetch' function and implement corresponded logic in it. But this is not simple.

Micdu70:
1. No magnet link for this private tracker.
2. Found a simple solution but I don't understand why it was removed => #2118

Novik:
2. If this helps, then you may add to commonAccount something like that - 
```
protected function loadData( $client = null )
{
    return(privateData::load( $this->getName(), $client ));
}
```
and overwrite this function in your engine with loading refrer from cache.

Novik:
Or you may delete comment in commonAccount and try to solve problem with ZamundaNet by other way - on your wish.

PR #2106, opened 2020-09-24
update config.php with default executables

### LINKED ISSUE #1380: rss and rutracker_check plugins don't work + upload fail
Hello,

I checked:
+ ruTorrent Wiki
+ open and close issues
+ archlinux rutorrent [wiki](https://wiki.archlinux.org/index.php/RuTorrent)

And I can't solve this two problems:

```
rss: Some functionality will be unavailable. Webserver user can't access external program (curl).
rutracker_check: Plugin will not work. Webserver user must have execute access to the rtorrent session directory (/home/sdbox/.session/).
```

I made a manual install and [here](http://rawsec.ml/en/archlinux-install-rtorrent-rutorrent/) is the full install process I wrote (you'll have all info on the blog post).

### LINKED ISSUE #1635: Plugin will not load [rutorrent v3.8]
When I try to enable any plugin it either switches between:

(if Extern == '/usr/bin/mktorrent')
`create: Plugin will not work. rTorrent user can't access external program (php).`
or (if Extern == '')
`create: Plugin will not work. Webserver user can't access external program (mktorrent).`

Depending on whenever or not I've added (in this case create plugin) anything in the Extern variable in the config.php file.

Things I've tried so far:

- chmod the entire rutorrent dir with a+rwx
- tested test.sh with no errors
- tested running said command as that user (and it worked)
- checked that $PATH has said path to program needed
- checked that rtorrent user can write to rutorrent/share (it can)

All users (rtorrent, php & www) share each others group.

### LINKED ISSUE #1900: can't access pgrep/python
Hello

My rutorrent installation seems to be unable to access some programs. I was able to fix curl and php via the conf/config.php (even though they were in $PATH but i still get the following error message:

`[25.06.2019 16:03:15] WebUI started.
[25.06.2019 16:03:15] _task: Plugin will not work. Webserver user can't access external program (pgrep).
[25.06.2019 16:03:15] _cloudflare: Plugin will not work. Webserver user can't access external program (python).
[25.06.2019 16:03:15] create: Plugin will not work. It requires plugin(s) _task
[25.06.2019 16:03:15] unpack: Plugin will not work. It requires plugin(s) _task
[25.06.2019 16:03:15] mediainfo: Plugin will not work. It requires plugin(s) _task
[25.06.2019 16:03:15] screenshots: Plugin will not work. It requires plugin(s) _task
[25.06.2019 16:03:15] spectrogram: Plugin will not work. It requires plugin(s) _task`

`# sudo -u www-data pgrep
pgrep: no matching criteria specified
Try `pgrep --help' for more information.`

seems to work, any advice would be apreciated

### LINKED ISSUE #1967: rtorrent user unable to access external programs
Ok so for the past couple of days ruTorrent seems to be having issues communicating with rTorrent, although I don't know if the problem is with rTorrent or ruTorrent. Existing torrents are seeding fine but no plugins work, so autotools which I use to automate my torrent grabbing isn't working. If I reboot my server everything seems to start working fine again and autotools grabs the torrents but if I check back an hour or so later then I have the below errors in my log.

Any idea what the problem might be?

```
[18.11.2019 19:24:29] WebUI started.
[18.11.2019 19:24:30] rTorrent user can't access 'id' program. Some functionality will be unavailable.
[18.11.2019 19:24:30] screenshots: Plugin will not work. rTorrent user can't access external program (ffmpeg).
[18.11.2019 19:24:30] unpack: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] trafic: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] create: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] ratio: Some functionality will be unavailable. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] rss: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] datadir: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] autotools: Plugin will not work. rTorrent user can't access external program (php).
[18.11.2019 19:24:30] rss: Some functionality will be unavailable. rTorrent user can't access external program (curl).
[18.11.2019 19:24:30] mediainfo: Plugin will not work. rTorrent user can't access external program (mediainfo).
[18.11.2019 19:24:30] _task: Some functionality will be unavailable. rTorrent user can't access external program (pgrep).
```

piramiday:
this change would allow plenty of instances to work out of the box.
since `PATH` by default is not set, leaving those blank strings does not result in the executables being found, anyway.

close #1380
close #1635
close #1900
close #1967

### DIFF
--- a/conf/config.php
+++ b/conf/config.php
@@ -46,12 +46,16 @@
 
 	$XMLRPCMountPoint = "/RPC2";		// DO NOT DELETE THIS LINE!!! DO NOT COMMENT THIS LINE!!!
 
+	// Set default path for each executable.
+	// Since by default PATH is not set, an empty exe would not be found.
 	$pathToExternals = array(
-		"php" 	=> '',			// Something like /usr/bin/php. If empty, will be found in PATH.
-		"curl"	=> '',			// Something like /usr/bin/curl. If empty, will be found in PATH.
-		"gzip"	=> '',			// Something like /usr/bin/gzip. If empty, will be found in PATH.
-		"id"	=> '',			// Something like /usr/bin/id. If empty, will be found in PATH.
-		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
+		"php"	=> '/usr/bin/php',
+		"curl"	=> '/usr/bin/curl',
+		"gzip"	=> '/bin/gzip',
+		"id"	=> '/usr/bin/id',
+		"stat"	=> '/usr/bin/stat',
+		"python"=> '/usr/bin/python',
+		"pgrep"	=> '/usr/bin/pgrep',
 	);
 
 	$localhosts = array( 			// list of local interfaces
--- a/conf/config.php
+++ b/conf/config.php
@@ -46,16 +46,14 @@
 
 	$XMLRPCMountPoint = "/RPC2";		// DO NOT DELETE THIS LINE!!! DO NOT COMMENT THIS LINE!!!
 
-	// Set default path for each executable.
-	// Since by default PATH is not set, an empty exe would not be found.
 	$pathToExternals = array(
-		"php"	=> '/usr/bin/php',
-		"curl"	=> '/usr/bin/curl',
-		"gzip"	=> '/bin/gzip',
-		"id"	=> '/usr/bin/id',
-		"stat"	=> '/usr/bin/stat',
-		"python"=> '/usr/bin/python',
-		"pgrep"	=> '/usr/bin/pgrep',
+		"php" 	=> '',			// Something like /usr/bin/php. If empty, will be found in PATH.
+		"curl"	=> '',			// Something like /usr/bin/curl. If empty, will be found in PATH.
+		"gzip"	=> '',			// Something like /usr/bin/gzip. If empty, will be found in PATH.
+		"id"	=> '',			// Something like /usr/bin/id. If empty, will be found in PATH.
+		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
+		"python"=> '',			// Something like /usr/bin/python. If empty, will be found in PATH.
+		"pgrep"	=> '',			// Something like /usr/bin/pgrep. If empty, will be found in PATH.
 	);
 
 	$localhosts = array( 			// list of local interfaces


Novik:
> since PATH by default is not set

IMHO this is not true and depends from the concrete system and web-server (if we talk about web-server user).

> this change would allow plenty of instances to work out of the box.

.. and, may be, will broke a lot of existing installations.

piramiday:
a few weeks ago I installed ruTorrent on a desktop Linux Mint system: the default installation did not work because the PATH was not set, so I had to manually configure each executable.
recently I installed it on a new system with Ubuntu 20.04.1 Server -- and the same happens!
I then felt confident that this was the case for anyone who did not explicitly modify the executable paths, so that prompted me to try and open this PR.

but you are right, this is a poor modification and might break things.
I think the best compromise of a solution is this one:

> Proper fix would _probably_ be looking in `/usr/bin` if
> 
>     * path to tool is not set
> 
>     * `PATH` env variable is not set
> 
> 
> that should make it work out of the box on newer systems
>
> @ https://github.com/Novik/ruTorrent/issues/1900#issuecomment-626225296

where instead of looking always in `/usr/bin`, you look instead in the common defaults (e.g. `gzip` is not in `/usr/bin`).

I won't be able to help out in this case, though, since I wouldn't know where to put the fix without breaking the code.
feel free to push to my branch and propose a change!
I think that will be really helpful for all first-time users that expect most things to work out of the box.

I will push a change reverting my previous commit but adding the two lines regarding `pgrep` and `python`, which are useful to show to the user that those two executables, too, need exact paths.

Novik:
> recently I installed it on a new system with Ubuntu 20.04.1 Server -- and the same happens!

This is a strange. What about your web-browser?

> but adding the two lines regarding pgrep and python, which are useful to show to the user that those two executables, too, need exact paths.

Sorry, but python and pgrep are not used by ruTorrent core.
I.e. these paths must be specified (and yes, they are already specified) in the configuration of plugins, where these binaries are really used. 
See here - 
https://github.com/Novik/ruTorrent/blob/master/plugins/_cloudflare/conf.php#L5
https://github.com/Novik/ruTorrent/blob/master/plugins/_task/conf.php#L7

piramiday:
> This is a strange. What about your web-browser?

what do you mean? I don't have a browser on the server.
I use the same Linux Mint desktop to connect to the server via WebUI.

> Sorry, but python and pgrep are not used by ruTorrent core.

oh, I see... then I'll close this botched PR.
I would still suggest you to put some fallback mechanism in the code, so as to make vanilla cases (e.g. Ubuntu Server) work out of the box.

Novik:
> what do you mean? I don't have a browser on the server.

Sorry, must be "What about your web-server?", of course ;)

piramiday:
oh, yeah, that's right! on Ubuntu Server 20.04.1 I did, roughly:
```
sudo apt install nginx php7.4-fpm
```
everything works fine, but ruTorrent complains about pretty much any non-defined PATH.

Novik:
AFAIR in this case you may set path (and other environment variables, if you need them) in the variable ```env``` inside  php-fpm.conf. I.e. something like

```env[PATH] = /usr/local/bin:/usr/bin:/bin```

And yes, in this case PATH really may be empty without additional configuration.

piramiday:
yes, exactly, I knew that from https://github.com/Novik/ruTorrent/issues/1900#issuecomment-626225296 ... that's why I was proposing these changes.

what are your thoughts on adding a fallback mechanism for when the PATH is empty?

PR #2118, opened 2020-11-11
Fix adding torrent function

Micdu70:
Is there a valid reason to comment `$client->referer = $rt->referer;` in `plugins/loginmgr/accounts.php` (line 24)?
Because it's the easiest way to fix this => https://github.com/Novik/ruTorrent/pull/2092#issuecomment-675395900

### DIFF
--- a/plugins/loginmgr/accounts.php
+++ b/plugins/loginmgr/accounts.php
@@ -77,6 +77,11 @@ abstract class commonAccount
 		return( stripos($url,$this->url)===0 );
 	}
 
+	protected function loadData( $client = null )
+	{
+		 return(privateData::load( $this->getName(), $client ));
+	}
+
 	protected function updateCached($client,&$url,&$method,&$content_type,&$body)
 	{
 		return(true);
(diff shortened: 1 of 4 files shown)


Novik:
Yes. Some search engines require concrete referer for processing. For example, for ZamundaNet this referer must include parameters of current search. But if we load referer from cache then authorization request will not contain this referer.

See this commit for details
https://github.com/Novik/ruTorrent/commit/cfad7fcd8f977b021f9a6fc000425bed67b2a064

Micdu70:
Done! Thanks for your help @Novik 👍

plugins/loginmgr/accounts.php
```diff
@@ -112,9 +117,9 @@ public function check( $client, $login, $password, $auto )
 		$modified = privateData::getModified($this->getName());
 		if( ($modified===false) || ((time()-$modified)>=$auto))
 		{
-			$data = privateData::load( $this->getName() );
-			if($this->login($client,$login,$password,$url,$method,$content_type,$body,$is_result_fetched) && 
-				$client->status>=200 && 
+			$data = $this->loadData($client);
```
Novik, line 120: IMHO must be

```suggestion
			$data = $this->loadData();
```

Micdu70: OK ;)

PR #2140, opened 2021-02-01
Disallow passing XML strings with `execute` to rTorrent

Wouter0100:
We have gotten reported a case were a customer was able to pass specifically-crafted rTorrent XML data containing execute commands, resulting in creating a reverse shell and exposing the environment.

This fixes that vulnerability.

### DIFF
--- a/plugins/httprpc/action.php
+++ b/plugins/httprpc/action.php
@@ -458,7 +458,7 @@ switch($mode)
 	}
 	default:
 	{
-		if(isset($HTTP_RAW_POST_DATA))
+		if(isset($HTTP_RAW_POST_DATA) && strpos(preg_replace('/\s+/S', '', $HTTP_RAW_POST_DATA), "execute") === false)
 		{
 			$result = rXMLRPCRequest::send($HTTP_RAW_POST_DATA,false);
 			if(!empty($result))


Novik:
This is not a solution.
1) We have a lot of dangerous statements. "file.append", "import" etc.
2) Hacker may use various techniques for masking input, for example, usage of &#NN sequences instead normal characters.
3) For avoid security hole you should use this fork of rtorrent - https://github.com/Novik/rtorrent

Wouter0100:
Thanks for your response. The main reason why I approached it like this, is that I can see a similar statement seems also used over [here](https://github.com/Novik/ruTorrent/blob/f2d14d6df9e11a639d6c6deab0dbdb7b69f9771b/plugins/httprpc/action.php#L24). Or, is that statement there for a different reason?

Out of curiosity, what does your fork of rtorrent do differently?

Novik:
> Or, is that statement there for a different reason?

No, this is result of early attempts to fix a problem. This technique does not work (see reasons above).

> Out of curiosity, what does your fork of rtorrent do differently?

ruTorrent add a special flag for untrusted (i.e. all non-local) RPC connections:
https://github.com/Novik/ruTorrent/blob/master/php/xmlrpc.php#L106

rtorrent check this flag and try to filter incoming commands. 

UPD. And, as I see now, my fork is not solve this problem too. Even with it attacker may use dangerous commands inside various multicall methods. Will think.

Issue #2130, opened 2020-12-14
Extsearch plugin for YggTorrent with _cloudflare plugin returns "Fetch Error"


flvlef:
Hi, after having fixed some _cloudflare plugin warnings (install cloudscraper module and point to the correct python3 path in  `$pathToExternals['python']`), I'm now running all that without any warning in ruTorrent logs.

When I'm searching YggTorrent `returns` a Fetch Error no matter what.

Any idea what to check?

PR #2141, opened 2021-02-11
Missing parameter on YggTorrent extsearch.

hmatthieu:
YggTorrent returns a 301 if `attempt=1` GET parameter is missing. Note that 301 should be correctly handled by the extsearch/engines.php web client (https://github.com/Novik/ruTorrent/blob/master/plugins/extsearch/engines.php#L52).

This commit does not fix web client, it only adds `attempt` parameter to avoid the 301 and make YggTorrent external search works as expected.

### DIFF
--- a/plugins/extsearch/engines/YggTorrent.php
+++ b/plugins/extsearch/engines/YggTorrent.php
@@ -88,7 +88,7 @@ class YggTorrentEngine extends commonEngine
         $what = rawurlencode(rawurldecode($what));
 
         // Initial search to retrieve the page count
-        $search = self::URL . '/engine/search/?name=' . $what . $catParameters . '&do=search';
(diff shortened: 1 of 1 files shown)


hmatthieu:
#2130

hmatthieu:
@Novik Partial fix, YggTorrent randomly responds with a 301 that increase attempt parameter. Seems that following the 301 is mandatory to make it work properly.

hmatthieu:
Sooo, I did some code reading/debugging. The rutorrent web client is actually correctly following redirections (https://github.com/Novik/ruTorrent/blob/master/php/Snoopy.class.inc#L303), the issues comes from a redirection loop from YggTorrent (which leads to a max depth hit https://github.com/Novik/ruTorrent/blob/master/php/Snoopy.class.inc#L306). I'll check further in the next days.

belette75:
hi,
same problem also with yggtorrent.re

PR #2236, opened 2021-12-25
Upgrade jquery version to 3.6

stickz:
This pull request upgrades the jquery version to 3.6.  There is a noticeable improvement in page loading speeds.

It also upgrades `jquery.filedrop` to the latest version. This fixes a bunch of bugs and makes it compatible with jquery version 3.6.


Novik:
This is not such easy.
For example, only one moment:

> Deprecation Notice: The jqXHR.success(), jqXHR.error(), and jqXHR.complete() callbacks are removed as of jQuery 3.0. You can use jqXHR.done(), jqXHR.fail(), and jqXHR.always() instead.

And what we have in the js/rtorrent.js?
For correct migration we should use jquery-migrate-plugin. Check this - 
https://jquery.com/upgrade-guide/3.0/#jquery-migrate-plugin
https://github.com/jquery/jquery-migrate/#README

stickz:
I'm still working on this jquery upgrade. It's going to take a few days to fully complete. I am using the `jquery-migrate-plugin`.

js/rtorrent.js
```diff
@@ -694,7 +694,7 @@ rTorrentStub.prototype.getValue = function(values,i)
 		var el = value.childNodes[0];
 		while(!el.tagName)
 			el = el.childNodes[0];
-		ret = $type(el.textContent) ? $.trim(el.textContent) : 
+		ret = $type(el.textContent) ? String.prototype.trim(el.textContent) :
```
Novik, line 697: Try to use normal form (here and below). I.e. el.textContent.trim() istead String.prototype.trim(el.textContent).

stickz:
@Novik This pull request is ready for review again. I removed all the warnings from jquery migrate and resolved all the combability issues with JQuery version 3.6. Page loading speeds are slightly faster now which is great.

There is still some depreciated syntax used because it's not easily replaceable. It's not recommended, but it still works perfectly fine. An example of this is [jQuery.type()](https://api.jquery.com/jQuery.type/#jQuery-type-obj), which is often used as `$type()` in the plugin. `typeof` doesn't work in every situation.

I came across some issues with error handling, while I was thoroughly testing this upgrade. The dbody container (which stores all the torrents) became null. This was due an issue with plugins not having adequate permissions `canChange` in `plugins.ini`. I had to clean up the chain reaction of errors. They were interfering with my ability to test this JQuery upgrade.

PR #2237, opened 2021-12-26
Optimize 'tracklabels' plugin

stickz:
This pull request optimizes the fetching of images for the tracklabels plugin. 

It leverages web browser cookies to reduce web server load and increase page loading speeds. It checks with your web browser to see if a request for a special image has failed within the last 30 days. If so, it won't ask again unless you clear cookies.

It also uses `sendFile()` in `~/php/util.php` to cache the generic image sent, when a label or tracker doesn't have a special image. This is significantly faster for page refreshes than performing a 302 URL redirection to the location of the generic image.

When no special image is available, these requests now complete in ping + 30ms for a fresh load or 60ms for a cached load. Before this patch, there were instances of these requests taking 500ms to 800ms to complete.

### DIFF
--- a/plugins/tracklabels/action.php
+++ b/plugins/tracklabels/action.php
@@ -7,28 +7,35 @@ ignore_user_abort(true);
 set_time_limit(0);
 
 if(isset($_REQUEST["label"]))
-{
-	$label = function_exists('mb_strtolower')
-		? mb_strtolower(rawurldecode($_REQUEST["label"]), 'utf-8')
-		: strtolower(rawurldecode($_REQUEST["label"]));
-	$name = getSettingsPath().'/labels';
-	if(!is_dir($name))
-		makeDirectory($name);
-	$name.=('/'.$label.".png");
-	if(is_readable($name))
-	{
-		sendFile( $name, "image/png" );
-		exit;
-	}
-	$name = dirname(__FILE__)."/labels/".$label.".png";
-	if(is_readable($name))
-	{
-		sendFile( $name, "image/png" );
-		exit;
+{	
+	$request = rawurldecode($_REQUEST["label"]);
+	$cookie = str_replace(' ', '', $request);
+	if (!isset($_COOKIE[$cookie]))
+	{	
+		$label = function_exists('mb_strtolower')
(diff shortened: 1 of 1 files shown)


Novik:
1) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

>A cookie-name can contain any US-ASCII characters except for: the control character, space, or a tab. It also must not contain a separator characters like the following: ( ) < > @ , ; : \ " / [ ] ? = { }.

"Any US-ASCII characters". But label may contain non-US-ASCII characters.

2) For case (!isset($_REQUEST["label"] && isset($_REQUEST["tracker"]) && isset($_COOKIE[$_REQUEST["tracker"]]) we always will have /trackers/unknown.png as result. Why?

stickz:
1. I plan on improving the cookies to work for separator characters by encoding the cookie name in url format. I will also add a regex expression to detect if the label is not in US-ASCII format. This performance improvement should still benefit most people who use ruTorrent. I don't think it's worth discarding this change for a handful of other use cases.

2. The way issets are used here is very confusing. They do work as intended and will not always return `unknown.png`. I will add a few code comments, when I submit anther commit to this pull request. These will help to clarify how the issets code is used.

Novik:
> I plan on improving the cookies to work for separator characters by encoding the cookie name in url format. 

We already have encoded value in the $_REQUEST["label"]. But for unknown reason you construct cookie name from it after decoding.

> They do work as intended and will not always return unknown.png

I can't see how this may be... BTW, what we will have for case isset($_REQUEST["label"]) && isset($_COOKIE[$cookie])? As I may see, also unknown.png.
Sorry, but I can't understand, how this may work.

stickz:
I'm closing this pull request because it's not compatible with #2247. I will open a new pull request to optimize the track labels plugin after #2247 is dealt with. It will not include cookies. Just the image caching instead of 302 redirection improvement.

Refactoring the PHP code significantly reduced the amount of time required to complete this task. The code refactor is a better solution than cookies. It dropped the time required to complete this task down to the approximate network transfer time.

PR #2238, opened 2021-12-28
Improve file path lookups

stickz:
This pull request improves the internal method of looking up file paths on ruTorrent. 

**Some highlights include:**

1. A file path cache is created which stores the location of all the binaries in a .dat file. This cache is persistent in between page refreshes. It improves the overall page loading speed and reduces web server load (especially with multiple instances running).

2. An improved method of looking up the location of binaries on Linux is implemented. The `which` command is sent via php `shell_exec`. This has performance advantages. It also works in situations where the path lookup method will not work.

Most Linux distributions come with this package already installed. The ones that don't can easily add it. It might be worth adding a note about this package on the ruTorrent wiki.

### DIFF
--- a/conf/config.php
+++ b/conf/config.php
@@ -22,9 +22,21 @@
 	// for php
 	@define('PHP_USE_GZIP', false, false);
 	@define('PHP_GZIP_LEVEL', 2, true);
+	
 
-	$schedule_rand = 10;			// rand for schedulers start, +0..X seconds
-
+class RTorrentConfig
+{	
+	public $schedule_rand = 10;	// rand for schedulers start, +0..X seconds
+	
+	public $pathToExternals = array(
+		"php" 	=> '',			// Something like /usr/bin/php. If empty, will be found in PATH.
+		"curl"	=> '',			// Something like /usr/bin/curl. If empty, will be found in PATH.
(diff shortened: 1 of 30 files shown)


Novik:
Now, if I want to add path to the binary in the new plugin I may write in the plugin's init.php something like this:
```
if(empty($pathToExternals['tar']))
{
	$pathToExternals['tar'] = '';	// Something like /usr/bin/tar. If empty, will be founded in PATH.
}
```
How I may to do this after your changes?

stickz:
> Now, if I want to add path to the binary in the new plugin I may write in the plugin's conf.php something like this:
> 
> ```
> if(empty($pathToExternals['tar']))
> {
> 	$pathToExternals['tar'] = '/usr/local/bin/tar';	// If empty, will be found in PATH.
> }
> ```
> 
> How I may to do this after your changes?

I will add a static function to the ExternalPath class to add new file paths. Since this is a wrapper, you won't have to call empty anymore. It will work something like this: `ExternalPath::addFilePath('tar', '/usr/local/bin/tar');`

Also, please don't merge this pull request yet. There is a problem with the `RemoteRequests` class. I didn't realize it until after because I was working on a different branch. It's disabling plugins when the path to externals like PHP is successfully found.

PR #2239, opened 2021-12-29
Asynchronous WebUI Settings

stickz:
This pull request moves loading the JSON file for WebUI settings from PHP to Javascript. This allows dispatching of an alternate thread to perform this task, so we don't have to wait for the plugins to finish loading. This task can be done asynchronously without any negative impacts. All we need to do is request the file location from PHP. The settings being loaded here are just basic UI settings, which can be updated on the web server at any time. They have absolutely no dependencies.

Performance wise, this reduces loading times by approximately 300ms on average. In extremely bad cases, it may take seconds off the loading time. It also reduces web server load because we're not streaming the file contents anymore to send them. JQuery simply downloads the file to the local computer then creates a JSON object out of it.  The file is only 3.2kb, so it's done very quick.


php/util.php
```diff
@@ -393,6 +393,32 @@ function getUploadsPath( $user = null )
 	return( getProfilePath($user).'/torrents' );
 }
 
+function getWebUIJsonFile()
+{
+	global $profilePath;
+	$ret.= isset($profilePath) ? $profilePath : '../share';
+	$ret = str_replace('..', '', $ret);
+	
+	if(is_null($user))
```
Novik, line 402: How $user may be not null here?

php/util.php
```diff
@@ -393,6 +393,32 @@ function getUploadsPath( $user = null )
 	return( getProfilePath($user).'/torrents' );
 }
 
+function getWebUIJsonFile()
+{
+	global $profilePath;
+	$ret.= isset($profilePath) ? $profilePath : '../share';
+	$ret = str_replace('..', '', $ret);
+
```
Novik, line 401: Really? What we will have for "../../../blah-blah-blah", for example?

php/getsettings.php
```diff
@@ -2,19 +2,6 @@
...
-		$s = @file_get_contents($fname);
-		if($s==false)
-			$s = '{}';
-		flock($fo, LOCK_UN); 
-	}
-	fclose($fo); 
-}
-
-cachedEcho($s,"application/json",true);
+$fname = getWebUIJsonFile()."/uisettings.json";
```
Novik, line 5: What about race conditions (when we have reading and writing of this file at the same time), how you will prevent them?

Novik:
> Performance wise, this reduces loading times by approximately 300ms on average.

Please, don't repair non-broken things.

php/getsettings.php
stickz: I will prevent writing to this file until it has been read by the WebUI. Then after it has been read, will attempt to write any changes. `theWebUI.save()` is the only function which calls `?action=setuisettings` where it writes inside `setsettings.php`.

php/util.php
stickz: Good catch, I will create the real path sooner. To resolve this problem.

php/util.php
stickz: Actually this variable is undefined here. I'm just going to remove this if statement and set it's value.

stickz:
> > Performance wise, this reduces loading times by approximately 300ms on average.
> 
> Please, don't repair non-broken things.
This is broken. It has to read the file into a string and print it in PHP.  It should be downloading the file in JavaScript and parsing it into a JSON object instead. The script execution time is 300ms on the web server. It's better to parse in 5ms on the client-side. If you ever decide to make this JSON file bigger, there's going to be huge problems with this transmission method.

php/getsettings.php
stickz: I'm going to rewrite some code. I will use the rCache class to store the webui settings in a .dat file. I will let this class handle the file locking. There will be no race conditions when I'm done. The WebUI will still be able to make async requests.

stickz:
@Novik This pull request is ready for review again. The race conditions are gone now. There is a new `WebUISettings` class which uses `rCache` to store the ui settings in a `WebUISettings.dat` file now. 

**There are multiple advantages to this method:**
1. Memory mapping is now used when reading the file to speed things up.

2. It's no longer required to write the JSON contents to the disk unless they actually change.

3. We no longer have race conditions to complete tasks. 
 
The WebUI can send tasks asynchronously now. If the `WebUISettings.dat` file is currently being accessed, the thread will usleep for a short duration of time, until it's available. The previous behavior for writes was to not successfully complete the task. 

4. This class method is also more abstract and leaves additional room for improvement in the future. For instance, it may be possible to only post a JSON string of changed values to `setsettings.php`.

**Performance Advantages:** 
1. `getsettings.php` task completion time has been reduced from 300ms to 5ms (plus the network transfer time). It also runs asynchronously and allows the web ui to load these settings without waiting for `getplugins.php`.

2. Web server load is reduced because we're doing a lot more cached reading now. And writes only happen when the contents of the JSON string change.

stickz:
This pull request broke when I was fixing some things with my master branch. I will open a new pull request for this later.

PR #2242, opened 2021-12-31
Improve WebUI settings

stickz:
This pull request greatly improves the process of loading and saving WebUI settings. I originally created PR #2239, but decided to rethink how to implement this improvement. There are no race conditions here and it doesn't fix anything that is not broken. 

**List of Changes:**

1. In `rtorrent.js` the `getsettings.php` AJAX request is changed from a POST method to GET method. We are getting data from the server not sending data. Flagging the method as POST is not the correct way of completing this request.

2. The WebUI now sends an asynchronous request to retrieve the UI settings that are stored on the web server. This reduces the [TBT (Total Blocking Time)](https://web.dev/tbt/?utm_source=devtools) of the main thread before it's time to finish initializing the WebUI. The WebUI will not finish initializing until this task has completed, to ensure there's no untended side effects with saved settings or initializing plugins.

3. `theWebUI.config()` no longer contains the data from the `getsettings.php` AJAX request as a function parameter. None of the default ruTorrent plugins currently require this information. No third party plugins will require this information ether. A duplicate copy is stored in `theWebUI.settings` array. All of the default plugins have been updated to reflect this change. 

It is very important to implement this change because it prevents the possibly of a memory of a leak being created. It also prevents plugins from accidentally not sending this information back to the WebUI and causing bugs with loading settings. 

4. There is one bug with loading WebUI settings which has been resolved with this pull request. The wrong WebUI settings were being loaded because of a race condition with writing to `uisettings.json` on the web server. This problem is now fixed.

5. There is a new `WebUISettings` class for reading and writing WebUI settings on the web server. The WebUI settings are now stored in `WebUISettings.dat` instead of `uisettings.json`. They are loaded and saved by the `rCache` class instead.

- The WebUI settings are now cached into memory instead of being repeatedly read from the disk. This reduces overhead.
- The WebUI settings are no longer written to the disk unless the JSON string changes. This is possible due to cached loading.
- There is no longer a race condition when writing the WebUI settings to the disk. `rCache` will safely usleep the thread until the file lock is removed. The previous behavior was not to update the WebUI settings, if the file was locked.

**Future Improvements:** 
If this pull request is merged, it will allow for future improvements to be implemented for the WebUI settings. I did not include them with this pull request. It's better to have a smaller pull request to make code review and testing easier for everyone.

1. It will be possible to finish initializing the WebUI, even if the request to retrieve the WebUI settings from the web server fails.
- A translated warning can be displayed to the user, so they can attempt to correct the problem.
- The WebUI can temporarily load with the default settings, so the user can still use ruTorrent.
- The WebUI settings can be prevented from saving settings. This way, the user will not lose their originally saved settings.

2. It will be possible to only POST the changed value to the web server, when a WebUI setting is updated. This will greatly reduce the AJAX request overhead for `setsettings.php`. It will be possible to send multiple async requests at once without having to worry about the overhead. Also, a larger JSON string will be possible. This will make future improvements practical.

### DIFF
--- a/conf/config.php
+++ b/conf/config.php
@@ -22,9 +22,21 @@
 	// for php
 	@define('PHP_USE_GZIP', false, false);
 	@define('PHP_GZIP_LEVEL', 2, true);
+	
 
-	$schedule_rand = 10;			// rand for schedulers start, +0..X seconds
-
+class RTorrentConfig
+{	
+	public $schedule_rand = 10;	// rand for schedulers start, +0..X seconds
+	
+	public $pathToExternals = array(
+		"php" 	=> '',			// Something like /usr/bin/php. If empty, will be found in PATH.
+		"curl"	=> '',			// Something like /usr/bin/curl. If empty, will be found in PATH.
+		"gzip"	=> '',			// Something like /usr/bin/gzip. If empty, will be found in PATH.
+		"id"	=> '',			// Something like /usr/bin/id. If empty, will be found in PATH.
+		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
+	);	
+}
+	
 	$do_diagnostic = true;
 	$log_file = '/tmp/errors.log';		// path to log file (comment or leave blank to disable logging)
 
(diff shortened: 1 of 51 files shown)


stickz:
Sorry, I've experienced more problems with GitHub. I will fix this soon,

PR #2243, opened 2021-12-31
Improve WebUI settings

stickz:
This pull request greatly improves the process of loading and saving WebUI settings. I originally created PR #2239, but decided to rethink how to implement this improvement. There are no race conditions here and it doesn't fix anything that is not broken. 

**List of Changes:**

1. In `rtorrent.js` the `getsettings.php` AJAX request is changed from a POST method to GET method. We are getting data from the server not sending data. Flagging the method as POST is not the correct way of completing this request.

2. The WebUI now sends an asynchronous request to retrieve the UI settings that are stored on the web server. This reduces the [TBT (Total Blocking Time)](https://web.dev/tbt/?utm_source=devtools) of the main thread before it's time to finish initializing the WebUI. The WebUI will not finish initializing until this task has completed, to ensure there's no untended side effects with saved settings or initializing plugins.

3. `theWebUI.config()` no longer contains the data from the `getsettings.php` AJAX request as a function parameter. None of the default ruTorrent plugins currently require this information. No third party plugins will require this information ether. A duplicate copy is stored in `theWebUI.settings` array. All of the default plugins have been updated to reflect this change. 

It is very important to implement this change because it prevents the possibly of a memory of a leak being created. It also prevents plugins from accidentally not sending this information back to the WebUI and causing bugs with loading settings. 

4. There is one bug with loading WebUI settings which has been resolved with this pull request. The wrong WebUI settings were being loaded because of a race condition with writing to `uisettings.json` on the web server. This problem is now fixed.

5. There is a new `WebUISettings` class for reading and writing WebUI settings on the web server. The WebUI settings are now stored in `WebUISettings.dat` instead of `uisettings.json`. They are loaded and saved by the `rCache` class instead.

- The WebUI settings are now cached into memory instead of being repeatedly read from the disk. This reduces overhead.
- The WebUI settings are no longer written to the disk unless the JSON string changes. This is possible due to cached loading.
- There is no longer a race condition when writing the WebUI settings to the disk. `rCache` will safely usleep the thread until the file lock is removed. The previous behavior was not to update the WebUI settings, if the file was locked.

**Future Improvements:** 
If this pull request is merged, it will allow for future improvements to be implemented for the WebUI settings. I did not include them with this pull request. It's better to have a smaller pull request to make code review and testing easier for everyone.

1. It will be possible to finish initializing the WebUI, even if the request to retrieve the WebUI settings from the web server fails.
- A translated warning can be displayed to the user, so they can attempt to correct the problem.
- The WebUI can temporarily load with the default settings, so the user can still use ruTorrent.
- The WebUI settings can be prevented from saving settings. This way, the user will not lose their originally saved settings.

2. It will be possible to only POST the changed value to the web server, when a WebUI setting is updated. This will greatly reduce the AJAX request overhead for `setsettings.php`. It will be possible to send multiple async requests at once without having to worry about the overhead. Also, a larger JSON string will be possible. This will make future improvements practical.


stickz:
I just fixed an issue with the WebUISettings not loading properly on the first load. It sends an empty JSON string the first time settings load. The loading routine was expecting a value for "webui.lang" inside the JSON string. I had to add a null check.

Novik:
What about compatibility with [this](https://github.com/Novik/ruTorrent/pull/2247) ?

stickz:
> What about compatibility with [this](https://github.com/Novik/ruTorrent/pull/2247) ?
I will make this pull request compatible with PR #2247. I just have to update a couple files to use the new utility classes. It may take up to four days to complete this task. I have to return to work within one day.

js/webui.js
```diff
@@ -617,7 +633,8 @@ var theWebUI =
 				}
          		}
 		});
-		newSettings["webui.lang"] = GetActiveLanguage();
+		if ($type(newSettings["webui.lang"]))
```
Novik, line 636: This is a strange condition. Why we need it? 
If key 'webui.lang' is absent, then we must create it.

stickz: We need it for safety reasons. Otherwise, the WebUI will not initialize if `getsettings.php` sends an empty JSON string.

Novik: Sorry, I can't understand. 
For example, ```getsettings.php``` sends an empty JSON string, i.e. ```{}```.
In the ```newSettings``` we will have an empty object. 
Why we can't create new key for this object with a language setting?

stickz: Okay this problem has been resolved.

stickz:
I fixed setting of the webui lang in the last commit. I also resolved anther regression with trying to configure the webui multiple times. This was firing `setsettings.php` unnecessarily and was causing an overflow when trying to open the settings menu.

js/webui.js
```diff
@@ -597,6 +613,21 @@ var theWebUI =
 // settings
 //
 
+	initSettings: function(newSettings)
+	{
+		// If when receive an empty string for the webui settings,
+		// We must initialize "webui.lang", so it can be set
+		if (newSettings=="{}")
+			newSettings = {"webui.lang": ""};
+
```
Novik, line 622: 1) Your comment is incorrect, because "{}" is not an empty string.
2) AFAIK, in the ```newSettings``` we always have object. How we may have string in it?

js/webui.js
```diff
@@ -615,7 +615,7 @@ var theWebUI =
 
 	initSettings: function(newSettings)
 	{
-		// If when receive an empty string for the webui settings,
+		// If when receive an empty JSON object for the webui settings,
 		// We must initialize "webui.lang", so it can be set
 		if (newSettings=="{}")
```
Novik, line 620: 1) You can't compare Object with a String. I.e. you can, but you will always have 'false' here.
Try to write something like ```({}=="{}");``` in the browser console...
2) Why we need this?

stickz: 1. It works perfectly fine, I tested it. It's not an object, it's a string being sent here.

2. We can't reference "webui.lang" as a key then set it's value for a JavaScript array, if the variable is in string format.

stickz: Yes, it does return a JavaScript object. The object may have a primitive value of `"{}"` (which is basically a string), if for some reason `gettsettings.php->WebUISettings.php` can't load the .dat file. Like for instance on the first run, when it's not preset. This is why we can compare it as a string. If the JavaScript object had a value of say `{"webui.lang": ""}` this would always return false. The JavaScript object being returned can be a simple (primitive) object or a complex object.

Novik: > The object may have a primitive value of "{}" (which is basically a string)

Object is a object. String is a string. Object can't "have a primitive value of '{}'". Object may be an empty.
But you can't check this with a such strange way.

> Even if you're right and this is not a string, == is being used here which does not check the datatype.

This haven't relation to the question because Object cannot be brought to string by automatic way.

Yet once, open browser console and print in it 
```
({}=="{}");
```
This is a comparison of empty object and string "{}".
What do you see as result?
I see 'false'. 

If you want to check empty object you should write something like 
```
(JSON.stringify({})=="{}");
```
or [use jQuery.isEmptyObject()](https://api.jquery.com/jquery.isemptyobject/)

stickz: Alright, I used `JSON.stringify`. I don't think this is required, but it doesn't matter because it works ether way.

PR #2244, opened 2022-01-01
Fix plugin diagnostics

### LINKED ISSUE #2228: RuTorrent plugins can't access python, php and so on.
So I hade some issues on my CentOS 8 server. I wanted to upgrade python, however I were stupid enough to remove the current python (3.6) first... I finally got my system up and running again, got python 2.7, 3.6 and 3.9 now. I can access the rutorrent site but I get a filled log that the plugins can't access python, php and so on. I even tried pointing on these applications in /usr/bin/XX in the config-file. Yet it does not work.
Another thing is that if I specify python in the config-file, I get: `Bad response from server: (500 [error,getplugins]) Internal Server Error`. So using python from the PATH seems to work, but the plugins can't access any program?

stickz:
This pull request fixes #2228. `tesh.sh` throws the error `"bash: ./test.sh: /bin/sh^M: bad interpreter"` when being run by plugin diagnostics. The cause is the `tesh.sh` file being modified by Windows with "dos" line endings. This prevents the bash script from being run on Linux. The solution is to change the line endings on the file back to "unix" (Linux) format.

The `perl` package dependency is required to safely modify the bash script. Other methods may accidently delete the file contents. It must be installed on the same server as rTorrent. The user is warned to install the `perl` package first. ruTorrent may disable other plugins and incorrectly show missing packages on the rTorrent server, until this problem is fixed. 

This feature is only enabled when `$do_diagnostic` in `config.php` is set to true. It is purely for diagnostic purposes only.

### DIFF
--- a/conf/config.php
+++ b/conf/config.php
@@ -52,6 +52,7 @@
 		"gzip"	=> '',			// Something like /usr/bin/gzip. If empty, will be found in PATH.
 		"id"	=> '',			// Something like /usr/bin/id. If empty, will be found in PATH.
 		"stat"	=> '',			// Something like /usr/bin/stat. If empty, will be found in PATH.
+		"perl" 	=> '', 			// Something like /usr/bin/perl. If empty, will be found in PATH.
 	);
 
 	$localhosts = array( 			// list of local interfaces
--- a/lang/cs.js
+++ b/lang/cs.js
@@ -208,6 +208,8 @@ var theUILang =
  showScrollTables		: "Show table contents while scrolling",
  idNotFound			: "rTorrent user can't access 'id' program. Some functionality will be unavailable.",
  gzipNotFound			: "Webserver user can't access 'gzip' program. ruTorrent will not work.",
+ perlNotFound			: "rTorrent user can't access 'perl' program. ruTorrent will not work.",
+ perlDependency			: "Please install 'perl' package first. It may resolve other plugin errors.",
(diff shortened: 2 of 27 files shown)


Novik:
> The cause is the tesh.sh file being modified by Windows with "dos" line endings. 

Tomorrow somebody will modify another script locally, and what next? You will write yet another patch for this?
Sorry, but in my opinion we don't need this.

stickz:
> Tomorrow somebody will modify another script locally, and what next? You will write yet another patch for this? Sorry, but in my opinion we don't need this.
This is not what's happening. The user is transferring files from Windows to their Linux server most likely by FTP. This is an unavoidable problem, which is very annoying and difficult to fix. The problem is caused by the bash script in `getplugins.php`.

IMHO, we need to fix this problem if that bash script for remote requests is going to be supported. I will create anther pull request and check if the user is running rTorrent and ruTorrent on the same server. If they are, I will replace with the remote request with a local request for the program. This way the problem will at least be isolated to running separate servers. It will be faster anyways and have less overhead than sending the request through XMLRPC. I'll be able to do a cached lookup very soon.

Novik:
> The user is transferring files from Windows to their Linux server most likely by FTP

1) This is not a normal way for install software.
2) In this case your patch will not help, because user can't set correct execute permission for the .sh files.

> If they are, I will replace with the remote request with a local request for the program.

After this replacing we will have a problem with permissions, when binary is available for web-user and not available for rtorrent user.
Please, stop repair non-broken things.

PR #2245, opened 2022-01-01
Improve File Lookups

stickz:
This pull request implements a better solution for finding Linux packages and caches the results to prevent unnecessary lookups. It's required to move `findEXE()` to the `which.php` file. `util.php` is included too often to load file paths there.

The `which` package has been around for a very long time and is included with almost every Linux distribution. This is a more reliable way to find a program than relying on the PHP - "PATH" environment variable. In some cases the "PATH" environment variable is not available for PHP, then the package fails to be found automatically. It's more difficult to fix this problem.

In rare cases where user doesn't have `which`, they can add the file paths manually into `config.php` or install it. They will probably have to do this anyways, if they are using a special Linux distribution, where all packages are installed manually.


php/which.php
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function findEXE( $exe )
+{
+	global $pathToExternals;
+	if(isset($pathToExternals[$exe]) && !empty($pathToExternals[$exe]))
+		return(is_executable($pathToExternals[$exe]) ? $pathToExternals[$exe] : false);
+	
+	global $whichCache;
+	return($whichCache->getFilePath($exe));
```
Novik, line 69: Try to avoid global variables and direct call of constructors for singletons, if this possible.
For example, here we may have instead something like

	static private $theCache = null;

	static public function get($do_diagnostic = false)
	{
		if(is_null(self::$theCache))
		{
			self::$theCache = new WhichCache($do_diagnostic); // or WhichCache::load if you prefer this way
		}
		return(self::$theCache);
	}
	...
	WhichCache::get()->getFilePath($exe);

php/which.php
```diff
@@ -0,0 +1,72 @@
...
+		$cache = new rCache();
+		$which = new WhichCache();
+		$cache->get($which);
+		if ($diagnostic)
+			$which->pruneCache();
+		return $which;
+	}
+	
+	public function GetFilePath($exe)
+	{
```
Novik, line 20: **g**etFilePath maybe?

Novik:
> This is a more reliable way to find a program than relying on the PHP - "PATH" environment variable. In some cases the "PATH" environment variable is not available for PHP

"which" command also use search in the "PATH" environment variable. 
If "PATH" environment variable is not available for PHP, then it is not available for the "which" command too.

koblack:
'which' is deprecated; use `command -v' in scripts instead.

See https://lwn.net/SubscriberLink/874049/b563dc08d4eb5829/ for the full story.

stickz:
I switched to `command -v` instead of `which`. This shell built-in is guaranteed to work everywhere, has less overload, prevents adding anther dependency to the project and will always produce the same output. The previous PATH method was not working for myself, so this issue has been resolved. ruTorrent will no longer say a program is unavailable, when it actually is.

I could not resolve the global variable issue, but I managed to avoid directly calling the constructer by creating anther class. It's not possible to store static instances of the object you're caching with `rCache`. Otherwise, the feature will not work properly.

stickz:
Okay @Novik this pull request is ready for review again. The global variable and singleton issue is fixed now. I also did the same thing for `rCache`. I ended up creating anther class called `WhichInstance` to store these static methods. 

The only catch is that we have to call `WhichInstance::save()` when done initializing plugins. It's not possible to do this automatically anymore. It works better this way anyways because we're not writing to the disk as often on the first run.

PR #2247, opened 2022-01-02
Refactor code & implement autoloader

stickz:
This pull request refactors the ruTorrent utility code to reduce script execution times and improve overall code readability. It moves all of the functions in `php/util.php` out of the global context and into classes. When the utility file is included, only the required functionality is added is each time. This prevents loading of unnecessary functionality for PHP scripts.

It implements a very basic auto loader function to simplify the including process of utility classes. The only file which needs to be included is `php/util.php`. When a utility class is used, PHP will automatically go to the `/php/utility` folder to include the file.
```
spl_autoload_register(function ($class) {
    require_once 'utility/'. strtolower($class). '.php';
});
```
Utility class names can have any mixture of lower or upper case letters. However, the files in the utility folder:
1. Must contain only 1 class per file.
2. Must end with the file extension `.php`.
3. Must have the same file name as the class inside of it.
4. Must have a file name in lower case letters only.

**Other Notes:**
1. Some configuration variables which only need to be accessible by individual classes now belong to a new `ruTorrentConfig` class. This further helps to cut down on global variables and makes code much easier to debug.

2. The default value of `profilePath` variable had to be changed from `../share` to `../../share.` The reason is because it accessed from `fileutil.php` in `/php/utility` now instead of `util.php` in `/php`. The folder context is different to reach the root folder.

3. In the `sqlite_exists()` function in `/plugins/geoip/sqlite.php` the second argument `class_exists` had to be changed from `true` to `false`. It may be required to update the `spl_autoload_register()` function, if SQLite3 does not automatically load.


stickz:
I added some instance caching of results for a few static methods as well. I found out `preg_replace` was being called on `$_SERVER['REMOTE_USER']` about a hundred times during the plugin initialization phase.

Novik:
Very massive pull request, I will see it later.

```
spl_autoload_register(function ($class) {
    require_once 'utility/'. strtolower($class). '.php';
});
```

1) I can see 'spl_autoload_register' in the list of changes. Maybe you forget to include some files?
2) This will not correctly work in such form. We need absolute path here, i.e. ```require_once __DIR__ . ...```
3) Maybe, you will think about usage of [composer](https://getcomposer.org/) instead?

stickz:
`spl_autoload_register'` is there. It's buried behind 760 lines of removed code in `util.php`. Everything worked when I tested this.

Novik:
In the current implementation we may have several user's profiles for the single ruTorrent instllation.
And every user may have [his own](https://github.com/Novik/ruTorrent/wiki/Config#multiuser-and-singleuser-configuration) config.php file with a his own preferences (include profile path, proxy settings and so on).
How this feature will work after your changes?
Will class constants and variables be overridden correctly?

And even if yes, then now we have strange situation, when configuration partially inside global variables and partially in the class declarations. This is not good thing IMHO.
In the ideal situation we should to have only one global variable with something like this - https://github.com/hassankhan/config

stickz:
There is a problem with user specific `config.php` files. I will fix this issue in a couple of days with anther commit.

php/utility/fileutil.php
```diff
@@ -64,12 +63,12 @@ public static function getProfilePath( $user = null )
 		return(self::$profilePathInstance);
 	}
 	
-	public static function getSettingsPath( $user = null )
+	public static function getSettingsPath()
 	{
 		return( self::getProfilePath($user).'/settings' );
```
Novik, line 68: 1) This is incorrect, because now getProfilePath does not have parameters.
2) Please, keep '$user' parameter for getUploadsPath/getSettingsPath/getProfilePath functions. This is required for the third-party plugins compatibility.

stickz: Third party plugins are going to break anyways by this pull request. Are you okay with a new set of functions for this functionality? It's not possible to cache the profile path for the user, with this '$user' parameter.

**For instance:** getUploadsPathEx/getSettingsPathEx/getProfilePathEx could have the user parameter and no caching.

stickz:
I reverted the `ruTorrentConfig` class. A configuration loader can be added in anther pull request. This is beyond the intended scope of this pull request. The purpose here is to remove to functions in `php/util.php` from the global scope.

php/utility/fileutil.php
Novik: > It's not possible to cache the profile path for the user, with this '$user' parameter.

Why? For example, something like that:

```
if(!isset(self::$profilePathInstance[$user]))
{
	self::$profilePathInstance[$user] = ...
}
return(self::$profilePathInstance[$user]);
```

But this is not principal.

php/utility/json.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
```
Novik, line 2: What about ```require_once('utf.php');``` ?
In the other files (for example [here](https://github.com/Novik/ruTorrent/pull/2247/files#diff-22928530cdd4e4a8b28afbaffcfc053f070b2458bec06885ef4ab8c21b881593)) we have additional require statements.
Is this correct?

stickz: Yes, we should add this just to be safe. I don't know if the auto loader will do this automatically. Even if it does this automatically, it's a micro optimization to add it ourselves, if we know it's going to be required.

PR #2252, opened 2022-01-09
Add 'copyToClipboard' function in common.js + a few fixes

Micdu70:


plugins/bulk_magnet/init.js
```diff
@@ -50,13 +38,13 @@ if(plugin.canChangeMenu())
...
-				break;	
+				break;
 			}
 		}
 		return(ret);
 	}
 
-	plugin.copyProperty = function(type)
+	theWebUI.copyProperty = function(type)
 	{
```
Novik, line 48: Do you really have plan to use 'theWebUI.copyProperty' somewhere else?
If no, then do not extend global interface 'theWebUI', use local 'plugin' interface instead, please.
Pollution of global interfaces with a non-common functions is not a good thing.

Micdu70: I really would like to keep using the 'plugin' interface but it doesn't work:
=> "plugin.copyProperty()":
----- Get `plugin is not defined` error when using double quotes in 'theContextMenu.add' call.
----- Without double quotes, all 3 'copyProperty' functions are called when the context menu is opened in ruTorrent.
=> "plugin.copyProperty" without "()" at the end:
----- Get `plugin is not defined` error when using double quotes in 'theContextMenu.add' call.
----- Without double quotes, copyProperty function is never called (and that's why I created this PR).

Novik: What about previous implementation, is it work?

Micdu70: Previous implementation doesn't seem to work (at least not with my personal ruTorrent installation). I don't know the last time this worked.

Novik:
@Micdu70 I already fixed this, thanks.

PR #2253, opened 2022-01-10
Fix loading utility classes from namespaces

stickz:
This pull request fixes compatibility with 3rd party plugins which use namespaces. An autoloader was introduced in PR #2247. When we get the classname string from `spl_autoload_register`, we need to remove any namespaces from it. Otherwise, 3rd party plugins with namespaces will not load utility classes properly.

### DIFF
--- a/php/util.php
+++ b/php/util.php
@@ -9,7 +9,8 @@ spl_autoload_register(function ($class)
 {
 	// Remove namespaces from the classname string
 	// Important for compatibility with 3rd party plugins
-	$class = end(explode('\\',$class));
+	$arr = explode('\\',$class);
+	$class = end($arr);
 	
 	require_once 'utility/'. strtolower($class). '.php';
 });
--- a/php/util.php
+++ b/php/util.php
@@ -9,9 +9,7 @@ spl_autoload_register(function ($class)
 {
 	// Remove namespaces from the classname string
 	// Important for compatibility with 3rd party plugins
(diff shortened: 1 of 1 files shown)


php/util.php
```diff
@@ -5,8 +5,15 @@
 require_once( $rootPath.'/conf/config.php' );
 
 // Automatically include only the used utility classes
-spl_autoload_register(function ($class) {
-    require_once 'utility/'. strtolower($class). '.php';
+spl_autoload_register(function ($class) 
+{
+	// Remove namespaces from the classname string
+	// Important for compatibility with 3rd party plugins
+	if (preg_match('@\\\\([\w]+)$@', $class, $matches)) {
```
Novik, line 12: We really need PCRE here? What about
```
	$arr = explode('\\',$class);
	$class = end($arr);
```
?
IMHO this will be faster.

stickz: Alright done.

php/util.php
```diff
@@ -5,8 +5,13 @@
 require_once( $rootPath.'/conf/config.php' );
 
 // Automatically include only the used utility classes
-spl_autoload_register(function ($class) {
-    require_once 'utility/'. strtolower($class). '.php';
+spl_autoload_register(function ($class) 
+{
+	// Remove namespaces from the classname string
+	// Important for compatibility with 3rd party plugins
+	$class = end(explode('\\',$class));
```
Novik, line 12: PHP Notice:  Only variables should be passed by reference
I.e. we should use temporary variable for 'explode' result.

PR #2260, opened 2022-01-12
Fix timeouts on AJAX requests

### LINKED ISSUE #2256: Uncaught TypeError: T.abort is not a function
Master branch, updated to the lastest commit.
```
11.01.2022 17:41:09] JS error: [http://xxx.xxx.xxx.xxx:xx/js/jquery.js : 2] Uncaught TypeError: T.abort is not a function
```

stickz:
Resolves #2256 by fixing a regression with #2241. We can't nullify the abort component of the AJAX request object. jQuery needs to be able to call this function internally, to submit the timeout back to theWebUI.

### DIFF
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -1228,7 +1228,7 @@ function Ajax(URI, isASync, onComplete, onTimeout, onError, reqTimeout)
 	
 	// Nullify ajax request varriables to cleanup up memory leaks
 	request.onreadystatechange = null;
-	if (onTimeout == null)
+	if (reqTimeout==-1)
 		request.abort = null;
 	request = null;
 }
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -1228,7 +1228,8 @@ function Ajax(URI, isASync, onComplete, onTimeout, onError, reqTimeout)
 	
 	// Nullify ajax request varriables to cleanup up memory leaks
 	request.onreadystatechange = null;
-	request.abort = null;
+	if (onTimeout == null)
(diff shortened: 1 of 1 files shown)


Novik:
We *always* have a timeout due to this - 
```
timeout: reqTimeout || 10000,
```
IMHO we should remove nullification of 'onreadystatechange' and 'abort', because we haven't references to them in our code.

stickz:
The problem is that JQuery is using them and leaking memory. If we don't need this functionality, it should be nullified. There's no reason for us or JQuery to do an abort, if the AJAX request has no time out. If we want the request to abort, we can set a timeout. `onreadystatechange` is for IE9. We don't need this because we're not doing any aborts in our code.

Novik:
> There's no reason for us or JQuery to do an abort, if the AJAX request has no time out. 

Sorry, but my question is not "why we need nullfication of 'abort' and 'onreadystatechange' properties only when we we have timeout?".
My question is "why we need this nullfication at all?"

> The problem is that JQuery is using them and leaking memory

1) In this case *any* usage of jQuery ajax in *any* project without nullification of this properties will have as result memory leaking. Where I may read about this?
2) As I already told you, jqXHR has a lot of properties. For example, "ontimeout", "readyState", "upload" etc. Why you not nullify them all?

stickz:
Okay, I have removed nullification of `abort` entirely. If there are any issues with `onreadystatechange` being nullified, I will create anther pull request and remove it as well. This change is already incorporated into master. I don't want to fix what is not broken.

PR #2263, opened 2022-01-13
Optimize 'tracklabels' plugin

stickz:
This pull request optimizes the tracklabels plugin and improves overall loading times of ruTorrent. It contains two changes:

1. When an image is not found for tracker/torrent labels, it will remember this for 30 days. `unknown.png` will be loaded from the local computer. This saves us from having to make anther web server request and wait for script execution.

2. The tracker rows now refresh asynchronously, so it doesn't slow down the adding of tracker data into the WebUI. Calling `dxSTable.refreshRows` and `dxSTable.sort` can be done any time, as the object reference to the table is stored in the WebUI.

### DIFF
--- a/plugins/tracklabels/init.js
+++ b/plugins/tracklabels/init.js
@@ -139,7 +139,7 @@ theWebUI.updateLabels = function(wasRemoved)
 	plugin.updateLabels.call(theWebUI,wasRemoved);
 	if(plugin.enabled)
 	{
-		if(wasRemoved && !theWebUI.firstLoad)
+		if(wasRemoved)
 			theWebUI.rebuildTrackersLabels();
 		plugin.updateLabelsImages();
 	}
@@ -192,12 +192,6 @@ theWebUI.rebuildTrackersLabels = function()
 				}
 			}
 		}
-		if(plugin.canChangeColumns())
-		{
-			table.refreshRows();
-			if(table.sIndex !=- 1)
-				table.Sort();		
-		}
 		var ul = $("#torrl");
 
 		var keys = new Array();
@@ -237,6 +231,19 @@ theWebUI.rebuildTrackersLabels = function()
 		this.trackersLabels = trackersLabels;
 		if(needSwitch)
 			theWebUI.resetLabels();
+		
+		setTimeout(plugin.refreshTrackerRows, 0);
+	}
+}
+
+plugin.refreshTrackerRows = async function()
+{
+	if(plugin.canChangeColumns())
+	{
+		var table = theWebUI.getTable('trt');
+		table.refreshRows();
+		if(table.sIndex !=- 1)
+			table.Sort();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


plugins/tracklabels/init.js
```diff
@@ -123,7 +123,7 @@ theWebUI.trackersLabelContextMenu = function(e)
 	return(false);
 }
 
-plugin.updateLabelsImages = function()
+plugin.updateLabelsImages = async function()
```
Novik, line 126: We not need 'async' here. The async function declaration simply makes the function automatically return a promise.
Also you can use the await operator inside of async function. That's all.
Try to check something like this:
```
async function f(){
    console.log(2);
}

console.log(1);
f();
console.log(3);
```
You will have "1 2 3", not "1 3 2".

stickz: Okay, I have actually made this task async using `setTimeout`. The function is no longer returning a promise to `theWebUI.updateLabels`. I agree this was undesirable behavior. It's also a micro-optimization to do this because we're mitigating (in some cases removing) a forced reflow during `theWebUI.addTorrents`.

The real reason though is because other tasks like `cpuload` and `diskspace` were waiting on `tracklabels/action.php`. In some cases they were delayed by 200ms during startup. These tasks are higher priority tasks because they require a network transfer to complete. The labels can be loaded from cache anytime, but these need to start right away.

Novik: > Okay, I have actually made this task async using setTimeout

This task *already* is asynchronous. Function ```updateLabelsImages``` simple modify a set of ```src``` attributes. It does not wait the result of loading of corresponding images.
Your changes in this fragment is useless, IMHO.

stickz: I am using the performance profiler on Google Chrome. A forced reflow was eliminated, reducing the time it takes to load labels from 10ms to 3ms. But the real benefit is `cpuload/action.php` and `diskspace/action.php` were initiated before `tracklabels/action.php`. This made the cpu load and disk space icons load 150ms faster. They must retrieve the percentage used from the web server using a network request, before the icons can be loaded.

Novik: > A forced reflow was eliminated, reducing the time it takes to load labels from 10ms to 3ms

Sorry, but I think this is a simple measurement error. 

> But the real benefit is cpuload/action.php and diskspace/action.php were initiated before tracklabels/action.php

For example, on the my installation (master branch) they *already* initiated before, and your changes haven't relation to this.
First request of cpuload:
2022-01-13_21h53_17

First request of diskspace:
2022-01-13_21h59_35

First request of tracklabels:
2022-01-13_21h53_44

stickz: That's not how it working on my waterfall page. But I would like this pull request merged, so I reverted this change. I'll take what I can get. It's better to have two changes incorporated into master than none.

PR #2266, opened 2022-01-16
Ratio improvements & fixes

stickz:
This pull request improves the ratio group tab (for the ratio plugin) in the ruTorrent settings menu.

**The following changes are made by this pull request:**
1. The UL value is now measured in gigabytes instead of megabytes.
2. The default UL value is increased from 20MB to 100MB.
3. The min UL value is increased from 1MB to 10MB.
4. The max UL value is increased from 1TB to 1PB.
5. The UL value now accepts 2 decimal places.

6. A regression from #2247 is fixed when setting the ratio group UL Targets. The Math class was not referenced properly
7. A new translated legend is added beside the default ratio group selector. This explains to the user how the UL value works. It also makes it easier for the user to do terabyte and megabyte conversations, without having to look up the values.

**Here is a screenshot of how the ratio group tab looks now:**
ratio_group

**What is the motivation behind these changes?**
I wanted to set a 3 terabyte UL value for my torrent. I was not able to do that. 999999 megabytes which equals 1 terabyte was the max value. Also, it was difficult to set gigabyte UL values because the measurement unit was megabytes. Bandwidth has increased significantly since ruTorrent was first released. Setting megabyte UL values is no longer applicable for many people.

### DIFF
--- a/plugins/ratio/ratio.php
+++ b/plugins/ratio/ratio.php
@@ -16,17 +16,34 @@ class rRatio
 {
 	public $hash = "ratio.dat";
 	public $rat = array();
-	public $default = 0.1;
+	public $default = 0;
+	private $version = 3.10;
 
 	static public function load()
 	{
 		$cache = new rCache();
 		$rt = new rRatio();
 		if(!$cache->get($rt))
+		{
 			$rt->fillArray();
+			$rt->version = 3.11;
+		}
+		elseif ($rt->version != 3.11)
+		{
+			$rt->migrate();
+			$rt->pad();
+			$rt->version = 3.11;
+		}
 		else
 			$rt->pad();
 		return($rt);
+	}
+	private function migrate()
+	{
+		for($i=0; $i<count($this->rat); $i++)
+		{
+			$this->rat[$i]["upload"] /= 1000;
+		}		
 	}
 	public function pad()
 	{
@@ -44,7 +61,7 @@ class rRatio
 	{
 		$this->rat = array();
 		$this->pad();
-		$this->default = 0.1;
+		$this->default = 0;
 	}
 	public function getTimes()
 	{
@@ -222,7 +239,7 @@ class rRatio
 	public function set()
 	{
 		$this->rat = array();
-		$this->default = 0.1;
+		$this->default = 0;
 		for($i = 0; $i<MAX_RATIO; $i++)
 		{
 			$arr = array( "action"=>RAT_STOP, "min"=>100, "max"=>300, "upload"=>0.1, "name"=>"", "time"=>-1 );
--- a/plugins/ratio/ratio.php
+++ b/plugins/ratio/ratio.php
@@ -16,7 +16,7 @@ class rRatio
 {
 	public $hash = "ratio.dat";
 	public $rat = array();
-	public $default = 0;
+	public $default = 0.1;
 
 	static public function load()
 	{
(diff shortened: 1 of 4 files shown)


plugins/ratio/ratio.php
```diff
@@ -159,7 +159,7 @@ public function flush()
 					$req->addCommand( rTorrentSettings::get()->getRatioGroupCommand("rat_".$i,'ratio.enable',array("")) );
 					$req->addCommand( rTorrentSettings::get()->getRatioGroupCommand("rat_".$i,'ratio.min.set',$rat["min"]) );
 					$req->addCommand( rTorrentSettings::get()->getRatioGroupCommand("rat_".$i,'ratio.max.set',$rat["max"]) );
-					$req->addCommand( rTorrentSettings::get()->getRatioGroupCommand("rat_".$i,'ratio.upload.set',floatval($rat["upload"]*1024*1024)) );
+					$req->addCommand( rTorrentSettings::get()->getRatioGroupCommand("rat_".$i,'ratio.upload.set',floatval($rat["upload"]*1024*1024*1024)) );
```
Novik, line 162: What about compatibility with an old data?
For example, user already have data in the "ratio.dat". And this data in the old format, i.e. $rat["upload"] in MiB.
What he will have at this point? MiB instead GiB?

plugins/ratio/ratio.php
```diff
@@ -35,16 +35,16 @@ public function pad()
...
 		for($i=count($this->rat); $i<MAX_RATIO; $i++)
-			$this->rat[] = array( "action"=>RAT_STOP, "min"=>100, "max"=>300, "upload"=>20, "name"=>"ratio".$i, "time"=>-1 );
+			$this->rat[] = array( "action"=>RAT_STOP, "min"=>100, "max"=>300, "upload"=>0.1, "name"=>"ratio".$i, "time"=>-1 );
 	}
 	public function fillArray()
 	{
 		$this->rat = array();
 		$this->pad();
-		$this->default = 0;
+		$this->default = 0.1;
```
Novik, line 47: Why we need this?

plugins/ratio/ratio.php
stickz: The old data is currently setup to be whipped and replaced with new data when they upgrade the plugin.

If this is a problem, I can migrate the data to the new configuration by adding anther function to `ratio.php`.

plugins/ratio/ratio.php
stickz: I fixed this problem.

plugins/ratio/ratio.php
stickz: I added a new function to migrate the configuration to the new version of the plugin.

plugins/ratio/ratio.php
```diff
@@ -17,28 +17,45 @@ class rRatio
...
+		}
 		else
 			$rt->pad();
 		return($rt);
 	}
+	private function migrate()
+	{
+		for($i=0; $i<count($this->rat); $i++)
+		{
+			$this->rat[$i]["upload"] /= 1000;
```
Novik, line 45: 1024 maybe?

Novik: For example, we had 1024 in the upload.
In the past we sent to rtorrent value 1024 * 1024 * 1024 = 1073741824
Now we will send 1024 / 1000 * 1024 * 1024 * 1024 = 1099511630

stickz: Okay, I fixed it.

PR #2277, opened 2022-01-23
Reduce 'trafic' memory footprint

stickz:
This pull request reduces the memory footprint of the 'trafic' plugin. It only updates the ratio stats columns on the WebUI once during startup and when they change. `theWebUI.ratiosStat` contains a lot of data that was being accessed unnecessarily.

theRequestManager is not a good option anymore. We would have an extra check plus function call overhead for every torrent.

### DIFF
--- a/plugins/trafic/init.js
+++ b/plugins/trafic/init.js
@@ -361,9 +361,12 @@ if(plugin.canChangeColumns() && plugin.collectStatForTorrents)
 			var rStat = this.ratiosStat;
 			$.each(data.torrents, function(hash,torrent)
 			{
-				torrent.ratioday = rStat[hash][0]/torrent.size;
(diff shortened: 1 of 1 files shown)


plugins/trafic/init.js
```diff
@@ -357,8 +349,32 @@ if(plugin.canChangeColumns() && plugin.collectStatForTorrents)
...
+		plugin.ratioChanged = true;
 		window.setTimeout( plugin.startRatios, plugin.updateInterval*60000 );
 	}
+	
+	plugin.addTorrents = theWebUI.addTorrents;
+	theWebUI.addTorrents = function(data)
+	{
+		if(plugin.ratioChanged)
+		{
+			var rStat = this.ratiosStat;
```
Novik, line 361: Why you need additional variable here? Why you can't use ```theWebUI.ratiosStat``` directly?

PR #2287, opened 2022-02-02
Friendly support for other autoloaders

nelu:
Regarding https://github.com/Novik/ruTorrent/pull/2253

`A file check must be done first, or restrict the namespace, or suppress error with @require_once , since this is the first autoloader registered when getplugins.php is called, a single plugin cannot register its autoloader when used in plugindir init.php, because this one tries to include a php file from php/utility, which will throw an error`

### DIFF
--- a/php/util.php
+++ b/php/util.php
@@ -11,8 +11,12 @@ spl_autoload_register(function ($class)
 	// Important for compatibility with 3rd party plugins
 	$arr = explode('\\',$class);
 	$class = end($arr);
-	
-	require_once 'utility/'. strtolower($class). '.php';
+	$file = __DIR__.'/utility/'. strtolower($class). '.php';
+
+	if(file_exists($file))
+    {
+        require_once $file;
+    }
 });
 
 // Fixes quotations if php verison is less than 5.4


stickz:
I think we should just suppress this error. I don't think it's necessary to check if the file exists every time we include something.

@Novik What do you think about suppressing this error?

Novik:
I have no idea, why someone may want to include their own autoloader in his plugin.
In any case we should use 
1) ```include_once``` instead ```require_once```
2) ```is_file && is_readable``` instead ```file_exists```

stickz:
I committed https://github.com/Novik/ruTorrent/commit/3becce06c5ec3be40e97917e0362a0ffd525377c to master to throw a warning instead of an error. I agree with Novik this is better than suppressing the error. This is just a temporary fix for now, until this PR is done and tested.

nelu:
> I committed [3becce0](https://github.com/Novik/ruTorrent/commit/3becce06c5ec3be40e97917e0362a0ffd525377c) to master to throw a warning instead of an error. I agree with Novik this is better than suppressing the error. This is just a temporary fix for now, until this PR is done and tested.

This is what we are trying to avoid here, your autoloader should not block the code execution in any way (unhandled error, exceptions, etc) -- thats not the purpose of it. 
The current implementation still, will break the registration other autoloaders called after yours. See this - https://github.com/nelu/rutorrent-filemanager/commit/c4edd122eaa06f7aa14fd89f040460e283672562

stickz:
Novik has a really good point about the senselessness of using an autoloader inside a plugin. I am personally very hesitant to support such an extreme case. It's up to him if he merges this PR, once you make the requested changes. I won't be doing so because it adds too much overhead to ruTorrent. We shouldn't need to look if files exist before including them.

I did fix this problem of yours though. It no longer blocks code execution. It just displays a warning saying you shouldn't be doing that. A warning is **_NOT_** an error. There is a huge difference, as it does allow code execution to continue.

nelu:
> ... I won't be doing so because it adds too much overhead to ruTorrent. We shouldn't need to look if files exist before including them....
>
The overhead is neglectable, since the autoloading function is called only when php sees a class which is not already loaded (called once for each existing php class file).
PHP will call each autoloading function in the order they were registered, if the previous autoloader did not manage to find & load a requested php class, the next autoloader function will be called

The above is how autoloader works in php

The only way a plugin can create overhead (on file_exists or @ supress), is by trying to use a non existing class, in that case the developer who is working on that plugin needs to check the class names he is trying to use, since it wont be able to find it

stickz:
Alright, I pushed anther change to master. https://github.com/Novik/ruTorrent/commit/43cd1d967887e451e510cff619c90408e1ac6cea

If the user sets `$do_diagnostic = false; ` in `conf/config.php`, it will suppress the include warnings. We don't plan on disabling diagnostics by default though because it's a very useful feature to detect configuration issues.

PR #2298, opened 2022-02-08
I Translated new strings to Turkish.

babico:
I just translated big part of new things but I couldn't find Turkish meaning some of them.


lang/tr.js
```diff
@@ -181,102 +182,102 @@ var theUILang =
...
+ doRandomizeHash : "Çapraz seedleme için info hashleri rastgele düzenle",
+ Recreate_files : "Sıradaki Dosyaları Yeniden Oluşturma",
+ Address			: "Adres",
+ PeerDL				: "Eş DL",
+ PeerDownloaded			: "İndirmiş Eşler",
+ Save_path : "Yolu kaydet",
+ registerMagnet : "Başlangıçta mıknatıs protokolü işleyicisini kaydetmeyi deneyin",
+ linkTorTorrentRestored : "rTorrent ile bağlantı kuruldu.",
+ removeAllTegs : "Tüm etiketleri kaldır",
+ Torrentler: "Torrent(ler)",
```
mateon1, line 280: This seems wrong - renamed the key instead of just the string.

Novik: Fixed, thanks.

PR #2388, opened 2022-09-23
Support additional parameters in addtorrent.php

static53:
Adds additional parameter support to addtorrent.php. This is useful for third party tools which interact with addtorrent.php rather than the XMLRPC api directly.

### DIFF
--- a/php/addtorrent.php
+++ b/php/addtorrent.php
@@ -31,6 +31,9 @@ else
 		if((strlen($dir_edit)>0) && !rTorrentSettings::get()->correctDirectory($dir_edit))
 			$uploaded_files = array( array( 'status' => "FailedDirectory" ) );
 	}
+	$addition = array();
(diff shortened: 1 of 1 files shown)


php/addtorrent.php
```diff
@@ -31,6 +31,9 @@
 		if((strlen($dir_edit)>0) && !rTorrentSettings::get()->correctDirectory($dir_edit))
 			$uploaded_files = array( array( 'status' => "FailedDirectory" ) );
 	}
+	$addition = array();
```
stickz, line 34: Please initialize this as `null` to match the function prototype. We can sneak this into ruTorrent v4.0 stable if completed.

`static public function sendTorrent($fname, $isStart, $isAddPath, $directory, $label, $saveTorrent, $isFast, $isNew = true, $addition = null)`

PR #2396, opened 2022-10-22
Use Ipv4 address for port checker

dylansama:
Pass `--ipv4` parameter to curl to use Ipv4 address, when checking for ports. The port checker websites do not support Ipv6.


dylansama:
Technically the `preferIpv4` feature could be expanded to http portion of Snoopy client as well, since curl options are exclusively for https requests. It seems the only major changes would be in the `connect()` function, when creating a socket context you would bind to `0:0` which will only match ipv4 addresses, whereas `[0]:0` (or possibly `[::]:0`) will match an ipv6 address per docs.

If there is any desire for that as well. I see very few things nowadays not using https though.

dylansama:
Fixes https://github.com/Novik/ruTorrent/issues/2402

dylansama:
Added support for preferIpv4 on plain http requests as well. Switching to the more diverse function `stream_socket_client` over `fsockopen` for all requests. Guessing the function check for `stream_socket_client` was from very ancient PHP 4.x days. As noted before simply need to set `bindTo` option to `0:0` rather than letting system choose.

plugins/check_port/conf.php
```diff
@@ -5,3 +5,5 @@
 				// false - disable check_port plugin
 				// "yougetsignal" - use https://www.yougetsignal.com/tools/open-ports/
 				// "portchecker" - use https://portchecker.co/
+
+$preferIpv4 = true;		// prefer ipv4 addresses when checking ports
```
stickz, line 9: If you change this to false, I will merge the pull request before releasing 4.0 stable.

We can discuss enabling this feature by default a later date. I would like to get your contribution added to ruTorrent.

php/Snoopy.class.inc
```diff
@@ -654,9 +658,17 @@ class Snoopy
 			$port = $this->port;
 		}
 		$this->status = 0;
-		if(function_exists("stream_socket_client") && !is_null($this->IP))
```
stickz, line 657: We can't create the stream unless `$this->IP` is not `null`.

php/Snoopy.class.inc
```diff
@@ -654,9 +658,17 @@ class Snoopy
...
+		if(function_exists("stream_socket_client"))
 		{
-			$opts = array('socket' => array('bindto' => $this->IP.':0'));
+			$opts = array('socket' => array());
+			if(!is_null($this->IP))
+			{
+				$opts['socket']['bindto'] = $this->IP.':0';
+			}
+			elseif($this->preferIpv4)
```
stickz, line 668: We can go `if($this->preferIpv4)` `$opts['socket']['bindto'] = '0:0';` here.
Then `else` `$opts['socket']['bindto'] = $this->IP.':0';`

Then check further upward: `if(function_exists("stream_socket_client") && !is_null($this->IP))`

php/Snoopy.class.inc
```diff
@@ -654,9 +658,17 @@ class Snoopy
 			$port = $this->port;
 		}
 		$this->status = 0;
-		if(function_exists("stream_socket_client") && !is_null($this->IP))
+		if(function_exists("stream_socket_client"))
 		{
-			$opts = array('socket' => array('bindto' => $this->IP.':0'));
+			$opts = array('socket' => array());
```
stickz, line 663: PHP 8 allows this value to be null. We should ensure this value is initialized or not create the stream at all.

stickz:
Added Work In Progress label. Looking for changes and to disable by default to reduce the chances of a regression. 

Anther pull request will be required with improvements to enable the feature by default. `Snoppy` we need to be improved. `0:0` is not a good way to bind to the IPV4 address. This will create a regression if a static IP address is assigned for torrents. Please don't address this issue in the current pull request. We want to merge a more minimal set of changes first into `ruTorrent`.

plugins/check_port/conf.php
stickz: Also, can you change the variable names and the comments? We're not preferring IPV4 here. We're looking up the IPV4 address. We need to make this clear, so the client knows what they're enabling.

stickz:
Target changed to version 4.1 stable. Version 4.0 stable is being released next week.

php/Snoopy.class.inc
dylansama: Wouldn't you want `$this->IP` to be trump all other settings? Since its a manually specified setting? Just a little confused on the branching order you are talking about here.

php/Snoopy.class.inc
dylansama: So it should be?

```
if(function_exists("stream_socket_client") && !is_null($this->IP))
{
  // same as previous, no change
  $opts = array('socket' => array('bindto' => $this->IP.':0'));
  $context = stream_context_create($opts);
  $fp = stream_socket_client("tcp://".$host.":".$port, $errno, $errstr, $this->_fp_timeout, STREAM_CLIENT_CONNECT, $context);
}
elseif(function_exists("stream_socket_client") && $this->useIpv4)
{
  // new section
  $opts = array('socket' => array('bindto' => $this->'0:0'));
  $context = stream_context_create($opts);
  $fp = stream_socket_client("tcp://".$host.":".$port, $errno, $errstr, $this->_fp_timeout, STREAM_CLIENT_CONNECT, $context);
}
else
  // same as previous, no change
  $fp = fsockopen($host,$port,$errno,$errstr,$this->_fp_timeout);
```

I had thought `stream_socket_client` was functionally identical to `fsockopen`, it just allowed for specifying additional options with stream_context_create. So it made sense to just group everything up, and prefer the newer method.

php/Snoopy.class.inc
dylansama: You mean we can't have an empty array? $this->IP is already checked against null. I was testing with PHP 5.6 and it seemed to be fine as well.

plugins/check_port/conf.php
dylansama: Renamed to `$useIpv4`, or should it be more explicit? Like `$useIpv4Interface` ?

dylansama:
> Added Work In Progress label. Looking for changes and to disable by default to reduce the chances of a regression.
> 
> Anther pull request will be required with improvements to enable the feature by default. `Snoppy` we need to be improved. `0:0` is not a good way to bind to the IPV4 address. This will create a regression if a static IP address is assigned for torrents. Please don't address this issue in the current pull request. We want to merge a more minimal set of changes first into `ruTorrent`.

Yeah perhaps we could just skip the feature altogether for insecure `http` requests? I was just toying around to make it "feature complete' for Snoopy client between both http and https. Specifically this PR is really just for fixing the check_port plugin and both of the sites in the plugin use **only** https anyways. So all the stream context binding stuff is really un-used, the only truly useful change is the curl paramter `--ipv4`.

php/Snoopy.class.inc
dylansama: From what I found from testing, and perusing some of the source code if `bindto` is not specified it should fall back to the "let system choose" methodology. Which is functionally identical to what `fsockopen` was doing previously. I'm not finding any issues in PHP 7 either though if its unspecified.

Perhaps my code sample comment in other section makes most sense though. It will leave all codepaths identical to previous, only change is if `$useIpv4` is enabled. Just want to confirm before making additional changes.

stickz:
@dylansama I edited your pull request and squash merged the changes into version 4.0 stable. I wanted to show you how to create an effective pull request. The useIpv4 feature is enabled by default. There is no risk of regression. 

If you would like to make additional improvements, please create a new pull request.

PR #2445, opened 2023-01-29
Fix regression with 'extratio' plugin

### LINKED ISSUE #2443: Plugin 'extratio' does not works correctly anymore

- [X] Web browser cache cleared
- [X] Link provided to install script if applicable
- [X] Web browser, ruTorrent, PHP and OS version provided

N/A

N/A

### Describe the bug

Because method ```rTorrentStub.prototype.setlabelResponse``` is not called after recent changes in the ```js/rtorrent.js```.
This is because we call ```action=setlabel``` with a ```&list=1```.
Here - https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L1277 we override action from ```setlabel``` to ```list``` and ```getResponse``` is not called for action ```setlabel```...

### Steps to reproduce

Set debug output at https://github.com/Novik/ruTorrent/blob/master/plugins/extratio/init.js#L254 and try to set label for torrent.

### Expected behavior

_No response_

### Additional context

For quick fix we may insert something like
```
stub.getResponse(data);
```
at https://github.com/Novik/ruTorrent/blob/master/js/rtorrent.js#L1277 before ```Ajax``` call, but I have doubts about this method.

Novik:
closes #2443

### DIFF
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -1272,7 +1272,6 @@ function Ajax(URI, isASync, onComplete, onTimeout, onError, reqTimeout, partialD
 		Ajax_UpdateTime(jqXHR);
 		
 		stub.logErrorMessages();
-		var pending = !stub.isError() && stub.commandOffset < stub.commands.length;
 		if(!stub.isError()) {
 			var responseText = stub.getResponse(data);
 			if (partialData) {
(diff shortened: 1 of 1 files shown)


stickz:
Looks good to me. I just removed an unnecessary variable that was going to create memory leaks in the web browser. I also renamed the title of the pull request to document this change and used the "closes" tag to close the issue report on merge.

PR #2449, opened 2023-02-04
Fixed speed graph width & other changes

koblack:
adjust width of the columns so values like e.g. "1023.33 MiB" can fit
update Polish translation 
update default user agent
remove some dead sites


stickz:
Looks good to me. Merging the pull request.

Just remember, that you can never create too many pull requests. I would prefer 4 separate pull requests than have changes combined into one. It's much easier to document and track changes. If there's a problem, it's also easier to fix.

PR #2453, opened 2023-02-20
Some YggTorrent fixes

Arakmar:
The regex change fixes search results. (now that Cloudflare has been disabled...)
Some new categories have also been added


stickz:
Could you review and merge this @Novik ? I'm not familiar with search engines on ruTorrent.

PR #2503, opened 2023-05-03
Add pt-PT translations

Frias:
Split the Portuguese translation into pt-PT and pt-BR, used Chinese structure as precedent (didn't use the unofficial pt and pb abbreviations).
Historically the Portuguese translation in the repo is Brazilian Portuguese. It's difficult for me to contribute to the Brazilian version and I didn't want to destroy the good work already done by someone else - that's why I split them. pt-BR translation was left untouched. It's technically the same language and both would be correct but the most common terms used differ in most contexts.
Then I completed the pt-PT translation including the plugins. Tried to be as consistent as I could, some fixes may come later with use.

Also added the macos .DS_Store files to .gitignore - in a separate commit.


stickz (approved):

stickz:
LGTM!

PR #2536, opened 2023-07-22
PHP 5.x fixes

dylansama:
Found 2 small problems when running with PHP 5.6. I only found [one particular note](https://github.com/Novik/ruTorrent/wiki) about ruTorrent requiring only PHP5. If that has changed ignore this PR I suppose.

The RSS plugin had a big merge which incorporated the usage of `??`. This operator was newly added in PHP7. I switched this to the older style ternary operator `?:`.

There also seemed to be a problem (likely specific to PHP5 versions) which made the 2 yield expressions somewhat ambiguous to the parser. Adding parenthesis around them seems to allow them to be understood. RSS plugin now working as expected.

### DIFF
--- a/plugins/rss/rss.php
+++ b/plugins/rss/rss.php
@@ -242,7 +242,7 @@ class rRSS
 				}
 				$item = [
 					'title'=> $xText('title', $i),
-					'timestamp'=>strtotime($xText('updated', $i)) ?? false,
+					'timestamp'=>strtotime($xText('updated', $i)) ?: false,
 					'link'=> $urlPrefix.$xText('link/@href', $i),
 					'description'=> join("\n\n", $description),
 				];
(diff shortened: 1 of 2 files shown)


stickz:
Thanks for the pull request, however we do not wish to support PHP 5. Any code that support less than PHP 7 can be removed. Documentation can be updated to reflect this. 

Eventually, we'll be dropping support for PHP 7, but we can't do this yet. Too many platforms are still running it. It's at end of life already for security updates. ruTorrent is confirmed to work on PHP 8+. We resolved all known bugs with PHP 8+ in `v4.1`.

dylansama:
Time to start working on compiling PHP8 😅

Yeah I haven't done a pull in quite a while, and saw all the updates. Good to know!

stickz:
Alright sounds good. I'm going to close this off. If you find any issues with PHP8, feel free to submit a pull request.

Issue #2509, opened 2023-05-13
Retrackers Plugin "Remove Announce URLs" feature not working


AncientVamp:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Link provided to install script if applicable
- [X] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Firefox 113.0.1
ruTorrent 4.1.5
PHP 7.4.33
OS Debian 11

### Tell us how you installed ruTorrent

Swizzin

### Describe the bug

retrackers "Remove Announce URLs" feature is not working. Also when you add a torrent file that contains a tracker url you added in retracker to be removed, the torrent is being added in a stop state now. 

### Steps to reproduce

Go to retrackers plugin options, add a url you want to remove in Remove Announce Urls Box. Add a torrent file containing the tracker url you wanted removed

### Expected behavior

retracker would check each torrent file added for the url i want remove and then remove them. The torrent would begin downloading afterwards.

### Additional context

This plugin was last working fine for me when I was on rutorrent v4.1

stickz:
There were no changes between of the period of `v4.1` and `v4.1.5` which could have impacted this plugin.

Could you try something? `nano /srv/rutorrent/plugins/retrackers/run.sh` Remove the last character `&`. Save and retry.
https://github.com/Novik/ruTorrent/blob/48933089370b667e8d3b19fb8f9b92bc6db5409d/plugins/retrackers/run.sh#L3

AncientVamp:
Just tried that. Having the same issue. I could have been mistaken about rutorrent version it was last working for me. I know it was sometime in April I noticed the retrackers plugin last working for me. Could have been running rutorrent v4.0.4.

AncientVamp:
I took the retrackers plugin from v4.04 tree and installed it on rutorrent v4.1.5. It's working correctly now.

stickz:
There have been no changes to `retrackers` since `v4.0.4` except for a translations file and version bump. 

I highly suspect it's that `&` symbol causing a race condition. This tells the bash job to run in the background asynchronously against other tasks. If this is the case, it will reoccur again at random. The recommended course of action now that everything is working properly is to remove it.

Could you let me know if everything works correctly with that symbol removed? If so, I will release a hotfix tomorrow. There's no valid reason to have that symbol there. I think it was a developer error to silence output.

AncientVamp:
[...]

PR #2541, opened 2023-07-30
Fix retracker plugin update.php

dguskovv:
Fix retracker plugin

Plugin check each URL in the tracker list of the torrent.
Also when you add a torrent file that contains a tracker url you added in retracker to be removed, the torrent is being added in a start state.


stickz:
Thanks. closes #2509

PR #2559, opened 2023-08-26
Remove file filtering from .torrent selection

anthonyryan1:
This mitigates a bug in Chrome 116+ on Android 11+ where it's accepts mime types but not file extensions in the accept property. This should improve file selection support overall.


koblack:
Friendly note, this reverts commit "For IOS compatibility" 6da51e0
I don't have iOS but this seems problematic.

anthonyryan1:
Lets delay merging this for a bit, I'll see about getting some iOS testing done to ensure we're not causing further regressions.

anthonyryan1:
Confirmed the bug still exists on iOS. Proposing that we document the behaviour and stop using the accept attribute, because there is no valid combination that will work on all major operating systems.

koblack:
Thank you for fixing this.
Maybe the way qBittorrent chose is better?
```html
            <input type="file" id="fileselect" accept=".torrent, application/x-bittorrent" name="fileselect[]" multiple />
``` 
```javascript
        if (Browser.platform === 'ios') {
            $('fileselect').accept = ".torrent";
        }
``` 
REF: https://github.com/qbittorrent/qBittorrent/blob/master/src/webui/www/private/upload.html

anthonyryan1:
I have no objection to doing it that way if the maintainers prefer. We do have a precedent in the codebase with other user agent sniffing in [browserDetect()](https://github.com/Novik/ruTorrent/blob/v4.2.4/js/common.js#L16-L79) and a few other places.

I was just trying not to introduce more browser sniffing, because browser sniffing fixes, while good at introduction often bitrot into a source of bugs after a few years. Since browsers and operating systems often fix the bugs they were mitigating.

If we could do feature detection instead, looking for mime type support instead of a platform or browser, then I'm 100% on board.

stickz:
Changes should go to the `v4.2` branch. We're almost finished with `v4.1`.

> I have no objection to doing it that way if the maintainers prefer. We do have a precedent in the codebase with other user agent sniffing in [browserDetect()](https://github.com/Novik/ruTorrent/blob/v4.2.4/js/common.js#L16-L79) and a few other places.

Yes, let's use the browser detect method for this change. We should set the value of `accept` based on the web browser.

> I was just trying not to introduce more browser sniffing, because browser sniffing fixes, while good at introduction often bitrot into a source of bugs after a few years. Since browsers and operating systems often fix the bugs they were mitigating.

I'm okay with mitigating bugs for `accept`.  This is simply a file type filter. We can disable it on unsupported platforms.

> Friendly note, this reverts commit "For IOS compatibility" [6da51e0](https://github.com/Novik/ruTorrent/commit/6da51e0bf21066216823a94b363c0c3f9a31bde5) I don't have iOS but this seems problematic.

This is problematic. We should only remove this value for iOS, so we don't impact other platforms.

PR #2617, opened 2023-12-14
Fix tabbar positioning

xombiemp:
Before:
before

After
after


stickz:
Which web browser are you using? This is how google chrome looks for me. 
ruTorrent-Log-Text

xombiemp:
Hmm, it's in Safari 17.2. You're right, it does look normal in Chrome and Firefox without this change. Dang Safari!

stickz:
> Hmm, it's in Safari 17.2. You're right, it does look normal in Chrome and Firefox without this change. Dang Safari!

You can still resolve this problem. It's relatively simple. There is an example in the file to check if the web browser equals Safari.

Afterwards, we use the JQuery selector to set a css style `#tabbar` and `#tdcont`. This will override `style.css` for Safari.

PR #2634, opened 2024-01-23
fix DE translations for "alternating list background"

bogenpirat:
the current translation means "alternative background" as opposed to the likely intended "alternating background". in german, the latter would be "Alternierender Listenhintergrund", but that is clunky phrasing. the change in this commit translates to "alternating line backgrounds" and is verbiage common to e.g. Excel.


stickz:
LGTM Thanks! There's a bunch of other German translations missing from that file as well. If you want to update, I will merge!

PR #2635, opened 2024-01-28
Update pt-PT translation

Frias:
Add the translation for the 'views panel' and corrected some typos in the pt-PT translation of some plugins


stickz:
Thanks!

PR #2663, opened 2024-05-27
added a copy button to console window

jevenski:
Added a copy button to console window, making copying console log like mediainfo result a little bit more confortable, without having to select the entire text and hit Ctrl+C. Log a message and nothing else when the console log is empty, e.g. when screenshots are being generated.

console_log


stickz:
Hello, thanks for the contribution. Can you submit this to the `develop` branch instead? We will release this with `ruTorrent v4.4`.

PR #2665, opened 2024-05-27
added a copy button to console log window

jevenski:
Added a copy button to console window, making copying console log like mediainfo result a little bit more confortable, without having to select the entire text and hit Ctrl+C. Log a message and nothing else when the console log is empty, e.g. when screenshots are being generated.

console_log

### DIFF
--- a/plugins/_task/init.js
+++ b/plugins/_task/init.js
@@ -116,10 +116,23 @@ plugin.fromBackground = function( no )
 	plugin.onStart(this.foreground); 
 }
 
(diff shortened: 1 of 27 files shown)


jevenski:
Closed the [old request](https://github.com/Novik/ruTorrent/pull/2663) and created this new pull request. Made another pull request in between, which seemed to be causing quite a mess. Apologies if causing any trouble. I'm relatively new to git and still familiarizing myself with it.

PR #2667, opened 2024-05-29
Prepend "Save" button after spectrograms are generated

jevenski:
A "save" button is being added on the loading of the spectrogram plugin, which I believe is not the way it should behave in. 

It is there when any task is being carried out, when there is nothing to save. It is there when a new torrent is created, rendering two "save" buttons in the console windows, with another one being the real "save" button to download the torrent file.

save-spectrogram
create-torrent

### DIFF
--- a/plugins/spectrogram/init.js
+++ b/plugins/spectrogram/init.js
@@ -78,6 +78,14 @@ if(plugin.canChangeMenu())
 					plugin.setConsoleSize(this);
 				});
 				$("#soxtaskno").val(task.no);
+        $('#tsk_btns').prepend(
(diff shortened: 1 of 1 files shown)


Joshseibert:
yes

stickz:
Is this a bug fix for the spectrogram plugin? If so, could you provide a screenshot how it looks after this fix is applied?

jevenski:
Here are a few screenshots of how dialog buttons look like.

1. Spectrogram running (without a "Save" button):
spectrogram-running

2. Spectrogram finished (a "Save" button is prepended):
spectrogram-finished

3. Make torrent running (without a "Save" button):
make-torrent-running

4. Make torrent finished (only one "Save" button is prepended, which is the correct one to download the .torrent file):
make-torrent-done

PR #2673, opened 2024-06-14
Applied the flex layout to the add torrent dialog

jevenski:
Restructured the add torrent dialog using the CSS flex layout. Adjusted the CSS in MaterialDesign theme a little bit. 

And added another "Return" button next to the label text box, for returning to the label dropdown selection.

return-to-select

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 .project
 .jsdtscope
 
+# VSCode
+.vscode
+
 # ruTorrent
 conf/users/*
 share/users/*
--- a/css/style.css
+++ b/css/style.css
@@ -101,6 +101,7 @@ Input.TextboxShort {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; w
 Input.TextboxMid {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; width: 110px}
 Input.TextboxLarge {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; width: 220px}
 Input.TextboxVShort {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; width: 30px}
+input.textbox-grow {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; flex: 1;}
 button, input.Button {padding: 1px 10px; background: #F0F0F0 url(../images/h.gif) repeat-x center bottom; width: 80px; height: 19px; border: 1px solid #A0A0A0; cursor: pointer }
 button[disabled],input.Button[disabled] { opacity: 0.3; cursor: default }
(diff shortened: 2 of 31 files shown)


stickz:
Add the end of `style.css`, you'll see I added some media queries for mobile design optimization. What do you think about using these to put the `SELECT`, `ADD FILE` and `ADD URL` buttons on a new line if the screen width is greater than 1280 pixels?

We could improve the usability on desktops, if we make the text boxes a little bit bigger. But keep them as shown for mobile views to make better usage of the limited screen space.

https://github.com/Novik/ruTorrent/blob/fa4e4e29c4d0fe89faac960a56a0e00175ba75f9/css/style.css#L309-L329

stickz:
Actually, it's just the `Torrent File` and `Torrent URL` text boxes that will overflow. It might be easier for the user to have the `SELECT` button **beside** the text box. Then the `ADD FILE` and `ADD URL` **bellow** the text box.

jevenski:
> What do you think about using these to put the `SELECT`, `ADD FILE` and `ADD URL` buttons on a new line if the screen width is greater than 1280 pixels?

I've been using the MaterialDesign theme for too long that I forgot how things looked like in the Standard theme. In the MaterialDesign theme, the `<label>`-`<input>`-`<button>` set is put in one line, while in the Standard theme as well as other themes, the `<button>` is pushed to a new line with a `</br>`. That's why I've put them three all in the same line.

Speaking from my own experience, I think it's ok to have them three on the same line in a desktop browser, as long as the `<input>` doesn't get too long in between. I think I need to set an upper limit to it, especially the file input will grow as long as the file name, but users almost certainly don't want to read the file name word by word.

However, pushing the submit button to a new line is a good idea for a mobile web browser. I didn't think too much about mobile users. That's my bad. I'm thinking about altering the entire layout to make a more responsive UI, but it would require too much code adjustment for now.

stickz:
@jevenski If I merge this pull request to the develop branch, can you submit anther one in the near future to improve the torrent dialog? This is polished enough to merge. We just need a little bit more progress before a release is created. 

I don't have any specific criteria for the desktop layout. Anther option is to increase the dialog size with an upper limit as specified. I agree with your thoughts about the mobile responsiveness.

jevenski:
I'm working on a patch commit to the add torrent dialog, trying to replace the flex layout with the grid layout, thus making the elements easier to justify and align on both desktop and mobile devices. Need some time testing out the CSS rules. Please wait for it before merging this pull request into the develop branch.

stickz:
Okay no problem. You can create a new commit or file a new pull request. Up to you.

jevenski:
Applied the grid layout system to the add torrent dialog [Pull Request](https://github.com/Novik/ruTorrent/pull/2674).

This old pull request is now closed.

PR #2674, opened 2024-06-17
Apply grid layout on add torrent dialog for better responsiveness

jevenski:
The grid layout system is applied to the add torrent dialog for better responsiveness, in hope of make ruTorrent mobile friendly. The concepts of break points divided by `sm` `md` `lg` `xl` and `xxl` are borrowed from the Bootstrap framework. `.col-` and `.col-offset-` rules are defined using different media queries. 

1. Tablet and desktop web browsers (`md` for 768 pixels wide and up)
Screenshot 2024-06-17 191227
Screenshot 2024-06-17 215506
Layout remains mostly the same, with improved alignment and justification. A "Select" button is added next to the label input for returning to the label dropdown selection.

2. Smaller screens (under 768 pixels wide)
Screenshot 2024-06-17 191312
Screenshot 2024-06-17 215600
Elements are arranged vertically in a column, making full use of the narrow screen. The four labels for the directory input, the label input, the torrent file input, and the torrent URL input are hidden, as the legends in the field sets and the placeholders of the text inputs explain the purpose of each element clear enough.


css/style.css
```diff
@@ -314,10 +313,12 @@ div#tadd select { width: 220px }
 
 @media only screen and (min-width: 1280px) {
   .desktop { display: inline-block; }
+	.dlg-window {max-width: 60vw;}
 }
 
 @media only screen and (max-width: 1279px) {
   .desktop { display: none; }
+	.dlg-window {max-width: 100vw;}
 }
```
stickz, line 322: `min-width: 1280px` is the bigger screen and `max-width: 1279px` is the smaller screen. What is the reason these are inversed? Also there is a problem with line spacing.

css/style.css
```diff
@@ -327,3 +328,75 @@ div#tadd select { width: 220px }
 @media only screen and (max-width: 729px) {
   .mobile-search { display: none; }
 }
+
+/* X-Small devices (portrait phones, less than 576px) */
+/* No media query for `xs` since this is the default in Bootstrap */
+/* Break point settings borrowed from Bootstrap framework */
```
stickz, line 334: What do you think about adding Bootstrap to ruTorrent? This is the preferred solution over importing settings to `style.css`. Modular programming practices are very important for this project.

lang/en.js
```diff
@@ -140,6 +140,10 @@ var theUILang =
  time_m				: "m ", /*for x minutes remaining*/
  time_s				: "s ", /*for x seconds remaining*/
  Base_directory			: "Directory",
+ Torrent_options		: "Torrent Options",
+ Return_select_label		: "Select",
+ Add_from_file			: "Add From File",
+ Add_from_URL				: "Add From URL",
```
stickz, line 146: Could you resolve the spacing issues with `Add_from_URL` in the language files?

stickz (changes requested):

lang/en.js
jevenski: Yeah, sure. They appeared to be in a line in VSCode, but jumped to right in GitHub. That's confusing.

css/style.css
jevenski: Yeah, I think it's a good idea. I thought jQuery was the only dependency library that ruTorrent requires, making it a light-weighted web app. I was not sure about bringing another dependency into ruTorrent, so I reinvented the wheel. (Maybe just a tiny bit of the wheel, haha).

jevenski: Or maybe we could consider TailwindCSS, as it focuses on utility classes for highly customizable designs. We already have a very mature user interface already. We just need a more tailored polish for both all kinds of devices.

lang/en.js
stickz: All you need to do is backspace one of the tabs.

css/style.css
stickz: Okay, what about something like this instead? Over 1000 pixels is too large for 100% of the viewport.
```
.dlg-window {max-width: 60vw;}

@media only screen and (max-width: 729px) {
  .mobile-search { display: none; }
  .dlg-window {max-width: 100vw;}
}
```
If required, you can add anther media query for 80%. Depending on usability, I might actually recommend this.
```
@media only screen and (min-width: 730px) and (max-width: 1279px) {
  .dlg-window {max-width: 80vw;}
}
```

css/style.css
jevenski: Should we use Bootstrap 4, which has reached end of life, or the latest version of Bootstrap?

stickz: Good catch, let's use the latest version of Bootstrap. It will need to be distributed with the project like jQuery.

jevenski: Yeah, sure. I've been doing tests with the latest Bootstrap version 5.3.3. 

At first, I tried to link the entire Bootstrap library, and it seemed to interfere with the custom CSS classes defined in `/css/style.css`. The left side-panel became very crowded, and the tables became vertically unaligned between table heads and table bodies. The huge leap forward requires a lot of fine-tuning of the entire ruTorrent project.

So I tried to minimize the impact by using `bootstrap-grid.min.css` instead, as we are mainly working on importing the responsive grid system from Bootstrap. The effect is smooth. The modification on the add torrent dialog is mostly done. I will push a patch commit to this pull request in the coming days after doing more tests.

css/style.css
jevenski: Actually, I moved this `max-width` rule to the end of the `style.css` file, where there are media queries for different screen sizes, to keep consistency with the bootstrap break points.

css/style.css
```diff
@@ -327,3 +327,36 @@ div#tadd select { width: 220px }
...
+
+/* X-Large devices (large desktops, 1200px and up) */
+@media (min-width: 1200px) { 
+  /* Custom rules for x-large devices */
+}
+
+/* XX-Large devices (larger desktops, 1400px and up) */
+@media (min-width: 1400px) { 
+  /* Custom rules for xx-large devices */
+}
```
stickz, line 362: Please test this custom values here and let me know what you think. They expand the size of the text boxes and dialog window, so the user can see more of the text being inputted on larger screens.

```
.dlg-window {max-width: 95vw;}

/* Medium devices (tablets, 768px and up) */
@media (min-width: 768px) { 
  /* Custom rules for medium devices */
  .dlg-window {min-width: 525px;}
  div#tadd .col-md-6 {min-width: 300px;}
  div#tadd .col-md-3 {max-width: 100px;}
}

/* Large devices (desktops, 992px and up) */
@media (min-width: 992px) { 
  /* Custom rules for large devices */
  .dlg-window {min-width: 675px;}
  div#tadd .col-md-6 {min-width: 450px;}
  div#tadd .col-md-3 {max-width: 100px;}
}
```

jevenski: I've done the test, and here's the screenshot for the `md` sized screen. 
Screenshot 2024-06-22 173731

The four checkboxes below the directory input become unaligned with the directory input, as the `offset-md-3` uses the `margin-left` variable to implement the offset.

I think we can leave the `min-width` control on our dialog windows, to make better use of the wider display areas. However, on the columns inside a container, I think it's better to let Bootstrap handle the width distribution. For one thing, in this case, we can't have the maximum offset defined in CSS. For another, if we are to have nested columns in other dialog windows, it will require additional rules to exclude those nested columns. All in all, it will become harder to maintain in the future.

```CSS
/* Medium devices (tablets, 768px and up) */
@media (min-width: 768px) { 
  /* Custom rules for medium devices */
  .dlg-window {min-width: 525px;}  /* this line stays */
  div#tadd .col-md-6 {min-width: 300px;}  /* this line goes */
  div#tadd .col-md-3 {max-width: 100px;}  /* this line goes too */
}

/* Large devices (desktops, 992px and up) */
@media (min-width: 992px) { 
  /* Custom rules for large devices */
  .dlg-window {min-width: 675px;}  /* this line stays */
  div#tadd .col-md-6 {min-width: 450px;}  /* this line goes */
  div#tadd .col-md-3 {max-width: 100px;}  /* this line goes too */
}
```

index.html
```diff
@@ -20,6 +20,7 @@
 		<link rel="icon" type="image/png" href="./images/favicon-16x16.png" sizes="16x16" />
 		<meta name="msapplication-TileColor" content="#FFFFFF" />
 		<meta name="msapplication-TileImage" content="./images/mstile-144x144.png" />
+		<link href="./css/bootstrap-grid.min.css" rel="stylesheet" type="text/css" />
```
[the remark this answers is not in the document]
stickz: @jevenski Please add cache buster and I will merge. I have drafted an improved bootstrap layout with `col-md-8`. I will submit a new pull request afterwards. Then request your review, if that is okay with you.

Issue #2679, opened 2024-06-22
Feature Request: Improved directory selection in Add Torrent dialog


albino1:
### Is your feature request related to a problem?

Currently when you try to change a directory in the Add Torrent dialog you get a long list of subfolders with a tiny box and only 8 or 9 actually visible and no real way to quickly navigate them.  There's no standard browser style file selector, and typing to jump down the list only works for each individual letter press.  For example, if you have a subfolder that begins with `star` and you type `s` it jumps to the folders starting with `s`, but when you then hit `t` instead of going to `st` it now jumps down to folders starting with `t`, followed by up to folders starting with `a` and then back down to `r`.

If you only have a few subfolders this isn't a huge deal, but if you have hundreds or thousands it can be a real pain to navigate the tiny box, especially because so few are displayed at any one time.

### Describe the solution you'd like

The solution I would like to see is either a fully implemented standard file selection box, or at a minimum, the ability to type quickly subsequent letters without jumping up and down the subfolder list on each key press.  Maybe even just an autofill on the line if that's simple enough.

Thanks.

stickz:
Hi @albino1. We are working on re-doing the add torrent dialog in #2674. Once completed, I plan to make the textbook bigger, so you can type in the file path manually. Will this address your problem?

albino1:
Certainly a bigger textbox would help as currently you can only see a handful of rows, which makes it difficult and slow to navigate with the mouse wheel, arrow keys, or by dragging the slider.  I do think even something as simple as having the ability to type out `term` and get to folder starting with `term`, instead of ending up on `m`, would go a long ways towards making it more usable though.

stickz:
Completed in version 5 released.

PR #2680, opened 2024-06-23
Add a search bar to _getdir list

jevenski:
I added a search bar to the top of the directory list in`_getdir` plugin. This was inspired by this [Feature Request](https://github.com/Novik/ruTorrent/issues/2679).

HTML structure was altered by the way, from `table` to `div` with flex layout. Type a few letters to filter directories, and press `Esc` key to clear search result. 

Recording 2024-06-23 224517


plugins/_getdir/utils.js
```diff
@@ -0,0 +1,20 @@
...
+	dirs = document.querySelector(".rmenuobj").querySelectorAll("div");
+	for (let i = 0; i < dirs.length; i++) {
+		dir = dirs[i]
+    if (dir) {
+      txtValue = dir.textContent || dir.innerText;
+      if (txtValue.toUpperCase().indexOf(keyword) > -1) {
+        dirs[i].style.display = "";
+      } else {
+        dirs[i].style.display = "none";
+      }
```
stickz, line 15: A ternary operator can be used here for single variable assignment.
`dirs[i].style.display = txtValue.toUpperCase().indexOf(keyword) > -1 ? "" : "none";`

Also, please address spacing issues.

stickz (changes requested):

plugins/_getdir/_getdir.css
```diff
@@ -1 +1,11 @@
+html {height: 100%;}
+body {background-color: window; color: windowtext; border: 0px; margin: 0px; padding: 0px; -moz-user-select:none; overflow: hidden; height: 100%;}
+.dir-list {display: flex; flex-direction: column; align-items: stretch; height: 100%;}
+.dir-list div { padding-top: 1px; padding-bottom: 1px; padding-left: 0px; padding-right: 0px; cursor:default; font-size: 11px; font-family: Tahoma, Arial, Helvetica, sans-serif; }
+.search-bar {display: flex; flex-direction: row;}
+.search-bar input[type="text"] {flex-grow: 1;}
+.rmenuobj {margin-bottom: 5px; border-width: 0; overflow-y: auto;}
+.rmenuitem { color: windowtext; }
+.rmenuitemselected { color: highlighttext; background-color: highlight; }
+
```
stickz, line 10: Please resolve the text overflow of menu items in the directory list. We do not want to create a new line.

stickz (changes requested):

plugins/_getdir/_getdir.css
jevenski: Good catch! The directory name wraps itself when it's longer than the width of the directory list. 
Screenshot 2024-06-24 110813

I think a single `white-space: nowrap;` rule will do the magic. A horizontal scrollbar appears, if users want to see the full long long directory name.
Screenshot 2024-06-24 111248

plugins/_getdir/_getdir.css
```diff
@@ -1 +1,20 @@
...
+  font-family: Tahoma, Arial, Helvetica, sans-serif;
+  white-space: nowrap;
+}
+.search-bar {display: flex; flex-direction: row;}
+.search-bar input[type="text"] {flex-grow: 1;}
+.rmenuobj {margin-bottom: 5px; border-width: 0; overflow-y: auto;}
+.rmenuitem { color: windowtext; }
+.rmenuitemselected { color: highlighttext; background-color: highlight; }
+
 input.browseButton {width:30px}
```
stickz, line 20: New line missing at end of file.

plugins/_getdir/getfiles.php
```diff
@@ -195,40 +188,42 @@ function menuDblClickAndExit(obj)
...
+				}		
+				echo "&nbsp;&nbsp;";
+				echo $key;
+				echo "</div>";
+			}
+			?>
+		</div>
+	</div>
+	<script type="text/javascript" src="./utils.js?v=430"></script>
 </body>
```
stickz, line 229: New line missing at end of file.

plugins/_getdir/getdirs.php
```diff
@@ -157,38 +150,40 @@ function menuDblClickAndExit(obj)
...
-	else
-		echo "<tr><td code='".rawurlencode($data)."' class='rmenuitem' id='i".$chr."' nowrap onclick='menuClick(this); return false;' ondblclick='menuDblClick(this); return false;'>";
-	echo "&nbsp;&nbsp;";
-	echo $key;
-	echo "</td></tr>";
-}
-?>
-</table>
+	<div class="dir-list">
+		<div class="search-bar"><input type="text" class="filter-dir" placeholder="Type to filter..." /></div>
```
stickz, line 157: Can we assign an id to this input type and focus it when the directory list iframe is created? This would prevent the user from having to click onto it to start typing.

plugins/_getdir/utils.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+document.getElementById("dir-search-bar").addEventListener("keyup", filterDir);
+document.addEventListener("readystatechange", () => {
+	if (document.readyState === "complete") {
+		setTimeout(() => {
+			document.getElementById("dir-search-bar").focus()
+		}, 500);
+	}
+});
```
jevenski, line 22: Added this event listener when the iframe is in ready state. I had to use the `setTimeout` function, otherwise JavaScript wouldn't help us autofocus on the text input.

stickz: `setTimeout` can be set to `0`. Please test to confirm this works for you. Also please let a comment that `setTimeout` is required to queue the focus last in sequence, so we don't accidently remove it.

jevenski: Looks like it's a browser related issue. I've tried Firefox and Chrome, both the latest version. Even a plain `.focus()` without `setTimeout()` would work perfectly on Chrome, but not on Firefox. 

I've tried `0` and something below `100`. The autofocus functionality just worked occasionally. A timeout value of `100` seems to be a fair workaround. Our users could barely notice the time gap.

jevenski: OK, that's the way to go. I'll take some time looking into that part. It will probably need a few more days, because I've been busy with work lately.

PR #2683, opened 2024-06-27
Apply grid layout on create torrent dialog

jevenski:
Taking another step closer to making ruTorrent mobile friendly, by applying the bootstrap grid layout to the create torrent dialog. I think we still need to re-work the "URL replacement in RSS" dialog and the "Ratio Rules" dialog before the next minor update release.

1. `md` screen size and larger:
Screenshot 2024-06-27 220927

2. Smaller screens:
Screenshot 2024-06-27 220953

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -334,7 +334,7 @@ div.space {flex-grow: 1;}
 /* Custom rules for medium devices */
 div#tadd {max-width: 95vw;}
 .buttons-list {
-  margin: 0 0.5rem 0.5rem 0.5rem;
+  margin: 0.5rem;
   gap: 0.25rem;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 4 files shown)


stickz (changes requested):
Increase dir browser height to 375.
`plugin.btn = new theWebUI.rDirBrowser( 'tcreate', 'path_edit', 'browse_path', null, true, 375 );`

stickz:
> Taking another step closer to making ruTorrent mobile friendly, by applying the bootstrap grid layout to the create torrent dialog. I think we still need to re-work the "URL replacement in RSS" dialog and the "Ratio Rules" dialog before the next minor update release.
> 

ruTorrent road map adjusted. We will jump straight to version 5.0.

jevenski:
> Increase dir browser height to 375. `plugin.btn = new theWebUI.rDirBrowser( 'tcreate', 'path_edit', 'browse_path', null, true, 375 );`

A height of 375px would make the dir list exceeds the create torrent dialog a bit, so I increased the horizontal margin for the button group down below, to make the dialog window high enough to contain the entire dir list.

> ruTorrent road map adjusted. We will jump straight to version 5.0.

Awesome and exciting, and we'll have much more work to do before that happens.

PR #2684, opened 2024-06-28
Apply grid layout to Ratio Rules dialog

jevenski:
Applied Bootstrap grid layout to the Ratio Rules dialog, as well as added two quick keyboard shortcuts for moving rules up and down in the list.

1. `md` screens and larger
Screenshot 2024-06-28 223503

2. Smaller screens
Screenshot 2024-06-28 223528

### DIFF
--- a/plugins/extratio/extratio.css
+++ b/plugins/extratio/extratio.css
@@ -25,12 +25,12 @@ div#dlgEditRatioRules {max-width: 95vw;}
 /* Small devices (landscape phones, 576px and up) */
 @media (min-width: 576px) { 
   /* Custom rules for small devices */
-  div#dlgEditRatioRules {min-width: 600px;}
 }
 
 /* Medium devices (tablets, 768px and up) */
 @media (min-width: 768px) { 
   /* Custom rules for medium devices */
+  div#dlgEditRatioRules {min-width: 600px;}
 }
 
 /* Large devices (desktops, 992px and up) */
--- a/plugins/extratio/extratio.css
+++ b/plugins/extratio/extratio.css
@@ -1,30 +1,49 @@
-div#dlgEditRatioRules {width: 610px; height: 300px; font-size: 11px; font-family: Tahoma, Arial, Helvetica, sans-serif; overflow: visible}
+div#dlgEditRatioRules {font-size: 11px; font-family: Tahoma, Arial, Helvetica, sans-serif; overflow: visible}
 div#dlgEditRatioRules div.dlg-header {background-image: url(../../images/settings.gif)}
+div#dlgEditRatioRules .container {margin: 0.25rem; padding: 0; width: auto;}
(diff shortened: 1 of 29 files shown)


plugins/extratio/extratio.css
```diff
@@ -1,30 +1,49 @@
...
+/* Small devices (landscape phones, 576px and up) */
+@media (min-width: 576px) { 
+  /* Custom rules for small devices */
+}
+
+/* Medium devices (tablets, 768px and up) */
+@media (min-width: 768px) { 
+  /* Custom rules for medium devices */
+  div#dlgEditRatioRules {min-width: 600px;}
+}
```
stickz, line 34: I fixed some style issues with buttons list. Please update `extratio.css` accordingly.  See https://github.com/Novik/ruTorrent/commit/bed2f021213f14e2e8790a35ffd55b46d3fc6cbc

stickz (changes requested):

stickz:
@jevenski I have a high priority task for you - if you're interested. We need to remove the throttle dependency for `extratio` and make it optional. Over 100k instances of ruTorrent have `extratio` disabled because it requires `throttle`. Throttle limits max speed to approximately 1.6gbps in both directions due to a limitation with rTorrent. 

https://github.com/Novik/ruTorrent/blob/9bf6e15c4864a71ff9e9ca5ea689fd001931279b/plugins/extratio/plugin.info#L5

We can check if the `throttle` plugin is loaded. Then adjust our GUI and xmlrpc queries accordingly.

jevenski:
That's interesting. I didn't notice that. So from what I understand, we can do a dependency check on the `throttle` plugin as the first step. If the check returns `True`, it just flows as the old way. And if it returns `False`, we disable the `set throttle to` part of the rule setting.

As for the xmlrpc queries, we just remove the channel from the query strings. No adjustments in php scripts are needed. Is that correct?

```javascript
rTorrentStub.prototype.setratiorules = function()
{
	this.content = "mode=setrules";
	plugin.storeRuleParams();
	for(var i=0; i<plugin.rules.length; i++)
	{
		var rle = plugin.rules[i];
		var enabled = $("#_rre"+i).prop("checked") ? 1 : 0;
		var name = $("#_rrn"+i).val();
		this.content = this.content+"&name="+encodeURIComponent(name)+"&pattern="+encodeURIComponent(rle.pattern)+"&enabled="+enabled+
		        "&reason="+rle.reason+
			"&channel="+rle.channel+"&ratio="+rle.ratio+
			"&no="+rle.no;
	}
	this.contentType = "application/x-www-form-urlencoded";
	this.mountPoint = "plugins/extratio/action.php";
	this.dataType = "json";
}
```

stickz:
Great job @jevenski. I'll merge this now. And yes it looks like you're right about the `xmlrpc` queries!

PR #2685, opened 2024-06-29
Make throttle an optional dependency for extratio

jevenski:
Added a conditional statement to check if the throttle plugin is installed, and to initialize the WebUI and adjust the http request accordingly. 

There is one thing that's worth our notice, though. If we set up a rule that contains a throttle option, and disable the throttle plugin, and modify the rule setting again, and then enable the throttle plugin again, the original throttle option will be lost. It behaves this way because the PHP scripts take every submission of rule sets as creating new rules, instead of updating the existing ones.

1. With the throttle plugin installed
Screenshot 2024-06-29 154820

2. With the throttle plugin disabled
Screenshot 2024-06-29 160054

### DIFF
--- a/plugins/extratio/init.js
+++ b/plugins/extratio/init.js
@@ -50,19 +50,27 @@ plugin.loadRules = function( rle )
 	if(fltThrottle.length)
 	{
 		$('#dst_throttle option').remove();
-		fltThrottle.append("<option value=''>"+theUILang.dontSet+"</option>");
+		fltThrottle.append(
+			$("<option>").val("").text(theUILang.dontSet),
+		);
 		for(var i=0; i<theWebUI.maxThrottle; i++)
 			if(theWebUI.isCorrectThrottle(i))
-				fltThrottle.append("<option value='thr_"+i+"'>"+theWebUI.throttles[i].name+"</option>");
+				fltThrottle.append(
+					$("<option>").val("thr_" + i).text(theWebUI.throttles[i].name),
+				);
 	}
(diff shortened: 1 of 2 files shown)


plugins/extratio/init.js
```diff
@@ -219,10 +231,12 @@ rTorrentStub.prototype.setratiorules = function()
...
-		this.content = this.content+"&name="+encodeURIComponent(name)+"&pattern="+encodeURIComponent(rle.pattern)+"&enabled="+enabled+
-		        "&reason="+rle.reason+
-			"&channel="+rle.channel+"&ratio="+rle.ratio+
-			"&no="+rle.no;
+		this.content += "&name="+encodeURIComponent(name) + "&pattern=" + encodeURIComponent(rle.pattern) + 
+			"&enabled=" + enabled + "&reason=" + rle.reason + 
+			"&ratio=" + rle.ratio + "&no="+rle.no;
+		if (thePlugins.isInstalled("throttle")) {
+			this.content += "&channel=" + rle.channel;
+		}
```
stickz, line 239: Please create a variable outside the for-loop to check if the throttle plugin is enabled. If there are 10 rules, we can avoid up to 40 decision structures, by caching the results outside the for-loop.

https://github.com/Novik/ruTorrent/blob/9bf6e15c4864a71ff9e9ca5ea689fd001931279b/js/plugins.js#L75-L79

stickz (changes requested):
Overall looks good. Just one thing to improve, before I test this change.

stickz:
Looks good, thanks!

PR #2691, opened 2024-07-13
Apply flex layout to RSS Rule Manager dialog

jevenski:
This is basically a copy of the previous refined Ratio Rules dialog. Besides, I set some universal rules for general HTML elements like `div`, `input`, and `select`, instead of defining the display styles in each and every plugin separately. 

1. Desktop screens (equal to or wider than 768px):

2. Mobile screens (smaller than 768px):

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,7 +1,27 @@
 * { -webkit-touch-callout: none; -webkit-tap-highlight-color: rgba(0,0,0,1.0); }
 html, body {border: 0; margin: 0; padding: 0; cursor: default; overflow: hidden; background-color: #F5F5F5; height: 100%;}
 html, body, input,select,button,textarea{font-family: Tahoma, Arial, Helvetica, sans-serif; font-size: 11px;}
-input[type=file] {border: solid 1px #676774; border-radius: 3px; width: 100%;}
+div {
+  box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  -webkit-box-sizing: border-box;
+}
+fieldset {
+  border: 1px solid #D0D0D0;
+  border-radius:6px;
+  margin: 0.25rem 0;
+  padding: 0.25rem 0.5rem;
+}
+select, input[type=file], input[type=text] {
+  border: 1px solid #B0B0B0;
+  border-radius: 3px;
+  margin: 0.25rem 0;
+  padding: 0.25rem;
+  width: 100%;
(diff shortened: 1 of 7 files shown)


stickz:
Can we put the "Ok" and "Cancel" buttons beside each other on mobile to save space?

css/style.css
```diff
@@ -350,11 +350,13 @@ div.space {flex-grow: 1;}
 /* No media query for `xs` since this is the default in Bootstrap */
 /* Custom rules for medium devices */
 div#tadd {max-width: 95vw;}
-div#tadd .buttons-list {
+div.dlg-window .buttons-list {
```
jevenski, line 353: Using this rule to set the buttons lists in all dialog windows.

css/style.css
```diff
@@ -350,11 +350,13 @@ div.space {flex-grow: 1;}
...
   margin: 0.5rem;
   gap: 0.25rem;
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
```
jevenski, line 357: Change buttons direction from `column` to `row` on small screens.

stickz:
The global style changes in this pull request break a lot of things on the status bar and in the settings menu. We can't unify these things in `style.css`. They need to be applied individually to each of the plugins in their respective `plugin.css` files.

jevenski:
These changes do cause come display issues on the settings dialog window and some other dialog windows, such as footer buttons going out of boundary, and text boxes becoming excessively large. However, the status bar should be fine. I didn't see any mess-ups from my test.

This was before the rest of the dialog windows are rewritten for flex layout. They are all going to be rewritten anyway, and eventually, the CSS styles should be simplified and unified, in my opinion. I made this change upfront, and will adapt other components to this change in the next steps. However, the status bar should be fine. I didn't see any mess-ups from my test.

stickz:
Okay, could you rebase [jevenski:rssurlrewrite](https://github.com/jevenski/ruTorrent/tree/rssurlrewrite) with [Novik:develop](https://github.com/Novik/ruTorrent/tree/develop) ? We need to cleanup these files changed for future reference.

jevenski:
The rebase process now seems to be done. 9 files are changed in total. Really had some troubles during the rebase process. I thought I was going to give this pull request up and make a new one.

stickz:
Okay, looks good thanks.

PR #2692, opened 2024-07-19
Apply flex layout to save-to dialog window

jevenski:
This is a relatively simple dialog window, so the layout is quite straightforward. Below are the screenshots. Besides the layout, I also set some general rules for `button`s in `css/style.css`.

1. Desktop screens (min 768px wide):

2. Smaller screens (less than 768px wide):

### DIFF
--- a/plugins/_getdir/init.js
+++ b/plugins/_getdir/init.js
@@ -8,12 +8,22 @@ theWebUI.rDirBrowser = function( dlg_id, edit_id, btn_id, frame_id, withFiles, h
 	if(!frame_id)
 		frame_id = edit_id+"_frame";
 	var self = this;
-	this.btn.val("...").on( "click", function() { return(self.toggle()); } ).addClass("browseButton");
+	this.setButtonText("...")
+	this.btn.on( "click", function() { return(self.toggle()); } ).addClass("browseButton");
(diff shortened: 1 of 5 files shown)


plugins/_getdir/init.js
```diff
@@ -8,14 +8,24 @@ theWebUI.rDirBrowser = function( dlg_id, edit_id, btn_id, frame_id, withFiles, h
...
+	this.setButtonText("...")
+	this.btn.on( "click", function() { return(self.toggle()); } ).addClass("browseButton");
 	this.edit.prop("autocomplete", "off").on( browser.isIE ? "focusin" : "focus", function() { return(self.hide()); } ).addClass("browseEdit");
 	this.frame = $("<iframe>").attr( {id: frame_id, src: ""} ).css({position: "absolute", width: 0, visibility: "hidden"}).addClass("browseFrame");
 	this.dlg_id = dlg_id;
 	$('#'+dlg_id).append( this.frame );
 	this.height = !height ? 150 : height
 }
 
+theWebUI.rDirBrowser.prototype.setButtonText = function(buttonText)
```
jevenski, line 20: This function is for setting texts of different button elements.

PR #2693, opened 2024-07-19
Minor PT-pt translation corrections

Frias:


stickz:
Thanks @Frias. There's also a bunch of new phrases to be translated in the develop branch, if you want to get a head start before the big v5.0 release. https://github.com/Novik/ruTorrent/tree/develop

PR #2694, opened 2024-07-20
Apply flex layout to edit dialog window

jevenski:
Beside the rewriting of the dialog window, I also set some general CSS rules for `textarea` element in the `css/styles.css` file. Below are the screenshots:

1. Desktop screens:

2. Mobile screens:

### DIFF
--- a/plugins/edit/edit.css
+++ b/plugins/edit/edit.css
@@ -6,15 +6,15 @@ div#tedit div.dlg-header {
   background-image: url(../../images/settings.gif);
 }
 
-textarea#etrackers {
-  height: 150px;
-  word-break: break-all;
-}
-
 /* X-Small devices (portrait phones, less than 576px) */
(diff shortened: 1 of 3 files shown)


plugins/edit/edit.css
```diff
@@ -1,58 +1,43 @@
...
-{
-	border: 1px solid #D0D0D0; 
-	padding: 5px;
-	box-sizing: border-box;
-	-moz-box-sizing: border-box;
-	-webkit-box-sizing: border-box;		
+textarea#etrackers {
+  height: 150px;
+  word-break: break-all;
 }
```
stickz, line 13: Could we increase the height of the trackers textarea on desktops, while we're improving the edit dialog? I would like to be able to fit at least 8 trackers (with one line in between) without scrolling.

The line in between is a feature for tracker announce groups. It signals a new announce group for each tracker.

PR #2695, opened 2024-07-22
Migrate from grid-only to full Bootstrap

jevenski:
* Substitute `bootstrap-grid.min.css` with `bootstrap.min.css`.
* Add a couple of CSS rules to preserve styles, which would have been overridden by Bootstrap's Reboot components.
* Fix S-Table column widths issues.


jevenski:
Added margins for `checkbox` and `radio`. Their styles should be consistent with `input` of other types now.

stickz:
Looks good, thanks!

PR #2696, opened 2024-07-23
Add a responsive top menu

jevenski:
The previous step of adopting full bootstrap was a huge leap forward, reducing a huge amount of work adding responsiveness by using bootstrap's built-in components and functionality. As for the looks, there are 2 major changes:
- Borders were added to top menu buttons, which made button boundaries easier to spot on mobile screens.
- Buttons' hover effect was set to `brightness` filter. The previous implementation was using a rectangle tile as the background image, which would become odd on mobile screens, because the buttons are going to grow, the background tiles, however, remains the same size.

Some current TODO's:
- I really want to replace the image icons with bootstrap's monocolor svg icons. This should be approved beforehand, though.
- The big gap on mobile screens between the left side panel and the main torrent list is also on the agenda for rewriting. We can use a bootstrap's offcanvas on the side panel.

1. Desktop screens:

2. Mobile screens:

3. Mobile screens with top menu open:


stickz:
Could we remove the borders on the top menu icons on desktop displays?

Also, can we fix the fix the theme breakage with the top menu? I believe it's caused from changing from `div#t` to `nav#t`.

css/style.css
```diff
@@ -138,35 +138,73 @@ ul.CMenu li:hover ul.left {left: -150px}
...
+}
+div.navbar-nav div.nav-icon {
+  width: 24px;
+  height: 24px;
+  cursor: pointer;
+  background-color: transparent;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+div#add, div#remove, div#start, div#pause, div#stop, div#search, div#rss, div#setting {background-image: url(../images/toolbar.png);}
```
jevenski, line 186: Reduced the specificities on these selectors, in order that the plugin- and theme-specific rules can override them.

jevenski:
The color schemes of the collapse menu for every theme still need to be adjusted. Please allow a few more moments.

jevenski:
All the themes are now modified and tested. Everything should fit in with the overall display styles of each theme now.

PR #2697, opened 2024-07-26
Update/create PT-pt translations

Frias:
One file was created for the plugin 'Dump', if this should not be done this way please advise and I'll do it the correct way.


stickz:
Looks good, thanks!

PR #2700, opened 2024-07-29
s-table flex layout & various bug fixes

jevenski:
These small tweaks are meant to simplify the dimension control mechanism of `S-Table`s, as a preparation before rewriting the main-content part of the WebUI.


css/stable.css
```diff
@@ -1,23 +1,50 @@
-.stable {border: 1px solid; border-color: #D0D0D0 #FFFFFF #FFFFFF #D0D0D0; overflow: hidden; display: block; -moz-user-select: none; -moz-user-focus: normal; -moz-user-input: enabled; }
+.stable {
+  background-color: #FFFFFF;
+  border: 1px solid;
+  border-color: #D0D0D0 #FFFFFF #FFFFFF #D0D0D0;
+  overflow: hidden;
+  display: flex;
```
jevenski, line 6: Set to `flex` display for a smoother layout.

stickz:
I found two disk space icons on the Material Design and Oblivion themes on the status bar. I'm not sure if it was caused by this pull request. Ether way, could we address the regression here?

jevenski:
The issues were introduced when changing the status bar for responsiveness. The fixes are committed in this PR now.

stickz:
> The issues were introduced when changing the status bar for responsiveness. The fixes are committed in this PR now.

The check boxes are overlapping in the settings menu on the `Blue` theme. Could this PR resolve that problem as well?

jevenski:
> The check boxes are overlapping in the settings menu on the `Blue` theme. Could this PR resolve that problem as well?

That's one of the reasons why I wanted to remove some of the themes. The Blue theme has this `jquery.checkbox.js` plugin that alters the styles of checkboxes, by replacing the original checkboxes with images of checkboxes. The comments in the beginning of the file says the plugin was copyrighted back in 2008, which was 16 years ago. The link provided by the author cannot be reached now, and there's not much documentation in the files themselves, so I don't have a clear clue what this plugin was intended for exactly.

Apart from that, the theme's `init.js` alters the DOM of dialog windows a lot. These changes make this plugin hard to maintain, costing too much time for just one single plugin.

stickz:
> > The check boxes are overlapping in the settings menu on the `Blue` theme. Could this PR resolve that problem as well?
> 
> That's one of the reasons why I wanted to remove some of the themes. The Blue theme has this `jquery.checkbox.js` plugin that alters the styles of checkboxes, by replacing the original checkboxes with images of checkboxes. 

I don't like this theme feature. Can we remove `jquery.checkbox.js` and use restyled regular checkboxes to make the theme easier to maintain? Do you think this is a better option than removing the theme entirely?

jevenski:
> I don't like this theme feature. Can we remove `jquery.checkbox.js` and use restyled regular checkboxes to make the theme easier to maintain? Do you think this is a better option than removing the theme entirely?

Yeah, personally I think themes like Blue and Acid look old-fashioned. We can remove these additional features like `jquery.checkbox.js` and make it simpler. In the long run, maybe we can open a poll for our users to vote?

stickz:
I don't miss the query checkboxes at all for the Blue theme. I think the default ones look better. That problem is resolved.

**I found one more issue in Settings->Download->Check hash after download.**

check_hash_issue

plugins/theme/themes/Blue/init.js
```diff
@@ -20,59 +10,46 @@ plugin.showShadow = function(id,callback)
...
-			theWebUI.loadFilters = function( flt, additions )
-			{
-				plugin.loadFilters.call(theWebUI,flt, additions);
-				plugin.correctCheckboxes();
-			}
+	if (thePlugins.isInstalled('rss')) {
+		plugin.loadFilters = theWebUI.loadFilters;
+		theWebUI.loadFilters = function(flt, additions) {
+			plugin.loadFilters.call(theWebUI,flt, additions);
+			plugin.correctCheckboxes();
```
stickz, line 20: Not a function

plugins/theme/themes/Blue/init.js
```diff
@@ -20,59 +10,46 @@ plugin.showShadow = function(id,callback)
...
-			plugin.addNewFilter = theWebUI.addNewFilter;
-			theWebUI.addNewFilter = function()
-			{
-				plugin.addNewFilter.call(theWebUI);
-				plugin.correctCheckboxes();
-			}
+		plugin.addNewFilter = theWebUI.addNewFilter;
+		theWebUI.addNewFilter = function() {
+			plugin.addNewFilter.call(theWebUI);
+			plugin.correctCheckboxes();
```
stickz, line 26: Not a function.

plugins/theme/themes/Blue/init.js
```diff
@@ -20,59 +10,46 @@ plugin.showShadow = function(id,callback)
...
-			plugin.fillRSSGroups = theWebUI.fillRSSGroups;
-			theWebUI.fillRSSGroups = function()
-			{
-				plugin.fillRSSGroups.call(theWebUI);
-				plugin.correctCheckboxes();
-			}
+		plugin.fillRSSGroups = theWebUI.fillRSSGroups;
+		theWebUI.fillRSSGroups = function() {
+			plugin.fillRSSGroups.call(theWebUI);
+			plugin.correctCheckboxes();
```
stickz, line 32: Not a function.

plugins/theme/themes/Blue/init.js
```diff
@@ -20,59 +10,46 @@ plugin.showShadow = function(id,callback)
...
-			plugin.trafDraw = theWebUI.trafGraph.draw;
-			theWebUI.trafGraph.draw = function()
-			{
-				plugin.trafDraw.call(this);
-				plugin.correctCheckboxes();
-			}
+		plugin.trafDraw = theWebUI.trafGraph.draw;
+		theWebUI.trafGraph.draw = function() {
+			plugin.trafDraw.call(this);
+			plugin.correctCheckboxes();
```
stickz, line 38: Not a function.

plugins/theme/themes/Blue/init.js
```diff
@@ -20,59 +10,46 @@ plugin.showShadow = function(id,callback)
...
-			{
-				plugin.loadRules.call(theWebUI,rle);
-				plugin.correctCheckboxes();
-			}
+	}
+	if (thePlugins.isInstalled('rssurlrewrite')) {
+		plugin.loadRules = theWebUI.loadRules;
+		theWebUI.loadRules = function(rle) {
+			plugin.loadRules.call(theWebUI, rle);
+			plugin.correctCheckboxes();
```
stickz, line 45: Not a function.

stickz (changes requested):
Please remove all instances of `plugin.correctCheckboxes` since the function was removed. It is outputting JavaScript errors.

`Uncaught TypeError: plugin.correctCheckboxes is not a function from TypeError: plugin.correctCheckboxes is not a function`

stickz:
Thanks for the all the bug fixes!

PR #2703, opened 2024-08-03
Apply flex layout to several dialog windows

jevenski:
- Changes include the Edit Label dialog, the About dialog, the Help dialog, and the Torrent Properties dialog.
- Remove some dialog window id-based rules and use class-based rules instead. This should simplify the css rules and bring us easier maintenance.


koblack:
Typo s/µTorrent/uTorrent/ ?

jevenski:
Thanks for the heads up! Typo fixed.

jevenski:
Most of the windows look the same as before, except for the torrent property window, where responsiveness was added. Adding some screenshots:

stickz:
This pull request is going to be merged next, if you'd like to rebase with develop for testing purposes.

jevenski:
I merged the up-to-date develop branch into this branch. Things work out.

stickz:
I noticed the keyboard shortcuts in the help dialog are harder to see with the black background. This includes on both desktop and mobile displays. What do you think about reverting this change? Everything else is otherwise good.

jevenski:
I'm OK with reverting that. I was trying to be semantic in this part by using the `kbd` HTML element to represent key names, and then they are automatically styled by Bootstrap. It's fine to have them just stay in a `label` wrapper.

stickz:
Thanks. I found a couple bugs. I'm just wondering if these should be fixed now or later.

1) Clicking on `Plugins...` on mobile view, the options list will not show long enough to click it.
2) The buttons on the add RSS feed dialog are off the screen.

jevenski:
> 1) Clicking on `Plugins...` on mobile view, the options list will not show long enough to click it.

This one is related to the context menu click mechanism. I plan to rewrite this part in the future too, but we can use a temporary fix for now.

> 2) The buttons on the add RSS feed dialog are off the screen.

I think this one can wait till we rewrite the rss plugin, which I have been already doing in one of my local feature branches. The rss plugin includes most kinds of components to be written for reaponsiveness, such as dialog windows, option contents, s-table lists, etc.

stickz:
Thanks!

PR #2704, opened 2024-08-08
Rewrite _getdir plugin

jevenski:
I did some further enhancements on the _getdir plugin, including:

- combine `getfiles.php` and `getdirs.php` into one script `listdir.php`;
- return the directory list in `JSON` format and render it in an HTML `<div>` instead of returning server-rendered HTML and display it in an `<iframe>`;
- attach the directory list frame to the root node (actually a wrapper `#dir-container`), instead of the containing dialog window, so that the frame can grow outside the boundary of the dialog window and still show up properly. Otherwise, the directory list would be "clipped" when the height exceeds that of its containing window. The original `Save To` window used an invisible pad attached to the bottom to contain the extra height of the frame. In my previous flex layout patch, I increased the gap between the options so that the window would be high enough to contain the frame. Now the problem is solved.
- add a linkage to the directory list frame and its containing window for drap-n-drop movement, since the frame now don't belong to the containing window, it would not have moved automatically with the window.
- allow navigating through the directories using arrow keys (`ArrowUP` and `ArrowDown` for selecting, `ArrowRight` for entering the selected directory, `ArrowLeft` for going back to the parent directory);
- adapt the directory list frame to different themes.


jevenski:
Yeah, sure. This is a change that I just did while working towards a mobile friendly interface. The current _getdir plugin works fine, so there's no rush in pulling this one. I'll send pull requests related to a responsive interface first. The next one will be applying flex layout to the main content, and then another one to add an off-canvas side panel. Then I think the interface should be qualified as "responsive". The rest is just to take some time to follow the same pattern and rewrite dialog windows and tab contents.

jevenski:
This PR now conflicts with upstream changes. I'll cherrypick and test it later after the responsive sidepanel is merged.

stickz:
> This PR now conflicts with upstream changes. I'll cherrypick and test it later after the responsive sidepanel is merged.

I would like to delay this until `v5.1`, if that is okay with you. This will accelerate the time frame by 2 weeks and reduce regressions for `v5.0`.

jevenski:
> I would like to delay this until `v5.1`, if that is okay with you. This will accelerate the time frame by 2 weeks and reduce regressions for `v5.0`.

I'm totally good. There's just one thing that might require a fix in `v5.0`: the directory list frame doesn't align with the path input box in the settings dialog, e.g. in the "unpack" section and the "autotools" section.

stickz:
> > I would like to delay this until `v5.1`, if that is okay with you. This will accelerate the time frame by 2 weeks and reduce regressions for `v5.0`.
> 
> I'm totally good. There's just one thing that might require a fix in `v5.0`: the directory list frame doesn't align with the path input box in the settings dialog, e.g. in the "unpack" section and the "autotools" section.

We can let that go. The time and success impact it is higher the impact of the bug for the `v5.0` scope. I agree with this full solution for `v5.1`. We just need to delay it until this next minor release.

There's a more critical regression with v5.0 to resolve. I believe it was introduced in #2705. There's an element overlap here.
rutorrent_tabs_overlap

jevenski:
> There's a more critical regression with v5.0 to resolve. I believe it was introduced in #2705. There's an element overlap here.
> 
> rutorrent_tabs_overlap

Yes, I've seen that too and have already working on that.

PR #2709, opened 2024-08-24
Add responsive side panel

jevenski:
- Remove some CSS rules and use the equivelant Bootstrap classes for easier maitenance, since CSS rules are harder to maintenant and tend to grow larger and messier overtime.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -245,10 +245,12 @@ Input.TextboxLarge {border: 1px solid #B0B0B0; font-size: 11px;  padding: 2px; w
 
 a {color: #686868;}
 #maincont {height: calc(100vh - 36px - 5px - 25px - 5px);}
-#offcanvas-sidepanel {min-width: 200px;}
+#offcanvas-sidepanel {
+  min-width: 200px;
+  background-color: #FFFFFF;
+}
 #side-panel {
   border: 1px solid #A0A0A0;
-  background-color: #FFFFFF;
 }
 #list-table {
   border: 1px solid #A0A0A0;
--- a/css/style.css
+++ b/css/style.css
@@ -102,19 +102,18 @@ dialog {
   border-color: var(--menu-border-color);
 }
 form{margin:0;}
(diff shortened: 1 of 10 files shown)


stickz:
Could you review the white background on the responsive side panel for different themes and make adjustments where appropriate?

TrimmingFool (commented):
@jevenski Hi,
I like the idea of a responsive side panel.
I have not yet tried out these changes but was wondering where `tailwind.min.css` is. Has it been added in a previous PR on develop?

Nevertheless, I think, adding tailwind to rutorrent was/is not the best idea, (except for maybe the new web-components panel-label, category-panel).
I feel, that tailwind is better suited for styling individual components, where styles rarely need to be reused. I imagine, that changing the jquery code to dynamically add and remove tailwind classes will not improve the readability but rather make it worse.

chatgpt:
> HTML Clutter: Utility-first CSS can lead to verbose HTML with many classes applied to elements. This can make your HTML harder to read and maintain, especially in a jQuery-heavy codebase where you might be dynamically adding or removing classes.

stickz:
@TrimmingFool We are using Bootstrap not Tailwind. We need this wrapper for mobile purposes.

TrimmingFool:
@stickz Ah, thanks! That explains it :smile: Apparently, I am out pretty of the loop.

I fear that, besides the clutter, dynamically toggling utility css-classes with jquery means that model, view and control can/will not really be separate.

I just wanted to raise this concern, as [turning the category list into a web-component](https://github.com/Novik/ruTorrent/pull/2588), had the intention of improving this separation.
It seems a little late to pivot now but maybe this could be considered at another time.

@jevenski @stickz Anyway, thank you, for taking this on!

jevenski:
> I feel, that tailwind is better suited for styling individual components, where styles rarely need to be reused. I imagine, that changing the jquery code to dynamically add and remove tailwind classes will not improve the readability but rather make it worse.

HTML clutter might be the only downside of adding utility classes to web components. We're mainly manipulating the DOM by jQuery, and will mainly be reading the JS code, so I think adding these classes can make things easier, especially for different theme styling, which is at the moment really a pain in the back. In the future I am thinking about defining a set of custom pallettes for each of the themes, so that the work of adjusting themes can be reduced.

stickz:
> It seems a little late to pivot now but maybe this could be considered at another time.
> 

Yes, we can address separation in `v5.1`. @jevenski Please feel to do whatever you need (within reason) to get `v5.0` released with mobile functionality. 

I'm reviewing for functionality and polish not code quality. We need this narrow scope for `v5.0`.

stickz:
@jevenski Any ETA on addressing the white background of the responsive side panel for different themes? I have time to review again this weekend.

jevenski:
Okay, I'll try to get it done before this weekend begins.

stickz:
Merged, thanks!

PR #2710, opened 2024-09-02
Apply flex layout to option-related plugins

jevenski:
Apply flex layout to 4 plugins that add a page to the options window: 1) Accounts (loginmgr plugin); 2) Autotools; 3) XMPP; 4) Unpack. These changes are quite routine and just follow the previous patterns. Additionally, the unpack plugin also has a simple dialog window, and it's rewritten as well.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -338,6 +338,10 @@ div.op100l {width: 400px; float: left}
   background-color: var(--settings-background-color);
   overflow: hidden auto;
 }
+.stg_con legend {
+  position: sticky;
+  top: 0;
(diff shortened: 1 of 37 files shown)


stickz:
For the autotools setting page, I like the mobile layout better. I think we should use it for Desktop as well. It gives more room to type values into the text boxes. The label display is also much cleaner. 

The only exceptions being the highlighted elements. These should be beside each other to save space on the Desktop layout.

autotools_layout

stickz:
Thanks!

PR #2711, opened 2024-09-04
Apply flex layout to RSS plugin

jevenski:
- Adjust the layout of the RSS plugin related components, including:
  1. an option page;
  2. a dialog to add rss feed and another dialog to edit it;
  3. a dialog to add rss group;
  4. a dialog to load torrents from rss;
  5. a dialog to edit rss filters;
- Modify the structure of the header of dialog windows, using flex as well.
- Restrict the max height of dialog windows and set the overflow property of the content to `auto` for very long dialog windows, such as the RSS manager dialog on mobile display.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -123,6 +123,8 @@ div.dlg-window {
   border-radius: 8px;
   display: none;
   max-width: 95vw;
+  max-height: 95vh;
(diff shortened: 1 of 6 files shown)


stickz:
Thanks!

PR #2715, opened 2024-09-23
Apply flex layout to various plugins

jevenski:
- Flex layout was applied to the following plugins: cookies, lookat, retrackers, uploadeta, extsearch, history. Most of the changes were done in the sense of option pages, except for the extsearch plugin that has an add torrent from search result dialog window as well.
- Implement the connecting lines in the left-side navigation of the option window with CSS borders, and remove the previous background images. This could save a few bytes of bandwidth.


jevenski:
At the moment, there are only four plugins left to be rewritten for responsive UI: scheduler, channels, ragio groups, and screenshots.

stickz:
Looks good. I would like to request 2 minor changes before merge.

**1) Checkbox and textbox on separate lines for desktop on unpack plugin.**
ruTorrent_unpack_sep_lines

**2) Retrackers to fit in the settings menu without scrolling.**
ruTorrent_retrackers_no_scroll

koblack:
Maybe we can do something like this?

unpack-after

stickz:
> Maybe we can do something like this?
> 

Yes agreed.

stickz:
Thanks @jevenski. Any chance in working on the remaining 4 plugins next?

I would like to tag ruTorrent v5.0 Beta 1 afterwards. Then get the changes distributed, so we can get development feedback coming in. We have enough stability and everything is better than it was on mobile. Tagging a release shortly, will allow us to more easily resolve any potential regressions. We can continue mobile development. This is just a milestone for our progress.

PR #2716, opened 2024-09-25
Apply flex layout to three of the remaining plugins

jevenski:
In this pull request, three of the four remaining plugins were rewritten for the responsive UI. I modified the setting dialog layout a bit, so that the tables on the setting pages of these three plugins can overflow and scroll within the fixed parent box. These tables remained as HTML `<table>`, although previously most HTML `<table>`s that served as layout boxes were rewritten to `<div>`s to provide a smoother layout; these ones are real tables. As I consider again about the `<table>`s, I think the table that sets decimal sizes on the "Format" page, which was changed into a flex `<div>`, should also be reverted to an HTML `<table>`.

Changes on the last plugin screenshots are on the way, too.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -137,7 +137,6 @@ div.dlg-window div.row > div {
   align-items: center;
   padding: 0 0.25rem;
   margin-bottom: 0.1rem;
-  overflow: auto;
 }
 div.fxcaret {overflow: auto}
 div.dlg-header {
--- a/css/style.css
+++ b/css/style.css
@@ -46,7 +46,8 @@ textarea {
 }
 label {
   cursor: pointer;
-  margin: 0.15rem 0.25rem;
+  margin: 0.15rem;
+  overflow: hidden;
 }
 input[type=checkbox], input[type=radio] {margin: 0.25rem;}
 select, input[type=file], input[type=text], input[type=number], input[type=password] {
@@ -123,7 +124,7 @@ div.dlg-window {
   border: 2px solid #909090;
   border-radius: 8px;
   display: none;
-  max-width: 95vw !important;
+  max-width: 95vw;
   max-height: 95vh;
 }
 div.dlg-window div.row {
--- a/css/style.css
+++ b/css/style.css
@@ -10,12 +10,9 @@ div, td, label, fieldset, textarea, select, input, button, .Button {
   -moz-box-sizing: border-box;
   -webkit-box-sizing: border-box;
 }
-table {
+table, tbody, td, tfoot, th, thead, tr {
   /* prevent collision with bootstrap */
-  border-collapse: initial;
-}
-tbody, td, tfoot, th, thead, tr {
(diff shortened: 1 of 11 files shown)


koblack:
1. Unnecessary scrollbar
image

2. Maybe the Name field can be shorter to avoid scrollbar
image

3. The Help and Properties window is huge
image
image

koblack:
Thanks a lot for everything! A few minor things.

1. Open getdir from the Unpack or Autotools tab. Change tab and the directory selection box remains visible.
image

2. Autotools tab unnecessary scrollbar on mobile.
Screenshot_20240926_062937_Chrome2

3. Format tab unnecessary scrollbar on mobile.
Screenshot_20240926_062948_Chrome2

jevenski:
Thanks for testing and the bug catches, @koblack .I've been doing testing as well on as many devices as I can possibly reach lately, and been trying to catch any bugs. The second and the third scrollbar issues will be fixed.

As for the first one, I have done a written version of the _getdir plugin in another [pull request](https://github.com/Novik/ruTorrent/pull/2704), which has a lower priority. Some improvements and bug fixes were implemented in that one, so I think if we are to fix this bug, it's better that we merge the entire patch to solve a dozen problems all at once.

stickz:
Please fix alignment of scheduler days text on desktop.
ruTorrent_schedule_days

On mobile, I would like to request 2 rows with 12 bigger squares each. The reason is to avoid scrolling and make them selectable.
ruTorrent_schedule_days_mobile

This is optional. I will merge the pull request without it. The request is to reduce the limits from 3 to 2 for ease of usage.
ruTorrent_scheduler_limits

stickz:
> 
> As for the first one, I have done a written version of the _getdir plugin in another [pull request](https://github.com/Novik/ruTorrent/pull/2704), which has a lower priority. Some improvements and bug fixes were implemented in that one, so I think if we are to fix this bug, it's better that we merge the entire patch to solve a dozen problems all at once.

The plan is to finish this pull request, finish the screenshots plugin pull request, backport stability changes from master then tag ruTorrent v5.0 Beta 1. Afterwards, I'll move up #2704 into the ruTorrent v5.0. We'll address user feedback from Beta 1, merge the `_getdir plugin rewrite` then release Beta 2. Finally, we'll address user feedback from Beta 2 then tag a stable release.

All new features are now frozen except for the required `_getdir plugin rewrite` and screenshots plugin. I will release the freeze after Beta 2 is tagged. Beta 1 will be done on the develop branch. Beta 2 will be done on the master branch.

jevenski:
The scheduler graph was rewritten using bootstrap grid for a responsive UI.

The third limit was kept for now. Maybe we should mark it as deprecated in hope of users stop using it, and then remove it in the next major version release?

stickz:
Looks good thanks @jevenski. Let's keep the third limit for now. I don't want to delay the first beta timeline.

PR #2666, opened 2024-05-29
added a copy button to console log window

jevenski:
Added a copy button to console window, making copying console log like mediainfo result a little bit more confortable, without having to select the entire text and hit Ctrl+C. Log a message and nothing else when the console log is empty, e.g. when screenshots are being generated.

console_log

PS: Apologies again for all the mess by creating pull requests again and again... I made some adjustments with my fork and closed the old request accidentally. Now I'm getting a hang of the git philosophy.

PR #2717, opened 2024-09-29
Apply flex layout to screenshots plugin

jevenski:
Apply flex layout to other dialog windows
- Apply flex layout to screenshots plugin.
- Apply flex layout to screenshots plugin.
- Apply flex layout to other dialog windows including: Add Peer Window, Yes-or-No Window, Magnet Link Bulk Add Window, Peer Comment Window.
- Prevent button textes from wrapping, and set enable button width to auto on the history plugin option page.
- Switch to rutorrent's copyToClipboard() function for browser compatibility. Hide the copy button on task complete if the log output is empty. Hope it could fix this error: https://github.com/Novik/ruTorrent/pull/2666#issuecomment-2284822387

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -61,7 +61,7 @@ button, input[type=button], input[type=submit], .Button {
   padding: 1px 10px;
   background: #F0F0F0 url(../images/h.gif) repeat-x center bottom;
   width: 80px;
-  cursor: pointer;
+  text-wrap: nowrap;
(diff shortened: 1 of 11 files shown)


stickz:
Thanks @jevenski. We can fit #2704 into `ruTorrent v5.B.1` if you want. Everything else is completed and ready to go.

PR #2718, opened 2024-09-29
Rewrite _getdir plugin

jevenski:
Due to the heavy modifications made to upstream code, I made this new pull request based on the latest develop branch. Most of the features are the same as the original pull request, plus a few fixes for bugs that got their way in during the last few commits.

Plugin Features:
- Combine getfiles.php and getdirs.php into one script: `listdir.php`;
- Return the directory list in JSON format and render it in an HTML `<dialog>` instead of returning server-rendered HTML and display it in an `<iframe>``;
- Attach the directory list frame to the root node (actually a wrapper `#dir-container`), instead of the containing dialog window, so that the frame can grow outside the boundary of the dialog window and still show up properly. Otherwise, the directory list would be "clipped" when the height exceeds that of its containing window. The original Save To window used an invisible pad attached to the bottom to contain the extra height of the frame. In my previous flex layout patch, I increased the gap between the options so that the window would be high enough to contain the frame. Now the problem is solved.
- Add a linkage to the directory list frame and its containing window for drap-n-drop movement, since the frame now don't belong to the containing window, it would not have moved automatically with the window.
- Allow navigating through the directories using arrow keys: `ArrowUP` and `ArrowDown` for selecting, `ArrowRight` for entering the selected directory, `ArrowLeft` for going back to the parent directory, `Home` and `End` for going to the top and the end respectively;
- Adapt the directory list frame to different themes.

Original Pull Request: https://github.com/Novik/ruTorrent/pull/2704


plugins/_getdir/listdir.php
```diff
@@ -0,0 +1,43 @@
...
+	global $dir;
+	return is_dir($dir.$item);
+});
+$files = array_filter($items, function ($item) {
+	global $dir;
+	return is_file($dir.$item);
+});
+
+header("Content-Type: application/json");
+echo json_encode(array(
```
stickz, line 38: Please use the ruTorrent wrapper: `JSON::safeEncode(array(`. It's autoloaded, no further imports required.

plugins/_getdir/listdir.php
```diff
@@ -0,0 +1,43 @@
+<?php
+require_once( '../../php/util.php' );
+require_once( '../../php/settings.php' );
+eval(FileUtil::getPluginConf("_getdir"));
+
+$theSettings = rTorrentSettings::get();
+
+if(isset($_REQUEST['dir']) && strlen($_REQUEST['dir']))
+{
+	$dir = rawurldecode($_REQUEST['dir']);
```
stickz, line 10: Please declare a variable instead of repeatedly referencing `$_REQUEST['dir']`.

stickz (changes requested):
Please investigate why the directory browser does not default to the torrent files directory for all use cases. This is a regression with this pull request. Once these three issues are resolved, I'll merge and tag `ruTorrent v5.B.1`.

plugins/_getdir/listdir.php
```diff
@@ -0,0 +1,46 @@
...
+	{
+		$dir = $topDirectory;
+	}
+}
+else
+{
+	$dir = User::isLocalMode() ? $theSettings->directory : $topDirectory;
+	if(strpos(FileUtil::addslash($dir),$topDirectory)!==0)
+		$dir = $topDirectory;
+}
```
stickz, line 28: Here is my revision to this code block to display the directory list properly by default. Please test and make any necessary changes. The file list was not displaying on my web client, due to missing a slash at the end.
```
else
{
	$dir = User::isLocalMode() ? $theSettings->directory : $topDirectory;
	$dir = FileUtil::addslash($dir);
	if(strpos($dir,$topDirectory)!==0)
		$dir = FileUtil::addslash($topDirectory);
}
```

stickz (changes requested):
Please address the requested revision to make the file list display properly by default. Afterwards, revise the pull request. If it passes on my web client, I'll merge and tag the beta release!

plugins/_getdir/listdir.php
jevenski: Is it because the `$topDirectory` config value lacking a trailing slash? The conf file says it explicitly that it needs a trailing slash:
```php
$topDirectory = '/';			// Upper available directory. Absolute path with trail slash.
```

I added an `addslash()` function to this variable, hope it can solve the problem.

stickz: This did not solve my problem. I opted for a different fix instead.

PR #2722, opened 2024-10-02
Various bug fixes for v5-beta1

jevenski:
Fix legend text overlapping by adding a non-transparent background to it. Apply some minor display adjustment for MaterialDesign theme as well.

Screenshot 2024-10-02 222117


jevenski:
I mixed a few more bug fixes into this pull request, so we can speed up the code review process. Correct me if it's better practice to aim one target per pull request.
These new commits fixes the "Save" button on the spectrogram dialog and the create-torrent dialog. Old spectrogram save fix didn't get it right yet. The "Save" button was added every time we create a spectrogram, and it's still there after we close the dialog. So if we create 100 spectrograms, there will be 100 "Save" buttons. As for the create-torrent dialog, the "Save" button was added as visible, while it should be added as hidden.

stickz:
Thanks!

PR #2723, opened 2024-10-05
Various bug fixes for Beta 2

jevenski:
There are several issues there to be fixed. Here are some of those that are more familiar to me (introduced by myself, oops). Changes include:

* Optimize creating dialog window
Previously, the dialog windows were created using HTML strings. The strings were generated from jQuery objects, and would be converted to jQuery objects again when making dialog windows. This process back and forth dropped the event handlers which were bound to elements using JavaScript's `addEventListener()` method or jQuery's `on()` method, and only preserved those using inline HTML attibutes which was not the recommended way of binding event handlers to web elements. The optimized way of creating dialog windows using jQuery objects directly and thus can preserve both kinds of event handlers. And it can fallback to create jQuery objects if the input dialog content was in string format.
* Fix screenshots playing button symbol
* Add options for WebUI resize
Add options to restrict the min height of list table section, and the min width of the side panel.
* Fix _getdir dependency issue
Remove _getdir plugin installed check in WebUI initialization, which could cause errors when dealing with directory lists. To fully separate the _getdir plugin behavior from the main app initialization, use dialog handlers to close directory lists along with dialog windows instead.
* Adapt side panel close button themes
Use a variable to define button background colors, and define the color variable in different themes. This way, the code duplication can be reduced, and maintenance can be increased.
* Fix side panel triangle size shift in DarkBetter
* Add dumb lang files for dump plugin
* Hide views panel on mobile
This is a temporary workaround for the rename views dialog not editable on mobile, by hiding the views panel on mobile.
* Fix edit label icon error


stickz:
Please disable the views panel by default and hide the icon edit feature on mobile to mitigate unresolved bugs.

jevenski:
> Please disable the views panel by default and hide the icon edit feature on mobile to mitigate unresolved bugs.

More fixes have been added. Rename views panel has been hidden on mobile, as I haven't located the cause of bugs here. Edit icon error has been fixed.

Issue #2721, opened 2024-10-01
ruTorrent v5.1 - Beta (Please reply here for feedback)


stickz:
### What would you like changed about the web client?

I would like the legend not to overlap with text when scrolling on mobile.

### Describe the solution you'd like and propose possible alternatives.

Implementing appropriate CSS style changes to avoid the undesired behavior.

### Provide information and resources about the environment hosting the web client.

Server: ruTorrent v5.0-Beta 1 on Ubuntu 24.04 LTS 
Web Client: Google Chrome 129.0.6668.59
Installation Method: zip source from version release

### Additional context

I believe @jevenski is able to resolve this problem. Here is a screenshot to provide additional context.

![ruTorrent_v5_legend_issue](https://github.com/user-attachments/assets/fa4190f7-83dd-4a47-8b7d-0a40344ff531)

jevenski:
Can we preserve the sticky feature of the legend text, that sticks to the viewport until the cobtaining box is scrolled completely outside the viewport? I like this one, as it's cool and helpful for long fieldsets blocks. We can define the background color of the legend text to prevent the overlapping.

jevenski:
Yes, we can have the elements inside the group. We just add a non-transparent background to the legend text to cover the elements underneath to prevent overlapping. I've created a new pull request here (with a screenshot inside):

https://github.com/Novik/ruTorrent/pull/2722

Airman8:
### What would you like changed about the web client?

1: On mobile, the side menu accessed by the "..." button takes up the entire width of the viewport, resulting in there being no room beside the menu to touch and close out of it. If I switch my phone to landscape mode and then back to portrait mode again, the menu width is correct (the same as on my PC). Refreshing the page after switching orientation results in the menu again expanding to the full viewport width.

2: Another minor thing is the tabs in the lower torrent info section (General, Files, Trackers, etc.). When I press a tab, it slowly expands the tab bar vertically and pushes down the content of the tab below. It also results in the tab quickly snapping up and then sagging down again every time I press a different tab.

### Describe the solution you'd like and propose possible alternatives.

1: Probably needs a CSS fix to force the menu not to expand all the way, or perhaps a way to properly detect the media width on first load? Otherwise, a button could be added to the left of the "ruTorrent v5.B.1" label to close the side menu. Another option would be to shift the vertical position of the side menu down so as not to cover the button on the main page.

[...]

Issue #2596, opened 2023-11-14
Delete Key does not prompt a "Remove Torrent" confirmation, even with the option checked in.


Pabsilon:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Link provided to install script if applicable
- [X] Not using broken rtinst install script
- [X] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Web Broser: Brave 1.60.114
ruTorrent v4.2.6
OS: Debian Bulseye | 5.10.0-26

### Tell us how you installed ruTorrent

I'm running ruTorrent next to rTorrent in the crazymax docker container

### Describe the bug

When pressing delete with a torrent selected, the .torrent file and the downloaded files are deleted without a confirmation box. This happens even with the "Confirm when deleting torrents" option checked.

This does not happen when using the top menu bar, and it doesn't happen when using the right click menu provided with the Deletedata plugin.

### Steps to reproduce

1. Select a torrent from the list
2. Press the 'Del' key on your keyboard
3. Your torrent has been deleted without confirmation

### Expected behavior

The confirmation box with "Remove Torrent(s) : Do you really want to remove the selected torrent(s)?" should be shown before deleting.

stickz:
I labeled this as an enhancement. We can add an option to display a confirmation box, when using the delete key on `v4.2`. It needs to be a separate option though and I think it should be disabled by default.

FunThomas76:
i have the same problem, with very old version 4.0.2 it works fine and the setting for confirmation dialog for delete works as expected and in 4.2.8 it does not matter what the setting is it deletes without confirmation.
For me it is more secure to confirm -> since if you press delete you could maybe delete all torrents if you are in the wrong folder.

is it known which is the last version where this deletion confirmation dialog works?

stickz:
@FunThomas76 Could you go to `Settings -> General` and see if `confirm when deleting torrents` is checked off?

FunThomas76:
it saves the state under settings -> general -> confirm when deleting but checked or not checked makes here no difference
my first thought was that some config file had not the right file permissions but the state of the setting get saved but it does not change anything here

stickz:
Resolved in v5.0 releasing shortly.

FunThomas76:
The bug is back in my 5.1.5 :-(
switched the settings for confirmation dialog but nothing changed -> when mark a torrent and press DEL it does not ask for confirmation

PR #2724, opened 2024-10-09
Fix torrent deleted without confirmation

jevenski:
- This commit fixes an error that torrent(s) get deleted without confirmation, even if the option is checked in the options. Previously, the delete action worked as expected using the delete button in the top menu, asking for confirmation before taking actual action. However, using the "remove" option in the context menu or the `DEL` key on the keyboard would got the torrent(s) removed immediately. They were just not calling the correct remove function.
- Besides the correction of delete events handler, two other things were carried out by the way, although not on the scope of the entire app but just around the functions that were adjusted in this fix.
  1. Some ident related issues were picked out by the way, like mixing spaces with tabs on the same line.
  2. Documentations were added to some of the JS functions, according to my own understanding of the function. They might be incorrect, but can be polished while the project grows.

Related:
https://github.com/Novik/ruTorrent/issues/2596
https://github.com/Novik/ruTorrent/issues/2721#issuecomment-2394279265 (the part of accidentally tapping "remove" in the last paragraph)

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -235,7 +235,7 @@ function escapeHTML(str)
 /**
  * Confirm again before taking action.
  * @param {string} title Title of the dialog.
- * @param {string} content Hinit of the action to be confirmed.
+ * @param {string} content Hint of the action to be confirmed.
  * @param {Function | string} funcYesName Function to be executed if confirmed.
  */
 function askYesNo(title, content, funcYesName) {
--- a/js/common.js
+++ b/js/common.js
@@ -160,10 +160,10 @@ $.fn.extend(
 		}));
 	},
 
-	enableSysMenu: function()
-	{
-		return(this.on("contextmenu",function(e) { e.stopImmediatePropagation(); }).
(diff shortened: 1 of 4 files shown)


stickz:
Thanks!

stickz:
@jevenski Could I ask a small favour? Could you backport just the delete key fix to the `v4.3` branch? I want to push out a version that Gentoo Linux can mark as stable. It will be a major milestone for the ruTorrent project to get a stability mark on Linux!

jevenski:
> @jevenski Could I ask a small favour? Could you backport just the delete key fix to the `v4.3` branch? I want to push out a version that Gentoo Linux can mark as stable. It will be a major milestone for the ruTorrent project to get a stability mark on Linux!

Yes, sure thing. Just wanna confirm that we just need the correction on bringing up the confirmation dialog, *without* the indent and the documentation part?

geopsllc:
On that topic, is anyone else seeing an issue where delete with data is doing the data removal but the torrent stays in the webpage list until a full page reload?

Issue #2726, opened 2024-10-11
Minor API break in plugin.addPaneToStatusbar()


anthonyryan1:
### What would you like changed about the web client?

Just wanted to make note of a couple minor changes to the `plugin.addPaneToStatusbar()` that are causing minor bugs for 3rd party plugins.

Previously, a 3rd party plugin would make a call like the following to add something to the status bar: `plugins/test/init.js`:
```js
plugin.init = function() {
        plugin.addPaneToStatusbar('test',  $('<div>').text('Testing'));
};

```

After 5.0, this call creates duplicates in the status bar, with every second element being a duplicate.

![Screenshot](https://github.com/user-attachments/assets/b4e8b074-7d67-4932-a8e3-0ce4adc570c3)

-----

As a second API note. The second argument statusCell now requires a jQuery wrapped DOM object, when previously it accepted both a jQuery wrapped DOM object and a native DOM object.

This was never documented in any plugin I'm aware of, the API has internally always used jQuery wrapped DOM objects,  and it was mostly just luck it worked if it was ever used in the past. I would consider the passing of native DOM objects to the second argument undefined behavior and don't think that's worth fixing or noting (beyond this paragraph).

### Describe the solution you'd like and propose possible alternatives.

a. We add default arguments for addPaneToStatusbar()
------------

By having default arguments `no = 0` & `mobileVisible = true`  all plugins calling the old way will get matching behavior.

b. Document the API break and encourage plugins to pass the additional arguments.
-------

Because this break happened on 5.0, there's no violation of semantic versioning to leave it as is and require plugins to change their behavior. Some plugins may not be maintained though and will exhibit the buggy behavior until the maintainer returns or they're forked off.

### Provide information and resources about the environment hosting the web client.

Affecting version 5.0.0.

Bug verified in Chrome 129 & Firefox 131.

Installed on Gentoo Linux (rolling release, so no applicable version). Using an ebuild that is functionally identical to https://gitweb.gentoo.org/repo/gentoo.git/tree/www-apps/rutorrent/rutorrent-4.3.7.ebuild

stickz:
@jevenski What do you think about adding default arguments to `addPaneToStatusbar`, so we don't break third party plugins?

jevenski:
Yeah, I saw that. I'm working on a patch to fix the backward compatibility issue. Furthermore, this function can be refactored a little bit to work with both jQuery objects and HTML strings, just like `addPageToOptions()`.

PR #2732, opened 2024-10-13
Fix addPaneToStatusBar backward compatibility

jevenski:
Refactored the `addPaneToStatusbar()` function a little bit for better backward compatibility. The second parameter now accepts a jQuery object, an HTML element, or an HTML string. The last 2 parameters are now optional, and will be assigned a default value during running. Added documentation for this function as well by the way.

Related: https://github.com/Novik/ruTorrent/issues/2726


stickz:
Merged, going to do a release control (beta) version of v5.1. We will confirm everything works properly there.

Issue #2729, opened 2024-10-12
Mediainfo plug-in text selection and Copy formatting


chancemon:
### What would you like changed about the web client?

Prior to version 5.0 of the mediainfo plug-in, generated text from the plug-in could be manually selected using the cursor.  
Additionally the pasted output of text copied using the Copy button is missing needed formatting.

### Describe the solution you'd like and propose possible alternatives.

Restore cursor text selection and fix formatting of text copied via the Copy button.

### Provide information and resources about the environment hosting the web client.

Affecting version 5.0.0.

Bug verified in Chrome 129 & Firefox 131.

Installed on Gentoo Linux (rolling release, so no applicable version). Using an ebuild that is functionally identical to https://gitweb.gentoo.org/repo/gentoo.git/tree/www-apps/rutorrent/rutorrent-4.3.7.ebuild

mjakopa:
I concur that this is indeed an issue, since it's now impossible to get mediainfo information in a meaningful way using ruTorrent. The copy button is a welcome addition, but retaining the formatting is a must.

stickz:
> I concur that this is indeed an issue, since it's now impossible to get mediainfo information in a meaningful way using ruTorrent. The copy button is a welcome addition, but retaining the formatting is a must.

We're fixing this shortly and releasing a hotfix.

Issue #2708, opened 2024-08-14
MediaInfo Download button.


hon4:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

In the MediaInfo it would be very helpful to add a Download button downloading the MediaInfo as a .txt file.

stickz:
I agree. If anyone wants to submit a pull request for the `develop` branch, I will review and incorporate the feature into ruTorrent.

PR #2733, opened 2024-10-13
Fix task log copy formatting

jevenski:
- Preserve the break lines to maintain correct formatting while copying console log using copy button.
- Enable user select on the console log output.
- Add a button to save console log as a text file. This was planned for v5.1, but I thought since the two topics were so related, I would like to address them in one patch.

Related: 
https://github.com/Novik/ruTorrent/issues/2729
https://github.com/Novik/ruTorrent/issues/2708

### DIFF
--- a/plugins/_task/_task.css
+++ b/plugins/_task/_task.css
@@ -6,6 +6,7 @@
   overflow: auto;
   width: 580px;
   cursor: text;
+  user-select: text;
 }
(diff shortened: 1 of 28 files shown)


stickz:
Merged, going to do a release control (beta) version of v5.1. We will confirm everything works properly there.

PR #2736, opened 2024-10-13
MaterialDesign: update fonts

koblack:
The local fonts used in MaterialDesign contain only the basic character set. This causes problems with displaying characters correctly in some languages e.g. Polish. ruTorrent has been translated into 27 languages. To display dialectal characters correctly in many different languages, we need fonts with the full character set.
Importing fonts allows us to dynamically load the diacritics when needed. Not everyone needs to download all of them. Also, instead of TTF the WOFF2 format with compression makes the fonts smaller.


stickz:
I'm okay with this change to support different languages, but can we add all the resources to the ruTorrent project? There are multiple reasons why we load everything locally instead of from a third party source.

koblack:
I modified the pull request and added just more local fonts. In the case of English, the fonts had a total size of 186,523 bytes before the change and 125,328 bytes after the change, so they are now ~33% smaller.

stickz:
Thanks looks good. I'll verify it works tomorrow and merge if there are no problems. 

125KB is great considering they don't load unless the material design theme is used.

stickz:
Thanks!

Issue #2737, opened 2024-10-13
Disabling "Show categories on startup" presents user with empty sidebar


anthonyryan1:
### What would you like changed about the web client?

If a user toggles the "Show categories on startup" to an off state, and reloads the interface the categories panel on the left side of the interface will show an empty panel.

Behavior persists through F7 toggling.

When this setting is off, `#side-panel` hidden as seen here:
https://github.com/Novik/ruTorrent/blob/v5.1-beta1/js/webui.js#L441

Meanwhile toggleCategories() is acting on `#offcanvas-sidepanel` as seen here:
https://github.com/Novik/ruTorrent/blob/v5.1-beta1/js/webui.js#L2334

### Describe the solution you'd like and propose possible alternatives.

I believe the correct fix is to set

```js
                if (!this.settings["webui.show_cats"]) {
                        $("#offcanvas-sidepanel").addClass("d-md-none");
                        $("#HDivider").addClass("d-md-none");
                }
```

Instead, so both are operating on the same object.

### Provide information and resources about the environment hosting the web client.

Affecting version 5.0.0 and 5.1 beta 1.

Bug verified in Chrome 129 & Firefox 131.

Installed on Gentoo Linux (rolling release, so no applicable version). Using an ebuild that is functionally identical to https://gitweb.gentoo.org/repo/gentoo.git/tree/www-apps/rutorrent/rutorrent-4.3.7.ebuild

anthonyryan1:
`#Hdivider` should also likely be `d-md-none` but the fact that we're toggling both `.d-md-none` and `.d-md-block` causes it to do nothing. It looks like it's winding up set with both `.d-none` and `.d-md-block` and using the second.

Updating proposed fix above.

PR #2739, opened 2024-10-14
Fix side panel toggling

jevenski:
This PR addresses the issue regarding the side-panel display issue, as mentioned in https://github.com/Novik/ruTorrent/issues/2737. Apart from the bug fix, some other clean-ups were also made, such as removing verbose Bootstrap utility classes and adding documentations to some util functions.


stickz:
Thanks!

PR #2745, opened 2024-10-18
fire input change event for _getdir plugin

nelu:
Right now, the input value is being updated but no input change event is being triggered, like the normal behaviour of a input


stickz:
Merged!

PR #2746, opened 2024-10-19
screenshots plugin: take screenshot with correct aspect ratio from anamorphic video

koblack:
* take screenshot with correct aspect ratio from anamorphic video
* add m2ts extension to allow taking screenshot from bluray discs
* sort extensions alphabetically


stickz:
Thanks!

Issue #2734, opened 2024-10-13
Limited context menu improvements on mobile


stickz:
### What would you like changed about the web client?

I would like the expandable context menu for mobile to be improved for the next release.

### Describe the solution you'd like and propose possible alternatives.

1. I would recommend placing `Start`, `Pause` and `Stop` on the same line on mobile (if feasible) to create extra room.
2. I would recommend dropping the `Plugins...` main context on mobile. Then bring the sub contexts to main context on mobile using the Plugins icon. (It should also support custom plugins as required)
    - `URL Replacement in RSS`
    - `Ratio Rules`
    - `Bulk Loading`

### Provide information and resources about the environment hosting the web client.

**Server:** ruTorrent v5.0-Stable on Ubuntu 24.04 LTS
**Web Client:** Google Chrome 129.0.6668.59
**Installation Method:** zip source from version release

### Additional context

![ruTorrent_context_menu_improvements](https://github.com/user-attachments/assets/420a1f66-48d1-4034-b9bc-0792df4b35cf)

stickz:
@jevenski What are your thoughts on this suggested improvement for the next release?

jevenski:
Yes, we can definitely bring the three brothers and sisters (Start, Pause, Stop) onto one line. There is plenty of room horizontally for them. 

As for the Plugins submenu, I was thinking about adopting Bootstrap's [dropdowns](https://getbootstrap.com/docs/5.3/components/dropdowns/) to fit the navbar. The old dropdown menu on the navbar shares the same mechanism with the context menu in the torrent list, the file list, etc. I think this is inefficient because the submenu items are static on the navbar, while the menu items on the context menus are dynamic, depending on what users are right-clicking on. Look at the init check for the `theContextMenu` object at Line 302 - Line 307: that looks like an if-else nightmare. 
https://github.com/Novik/ruTorrent/blob/b125b6901d23085c784c7fb9d4a3c1e8b55e62c7/js/objects.js#L302-L307
So for the navbar sub-menus, we can add the sub-menu items directly to the DOM, and make the containing icon a dropdown item. The same goes for the search button on the navbar.

stickz:
Okay sounds good @jevenski! Let's going with the plan of Bootstraps dropdowns!

PR #2749, opened 2024-10-22
Limit context menu on mobile

jevenski:
I made the following changes to the top menu for now, including changing the Plugins button to a Bootstrap dropdown menu, and the grouping of some buttons. These changes could be breaking as well, if some third-party plugins call the `plugin.addButtonToToolbar()` or the `rPlugin.registerTopmenu()` functions.
The search engine button is more complex, so it stays unchanged for now.

- Replace plugins button with dropdown menu.
- Group buttons on top menu, sorting add torrent and remove torrent buttons into one group, and start, pause, and stop buttons into another group. Since the add torrent and remove buttons are grouped, the create torrent button will also go into the group because it will be inserted before the remove torrent button on plugin load.

Related: https://github.com/Novik/ruTorrent/issues/2734

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -212,15 +212,22 @@ div#t {
   border-bottom: 1px solid #D0D0D0;
   z-index: 30;
 }
+#top-menu .btn-group {
+  flex-direction: column;
+}
+.dropdown-toggle span {
+  /* Push dropdown triangle to the right end */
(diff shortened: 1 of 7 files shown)


stickz:
I tested this and it looks good. Thanks so much @jevenski ! Now we just have the persistent add and create torrent options.

Issue #2753, opened 2024-10-23
JS error: [ : 1153] Uncaught TypeError: Cannot read properties of undefined (reading 'length')


koblack:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Web browser and ruTorrent version provided

### Tell us about your environment

Web Browser: Google Chrome 130.0.6723.70
ruTorrent: master: https://github.com/Novik/ruTorrent/commit/e15310fc07b3f2e1443d611204985a97caf155c5

### What were you trying to do?

Try adding a new rule using the rssurlrewrite manager "URL Replacement in RSS"

### What us the error message?

JS error: [ : 1153] Uncaught TypeError: Cannot read properties of undefined (reading 'length')

### Give us the stack trace.

  Uncaught TypeError: Cannot read properties of undefined (reading 'length') from TypeError: Cannot read properties of undefined (reading 'length')
    at theWebUI.addNewRule (<anonymous>:1153:290)
    at HTMLButtonElement.onclick (http://.../:1:10)

PR #2760, opened 2024-10-24
Fix RSS URL rule error

jevenski:
- Fix error when trying to add new RSS rules using the add button. This error probably was probably introduced when the dialog was rewritten with jQuery.
- The scope of functions defined in this plugin (rssurlrewrite) seemed a little messy. Some functions need to go into the `plugin` object, some need to go into the `theWebUI` object, and others need to stay local. I moved some of these functions a bit, but that's still not perfect yet. Some further fine tunings are still required.

Related: https://github.com/Novik/ruTorrent/issues/2753

### DIFF
--- a/plugins/rssurlrewrite/init.js
+++ b/plugins/rssurlrewrite/init.js
@@ -75,19 +75,19 @@ theWebUI.loadRules = function( rle )
 	for(var lbl in this.rssLabels)
 		$('#RLS_rss').append("<option value='"+lbl+"'>"+this.rssLabels[lbl].name+"</option>");
 	plugin.rules = rle;
-	theWebUI.maxRuleNo = 0;
+	plugin.maxRuleNo = 0;
 	if (plugin.rules) {
 		plugin.rules.forEach((rule, index) => {
-			if (theWebUI.maxRuleNo < rule.no)
-				theWebUI.maxRuleNo = rule.no;
+			if (plugin.maxRuleNo < rule.no)
+				plugin.maxRuleNo = rule.no;
 			list.append(
 				makeRuleListItem(rule, index),
 			);
 		});
 		plugin.rules.forEach(rule => {
 			if (rule.no < 0) {
-				theWebUI.maxRuleNo++;
-				rule.no = theWebUI.maxRuleNo;
+				plugin.maxRuleNo++;
+				rule.no = plugin.maxRuleNo;
 			}
 		});
 	}
@@ -98,7 +98,7 @@ theWebUI.loadRules = function( rle )
 
 theWebUI.addNewRule = function() {
 	const list = $("#rlslist");
-	theWebUI.maxRuleNo++;
+	plugin.maxRuleNo++;
 	const f = {
 		name: theUILang.rssNewRule,
 		enabled: 1,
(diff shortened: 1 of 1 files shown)


plugins/rssurlrewrite/init.js
```diff
@@ -1,42 +1,40 @@
...
-if(browser.isIE && browser.versionMajor < 8)
+if (browser.isIE && browser.versionMajor < 8)
 	plugin.loadCSS("ie");
 plugin.loadLang();
 
-theWebUI.curRule = null;
-theWebUI.maxRuleNo = 0;
+plugin.curRule = null;
+plugin.maxRuleNo = 0;
+plugin.rules = [];
```
jevenski, line 8: These attributes should be local to the plugin itself, instead of the global `theWebUI` object.

plugins/rssurlrewrite/init.js
```diff
@@ -1,42 +1,40 @@
...
 
-theWebUI.showRules = function()
-{
-	theWebUI.request("?action=getrules",[theWebUI.loadRules, this]);
+function makeRuleListItem(rule, index) {
```
jevenski, line 10: Extract this action for it to be reused in `addNewRule()` and `loadRules()`.

plugins/rssurlrewrite/init.js
```diff
@@ -1,42 +1,40 @@
...
-		this.curRule.className = 'TextboxFocus';
-		var no = parseInt(this.curRule.id.substr(3));
-		var rle = this.rules[no];
+/**
+ * Actions to be performed when a RSS rule is selected.
+ * This function is going to be used as an event handler
+ * on rule's textbox focus events, so the `this` keyword
+ * refers to the event target, i.e. the focused rule item.
+ */
+function selectRule() {
```
jevenski, line 29: This function is only going to be used as an event handler for focusing on a rule list item, so a local function will work fine.

stickz:
@koblack Could you test this before I merge? I'm not an expert on the RSS elements of ruTorrent.

koblack:
Tested on 279f45b69604961a0cc58506a9af753d4e8e41f4

```
Uncaught TypeError: this.request is not a function from TypeError: this.request is not a function
    at theWebUI.checkCurrentRule (<anonymous>:1151:87)
    at HTMLButtonElement.dispatch (http://xxx/ruTorrent-jevenski/js/jquery.js?v=51B2:2:40035)
    at v.handle (http://xxx/ruTorrent-jevenski/js/jquery.js?v=51B2:2:38006)
```

jevenski:
I did some further adjustments towards the scope of plugin functions. The culprit was still the `this` keyword in JavsScript. 

As a sidenote, it seems to me that the `checkRule()` function sends the Regexp tests to the server and retrieve the test result from it. The PHP code block does nothing server-specific but running the Regexp tests and spitting the results. It looks like a waste of resources because this can be done perfectly on the client side. Correct me if I missed anything.

stickz:
@koblack Could you confirm this PR is ready to merge?

@jevenski We can adjust `checkCurrentRule()` in `v5.2`. I would also like to drop support for Internet Explorer in `v5.2` as well. Everything is perfectly functional right now. Let's keep it "as is" so we can release this version in the near future!

koblack:
Everything seems to be working fine now! Thanks a lot!

PR #2764, opened 2024-10-27
Allow wrapping in button groups on top toolbar

jevenski:
This patch address an issue that grouped buttons overflow on mobile if there are too many of them, for example the logoff plugin would add a log-off button before the add torrent button, so there will be four buttons in total in this group if the create plugin is also in use.

Besides, a duplicate event listener on the quick search text bar, and some duplicate CSS rules on the browse directory button were removed.


stickz:
Thanks!

PR #2765, opened 2024-10-27
lookat plugin: update urls

koblack:


stickz:
Backported to `v4.3` and releasing `v4.3.9` shortly thanks so much @koblack!

PR #2766, opened 2024-10-27
Fix s-table sorting error

jevenski:
- When an s-table was set to sort by a column which was created by a plugin, and then the plugin was disabled, the error would occur after a page reload.
- Remove some of the verbose Bootstrap utility class names.


stickz:
Backported to `v4.3` and releasing `v4.3.9` shortly thanks so much @jevenski!

PR #2748, opened 2024-10-20
Increase font size for mobile screens

jevenski:
- Set font size to 16px for mobile, and 11px for desktop.
- Adjust s-table's `createRow()` method to use jQuery to create table rows.
- Adjust CPU load and disk space plugins styling for larger font size on mobile.
- Adjust top menu and main content height for different screens.
- Adjust some of the setting pages to better suit for the enlarged font size.

PR #2773, opened 2024-11-01
Adjust task console log layout

jevenski:
### Before

Screenshots on mobile has overflow flaws.

> Screenshot_20241021_091054_Chrome

_Originally posted by @koblack in https://github.com/Novik/ruTorrent/issues/2748#issuecomment-2425825286_

### After

- Cleaned up the CSS rules on the task console window and its image frames to make it display within screen limitation on mobile. 
- Merge some of the rules for the screenshots plugin and the spectrogram plugin, since they have something in common.
- Using CSS to position images frames in the center, instead of the old method that was using JavaScript to dynamically position them centrally.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -636,11 +636,7 @@ div.dlg-window .buttons-list {
   div#tadd {min-width: 600px;}
   div.dlg-window .buttons-list {
     margin: 1rem 0.5rem;
-    gap: 0.25rem;
-    display: flex;
-    flex-direction: row;
     justify-content: end;
-    flex-wrap: nowrap;
   }
   div#stg {width: 660px;}
 }
--- a/plugins/_task/_task.css
+++ b/plugins/_task/_task.css
@@ -1,15 +1,29 @@
+#tskConsole {
+  width: 95vw;
+}
 .tskconsole {
   font-family: monospace;
-  line-height: 16px;
+  line-height: 1.5rem;
   background: #F5F5F5;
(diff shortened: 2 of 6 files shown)


css/style.css
```diff
@@ -636,11 +636,7 @@ div.dlg-window .buttons-list {
   div#tadd {min-width: 600px;}
   div.dlg-window .buttons-list {
     margin: 1rem 0.5rem;
-    gap: 0.25rem;
-    display: flex;
-    flex-direction: row;
     justify-content: end;
-    flex-wrap: nowrap;
```
jevenski, line 643: These media query rules are the same as the default ones for small screens, so we can remove them here to avoid duplication.

koblack:
Thank you so much for fixing this!

1. The progress bar still hides behind the buttons.
Screenshot_20241101_115906_Chrome

2. The desktop version keeps changing size
ezgif-3-fec371fad1

jevenski:
Thanks for testing and the feedback, @koblack . The previous solution was still flawed. I added 2 new commits to fix the issues. Would be appreciated if you could test it again. (I never got a chance to see screenshots showing up instantly one by one. Usually I just waited and waited and then they all showed up tegother at once.)

koblack:
Checked and everything is fine :-)

Screenshot_20241102_114055_Chrome

Nagrywanieekranu2024-11-02113310-ezgif com-video-to-gif-converter

Issue #2757, opened 2024-10-24
_getdir plugin issues displaying proper file names


nelu:
### What would you like changed about the web client?

Seems the _getdir plugins has a problem displaying file names containing multiple spaces " "
The issue is similiar with https://github.com/nelu/rutorrent-filemanager/issues/54

### Describe the solution you'd like and propose possible alternatives.

A similar fix with https://github.com/nelu/rutorrent-filemanager/issues/54 if possible: ``` textVar.replace(/ /g, '\u00a0') ```

### Provide information and resources about the environment hosting the web client.

Docker: https://hub.docker.com/r/unzel/rutorrent-filemanager/

### Additional context

Shows without whitespaces for a file named `"                     Poma"`: 
![image](https://github.com/user-attachments/assets/16a1db84-8b71-49cb-8622-c595a41f163d)
![image](https://github.com/user-attachments/assets/9a3a0af7-be52-47a9-82e2-fe2a12c92700)

While in the reality the file is named:
![image](https://github.com/user-attachments/assets/48e37614-6d49-4449-82c5-f3c8e3517d99)

PR #2780, opened 2024-11-06
Use CSS to indicate the importance of whitespace rather than replacing with nbsp's

### LINKED ISSUE #2779: _getdir regression with whitespace and nbsp
### What would you like changed about the web client?

Commit https://github.com/Novik/ruTorrent/commit/38c6489e60729fe841a5e6904ea288fd74e1f8fe / https://github.com/Novik/ruTorrent/pull/2768 appears to have introduced a regression with whitespace handling in 5.1 beta 4.

Since this commit,  `/plugins/create/action.php`  `cmd: "create"` is getting unicode non-breaking spaces `xC2A0`, instead ASCII spaces and throwing errors during torrent creation, when the file name creates a torrent. This is causing non-existing path errors.

Reproduction:
1. Create a file with spaces in the name `dd if=/dev/zero bs=1024 count=1 of="file with spaces.bin"`
2. Open Dev tools network tab, and filter by "plugins/create/action.php"
3. Select "file with spaces.bin" as your source in the "Create New Torrent" dialog.
4. Create...
5. Copy `path_edit` from the request arguments and convert to hex to observe that the spaces are not ASCII.

### Describe the solution you'd like and propose possible alternatives.

It looks like part of that change is juggling between \u00a0 (equal to hex `xC2A0`) and ascii whitespace. I expect the intention there was to tackle line wraps in the directory selector menu?

When the create plugin is getting the value out, it's using jQuery's `.val()` without any fancy substitution. as seen here: https://github.com/Novik/ruTorrent/blob/v5.1-beta4/plugins/create/init.js#L19  

I don't fully understand the initial motivation behind the non breaking spaces, but my instinct is that if the nbsp replacements were used for layout purposes, there's probably a CSS trait we could use instead and keep the proper ASCII spaces in the code and avoid juggling between valid strings, and invalid-but-displayed strings.

### Provide information and resources about the environment hosting the web client.

ruTorrent: 5.1_beta4
Browser: Firefox 132.0.1
PHP: 8.3.13
OS: Gentoo Linux (rolling release)
Install script:  Roughly https://gitweb.gentoo.org/repo/gentoo.git/tree/www-apps/rutorrent/rutorrent-4.3.9.ebuild

Create torrent backend is mktorrent.sh / mktorrent 2021-01-30

### Additional context

_No response_

anthonyryan1:
For your consideration, a CSS-only fix for the whitespace disappearing problem.

Closes #2779
Alternate fix for #2757
Partially reverts #2768

### DIFF
--- a/plugins/_getdir/_getdir.css
+++ b/plugins/_getdir/_getdir.css
@@ -18,6 +18,9 @@
   color: windowtext;
   padding: 0.15rem;
   text-wrap: nowrap;
+
+  /* whitespace here is deliberate, and should be shown as we set it */
+  white-space: pre;
 }
 .rmenuitem.active {
   color: highlighttext;
--- a/plugins/_getdir/init.js
+++ b/plugins/_getdir/init.js
@@ -121,7 +121,7 @@ theWebUI.rDirBrowser = class {
 
 	requestDir() {
 		$.ajax(
-			`plugins/_getdir/listdir.php?dir=${encodeURIComponent(this.edit.val().replace(/\u00a0/g, " "))}&time=${(new Date()).getTime()}${this.withFiles ? "&withfiles=1" : ""}`,
+			`plugins/_getdir/listdir.php?dir=${encodeURIComponent(this.edit.val())}&time=${(new Date()).getTime()}${this.withFiles ? "&withfiles=1" : ""}`,
 			{
 				success: (res) => {
 					this.frame.find(".filter-dir").val("").trigger("focus");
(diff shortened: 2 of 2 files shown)


anthonyryan1:
Don't merge this yet, more problems being investigated.

anthonyryan1:
Re-opening. I believe this is functionally correct. But there's also a separate, unrelated issue with creating torrents from a single file with spaces in the name that still needs more investigation.

PR #2782, opened 2024-11-08
Update style.css

Airman8:
Changing the units vh and vw to dvh and dvw in order to prevent overlap with system navigation bar on mobile devices.


Airman8:
Sorry! Didn't notice that Koblack already covered this one.

PR #2790, opened 2024-11-13
Enum key mismatch in extratio plugin

anthonyryan1:
The rule type select box on the Ratio rules page accidentally got reverted.

This pull request demonstrates the simplest approach to fixing it, fixing the order of the javascript list, so it matches the server side enum keys. As well as documenting the relationship so anyone changing it will see the comment.

If I were to argue best practices, we should probably switch to using strings instead of ints for enums. One benefit of string based enums is that they decouple storage from presentation a bit better. So that if design minded people wanted to re-order the options in the select box, order would stop mattering to the code.

It is however currently the precedent through the code that these enum constants are stored as integers (eg. `SCH_*` & `RAT_*` also following this pattern). So I'm submitting this with the minimal change and happy to revise if you guys would like to start moving towards strings for enum keys.


jevenski:
https://github.com/Novik/ruTorrent/blob/v4.3/plugins/extratio/init.js#L322-L325

The old code had a pitfall here that really got me when I was rewriting this part: the options were swapped in an order where value 3 went above value 2. Nice catch here bro!

PR #2796, opened 2024-11-15
ENV configuration in config.php

nelu:
Add ability to configure persistent data paths and network services from ENVIRONMENT configuration variables.


stickz:
We will merge this in ruTorrent `v5.2`.

davorb:
I think it would make sense to also document this somewhere

stickz:
> I think it would make sense to also document this somewhere

Agreed. @nelu Could you update the public accessible wiki? https://github.com/Novik/ruTorrent/wiki/Config

nelu:
I edit the wiki page with a new sample config, hope is sufficient.

stickz:
> I edit the wiki page with a new sample config, hope is sufficient.

Thanks this is good. Please see this change to the `develop` branch. I've updated it.

Issue #2784, opened 2024-11-09
Creating a torrent with no files


anthonyryan1:
### What would you like changed about the web client?

In the torrent creation dialog, if the user selects no files:

![Screenshot from 2024-11-08 21-18-47](https://github.com/user-attachments/assets/244f07cb-28b1-4a3f-ba4e-08572c926e45)

The creation proceeds, finds to files and returns some PHP warnings and a success message.

```
Warning: Undefined array key "name" in /var/www/localhost/htdocs/rutorrent/plugins/create/createtorrent.php on line 98
Warning: Undefined array key "piece length" in /var/www/localhost/htdocs/rutorrent/php/rtorrent.php on line 182
Done.
````

And it provides an empty .torrent file.

### Describe the solution you'd like and propose possible alternatives.

I think there's two fixes that need to be done here.

1. The "Create..." button shouldn't function if Select Source is complete empty.

Normally I'd suggest something like `<input required>` for the element, but in this case, because it's a JS form and not a HTML form, that attribute alone won't do anything to prevent submission. We'd need additional JS code to use those attributes.

2. We should likely return an error message from the `plugins/action.php` `cmd: "create"` when there's no files

We can't reasonably create a torrent file from nothing, it's better to return an error message and exit early than to let PHP try to run through the creation code with reasonable assumptions like there being files.

### Provide information and resources about the environment hosting the web client.

ruTorrent: 5.1_beta4
Browser: Firefox 132.0.1
PHP: 8.3.13
OS: Gentoo Linux (rolling release)
Install script: Roughly https://gitweb.gentoo.org/repo/gentoo.git/tree/www-apps/rutorrent/rutorrent-4.3.9.ebuild

PR #2797, opened 2024-11-17
Fix create torrent from empty path

jevenski:
- Check path length before proceeding with creating torrent files on both front-end and back-end.
- Move `checkCreate()` function to local scope to gradually reduce the size of `theWebUI`.
- Update `.gitignore` rules to include project code formatting styles.

Fixes: https://github.com/Novik/ruTorrent/issues/2784

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "editor.tabSize": 2,
+    "editor.insertSpaces": false,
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
stickz, line 1: `.gitignore` didn't exclude this file.

stickz (changes requested):

PR #2804, opened 2024-11-24
Fix torrent deleted without confirmation

koblack:
@jevenski this is your commit from pull request #2724 
I removed from original commit part renaming `ndx` -> `key` and this resolved issues with saving column sorting.
This should go into v5.2 where we will have more time for testing.

> * Fix torrent deleted without confirmation
> 
> - This commit fixes an error that torrent(s) get deleted without confirmation, even if the option is checked in the options. Previously, the delete action worked as expected using the delete button in the top menu, asking for confirmation before taking actual action. However, using the "remove" option in the context menu or the `DEL` key on the keyboard would got the torrent(s) removed immediately. They were just not calling the correct remove function.
> - Besides the correction of delete events handler, two other things were carried out by the way, although not on the scope of the entire app but just around the functions that were adjusted in this fix.
>   1. Some ident related issues were picked out by the way, like mixing spaces with tabs on the same line.
>   2. Documentations were added to some of the JS functions, according to my own understanding of the function. They might be incorrect, but can be polished while the project grows.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -160,10 +160,10 @@ $.fn.extend(
 		}));
 	},
 
-	enableSysMenu: function()
-	{
-		return(this.on("contextmenu",function(e) { e.stopImmediatePropagation(); }).
-			on("selectstart",function(e) { e.stopImmediatePropagation(); return(true); }));
+	enableSysMenu: function() {
+		return this
+			.on("contextmenu", (e) => e.stopImmediatePropagation())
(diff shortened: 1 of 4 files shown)


jevenski:
That's cool. It's time we start a new development cycle and add more features to it.

stickz:
Please submit change to `develop` branch where the new `v5.2` cycle is started.

PR #2805, opened 2024-11-25
Remove drag mask object

jevenski:
This patch removes the old drag mask object, and makes drag and drop action instant. It also separates moving directory list frames from moving the containing dialog windows for better maintenance.

Screen Recording 2024-11-24 131802

### Commit message:
- Without the drag mask object, dialog windows can be moved around instantly during dragging the mouse.
- Move dir list frame dnd movement function to the _getdir plugin for separation.


stickz:
Could we add a maximum width to the category list? So the user can't make the torrent info list-table disappear?

jevenski:
@stickz Sure thing. I think we can simply add a hard-coded max width restriction to the `resizeLeft()` function, say like 50% of the window width. This is less likely to be set as per user's preference in the WebUI options.

jevenski:
@stickz Sure thing. I think we can simply add a hard-coded max width restriction to the `resizeLeft()` function, say like 50% of the window width. We don't need to complicate the WebUI options with this one, as 50% should be more than enough for most use cases.

stickz:
Could we put a new tab in the settings menu called `developer`? Then put this restriction plus these two in there? It's helpful for us to be able to change these values, without having to find the hard-coded values.

dev_settings

jevenski:
How does this one look like? There is a difference in the units of min width (pixel) and max width (%). Setting a limitation on the max width using percentage should be more intuitive because max width is dependent on the width of the entire window, which is subject to change.

stickz:
> How does this one look like? 

Looks good.

js/webui.js
```diff
@@ -2223,9 +2224,12 @@ var theWebUI = {
 		if (!w)
 			return;
 		const offcanvas = $("#offcanvas-sidepanel");
-		if (theWebUI.settings["webui.list_table_min_height"]) {
+		if (theWebUI.settings["webui.side_panel_min_width"]) {
```
jevenski, line 2227: This is a typo fix that we can also backport to v5.1.

stickz: Yes, agreed! Can you submit a PR to master for this?

js/webui.js
```diff
@@ -123,6 +123,7 @@ var theWebUI = {
 		"webui.confirm_when_deleting":	1,
 		"webui.alternate_color":	0,
 		"webui.side_panel_min_width": 	200,
+		"webui.side_panel_max_width_percent": 50,
```
stickz, line 126: We can set this to a more sane value like `35`. It's not going to impact mobile devices. We can assume 720p or higher screen.

stickz (changes requested):

PR #2809, opened 2024-11-26
Use bootstrap dropdown on recent trackers

jevenski:
- Get recent trackers and previous create torrent settings on plugin load, and update the dropdown menu immediately. This replaces previous mechanism of showing recent trackers with a context menu.
- Move some utility functions from `theWebUI` object to the plugin itself. This reduuces the size of `theWebUI` object, and avoids potential function naming conflicts.
- Return new data immediately with the `rtdelete` request. This saves the need for another request to fetch the new data after deleting some trackers from recent ones.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -123,6 +123,9 @@ span {
   --bs-dropdown-link-hover-color: var(--menu-highlight-color);
   --bs-dropdown-link-hover-bg: var(--menu-highlight-background-color);
 }
+a.dropdown-item {
+  width: auto;
+}
 category-list.rightalign-labelsize panel-label::part(size) {
   margin-left: auto;
 }
@@ -255,7 +258,6 @@ span.htkey {text-align:right; position: absolute; right: 16px; z-index:100}
   align-items: center;
   justify-content: space-between;
   padding: 0.25rem 1rem;
-  width: auto;
 }
 .navbar-toggler {
   background-color: var(--btn-bg-color);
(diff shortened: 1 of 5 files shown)


stickz:
I found a critical bug with create torrent. After a torrent is created with a new line in-between the trackers, the trackers will save without the new line. It's imperative we have a new line in-between trackers, so the announce groups are created properly.

jevenski:
https://github.com/Novik/ruTorrent/blob/c38fc35937d19b677387615006ca0e9956b93097/plugins/create/init.js#L93

This line removes empty new lines when adding trackers from the recent list. Perhaps we should remove this filter after all?

I did two tests creating torrent files with and without empty new lines between trackers. It turned out the results were expected when submitting with explicit new lines in between.

### Test 1: with new lines
Screenshot 2024-11-27 235043

Screenshot 2024-11-27 235147

### Test 2: without new lines
Screenshot 2024-11-27 235318

Screenshot 2024-11-27 235405

stickz:
We need to trim any **trailing** new lines. But they should not be trimmed from the **middle** of the string.

jevenski:
Alright, that's a new lesson learned - I don't know new lines make a different before. Will fix this later.

jevenski:
Changes are made from both front- and back-end. This can also be backported to v5.1 if the changes are accepted.

stickz:
Works as intended thanks @jevenski! 

Can you backport just the tracker fix to master (v5.1)? A major provider is waiting for this fix to deploy.

PR #2819, opened 2024-12-07
Use bootstrap list group on options window

jevenski:
Screen Recording 2024-12-07 224040

- Use bootstrap list group component on settings navigation. As a side-effect, this introduces bootstrap's fade-in/-out effect when switching between different option pages.
- Implement the connecting lines in front of settings navigation items using pseudo-elements and reuse line styles and colors with variables. This simplifies the DOM structure and reduces code duplication.
- Extract default option pages creation and initialization from previous `makeContent()` function to `theOptionsWindow` object itself. This makes functions and methods co-located and easier to maintain, and reduces the size of the `makeContent()` function.
- Remove CSS rules for IE browser on settings window since we're dropping IE support in the next major update.


js/objects.js
```diff
@@ -448,3 +448,404 @@ var theContextMenu =
 		return(false);
 	}
 }
+
+// Options window
+const theOptionsWindow = {
```
stickz, line 453: Can we put this in a new JavaScript file?

stickz (changes requested):

stickz:
@jevenski On the "Advanced" tab of the settings menu, can we put the check boxes on the same line as the text labels? Can we also backport this improvement to the master branch to reduce scrolling on mobile?

stickz:
Thanks!

PR #2821, opened 2024-12-10
Change various s-table methods

jevenski:
- Adjust s-table's `createRow()` method to use jQuery to create table rows. This is more readable than generating table rows
  with HTML strings.
- Rewrite `toggleColumn` to use jQuery's `toggle()` method to toggle table cell display and use jQuery's selector function to replace a for loop.
- Deprecate `calcSize()` method with a deprecation warning since we're unifying behaviors on different browsers. This method will be removed in the next major version update.

PR #2825, opened 2024-12-13
Remove Internet Explorer support, and some useragent sniffing hacks

anthonyryan1:
User agent sniffing was a bad idea. I think we should eventually remove all of them so every browser runs the same code paths. The problem with looking at browsers to implement workarounds for their bugs, is that browsers are moving targets. The fix for a bug 20 years ago is more likely to cause of a separate bug today.

One fun one is https://github.com/Novik/ruTorrent/blob/v5.1.3/js/stable.js#L369 which used to try to sniff Webkit <537, but  because of other bug fixes sometime in the past 20 years, is applying to Chrome <537. I've just made it apply to all Chrome's for now (since that was the current behavior), but it may be revised in a future iteration.


anthonyryan1:
Just realized there's some overlap in #2821, I'm happy to rebase after that's done.

Issue #2817, opened 2024-12-06
Uncaught Error: Invalid dimensions for plot


koblack:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Web browser and ruTorrent version provided

### Tell us about your environment

Web Browser: Android Google Chrome 131.0.6778.104
ruTorrent: develop: https://github.com/Novik/ruTorrent/commit/036d8a90c0bd9b0b690d6bcfd87c4655b5e3173a

### What were you trying to do?

On mobile..
Error occurs when you are on the traffic tab and move your device between portrait and landscape position.

![example](https://github.com/user-attachments/assets/7ce998a0-0bb6-493e-8b1c-0dc159ccdd6e)

### What us the error message?

Uncaught Error: Invalid dimensions for plot, width = 585.786, height = 0

### Give us the stack trace.

```
webui.js?v=510:2439 Uncaught Error: Invalid dimensions for plot, width = 585.786, height = 0 from Error: Invalid dimensions for plot, width = 585.786, height = 0
    at V.resize (http://box/ruTorrent-develop/js/jquery.flot.js?v=510:32:33021)
    at W.resize (http://box/ruTorrent-develop/js/jquery.flot.js?v=510:32:26082)
    at rTraficGraph.update (http://box/ruTorrent-develop/js/graph.js?v=510:124:12)
    at http://box/ruTorrent-develop/js/graph.js?v=510:139:14
```

PR #2826, opened 2024-12-13
Distribute WebUi space using flex layout

jevenski:
- Resize category list, torrent list table and torrent detail tabs using `flex-basis` property. This reduces the amount of JS doing resizing work and lets CSS take over the space distribution.
- Add a deprecation warning for the old `theWebUI.resizeTop(w, h)` method and suggest using the new simplified `theWebUI.resizeTop(h)` method.
- Fix an error that occurs when traffic graph is shrinked down to less than 1px high.

Fixes https://github.com/Novik/ruTorrent/issues/2817

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -415,6 +415,7 @@ a {color: #686868; text-decoration: none;}
   display: flex;
   flex-direction: column;
   align-items: stretch;
+  flex: 0 0 auto;
 }
 .offcanvas-body {
   display: flex;
@@ -426,6 +427,8 @@ a {color: #686868; text-decoration: none;}
   flex: 1 1 auto;
 }
 #list-table {
+  flex: 0 0 auto;
+  overflow: hidden;
   border: 1px solid var(--container-border-color);
   background-color: var(--container-bg-color);
 }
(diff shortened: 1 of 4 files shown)


jevenski:
This patch also fixes this issue mentioned in an early feedback for v5.0.

> I also found that clicking on certain tabs in all the other themes causes the whole bottom section to grow vertically by 1 or 2 pixels. The tabs that do this the most often are the history and plugins tabs, but it happens with others depending on the theme. (Incidentally, this doesn't happen in the MaterialDesign theme.) I thought that it might be caused by the content of the active tab having x-overflow content, which could cause the scrollbar to appear and maybe bump the tab upward. But it happens even if I make the various fields in each tab narrower so that everything fits on the screen, so I'm not sure what's really making it happen. This isn't a big usability issue or anything. It's just a quirky UI thing where I can see the tab bar moving slightly up and down as I click through the tabs.
>
> _Originally posted by @Airman8 in https://github.com/Novik/ruTorrent/issues/2721#issuecomment-2392365795_

stickz:
@jevenski This will be the next pull request merged. Afterwards, we need a hotfix for all platforms on `v5.1` (without the deprecation). There's an issue when I tag my web browser out of maximize mode. The traffic graph does not update.
ruTorrent_max

ruTorrent_max2

koblack:
Toggle details [F6] does not work correctly.
After pressing [F6] only the contents of the tab are hidden but this part of the window remains visible.

Screenshot:
image

stickz:
Thanks @jevenski. I'll tag ruTorrent `v5.1.4` once this is backported for a hotfix!

PR #2830, opened 2024-12-17
Define getter functions on s-table

jevenski:
- Remove `rows` and `cols` properties from dxSTable class and define them using getter functions. Previously they need to be updated manually when adding or removing rows/columns. The new method extracts the up-to-date numbers directly from the rows/cols-data, so that the numbers don't need to be updated manually.
- Remove properties referring to various s-table parts and define getter functions to get a reference to them as well. This way the creation process of s-tables can be combined into one piece of code snippet with jQuery DOM creation functionality and thus making maintenance easier, instead of being scattered across many lines. The process can be further polished in the future, just so much for now to keep changes to a minimum.

### DIFF
--- a/js/stable.js
+++ b/js/stable.js
@@ -38,14 +38,12 @@ var TR_HEIGHT	=	19;
 
 var dxSTable = function() 
 {
-	this.rows = 0;
 	this.rowdata = {};
 	this.rowIDs = [];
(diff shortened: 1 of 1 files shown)


anthonyryan1:
Working on a pull request for this.

PR #2836, opened 2024-12-23
Define s-table header styles using variables

jevenski:
- Define header background with CSS gradient or plain color.
- Control sorting icons using CSS class names.
- Define sorting icons with inline base64 strings and remove binary image files.

Screen Recording 2024-12-23 215940


css/stable.css
```diff
@@ -8,23 +12,59 @@
...
-.stable td:first-child div {
-  display: flex;
-  align-items: center;
-  gap: 0.35rem;
+.stable td {
+  font: menu;
+  font-family: inherit;
+  font-size: inherit;
+  white-space: nowrap;
+  padding: 0.15rem;
```
stickz, line 20: Consider `padding: 0.15rem 0.25rem;`

stickz (changes requested):
I'm okay with the font changes on the default theme. It actually looks bigger and better. We just need to increase the top and bottom padding, so it doesn't look like one big blob of text.

koblack:
While we're at it, maybe add an ellipsis when the text is too long and overlaps desc/asc icon if possible.
I'm not sure if this is the right direction. Maybe there's a better solution?

something like `"Ratio/mon..."`
Screenshot

css/stable.css
jevenski: Just for conformation, perhaps you mean `padding: 0.25rem 0.15rem` if we're going to increase the top and bottom padding a bit?

jevenski:
> While we're at it, maybe add an ellipsis when the text is too long and overlaps desc/asc icon if possible. I'm not sure if this is the right direction. Maybe there's a better solution?
> 
> something like `"Ratio/mon..."` 

This is a good idea to implement, but it also brings up another question. In order to "ellipsis" the text overflow, we'll need to wrap the text in a new HTML element, and this will make the text alignment in table headers more complex. To simplify things, @stickz  shall we set the text in table headers always centrally aligned, instead of dynamically setting different alignment for different columns (align to left for text columns, align to right for numeric columns). The cells in the table body are still aligned in this way. Only the cells in the table header are set to align centrally.

css/stable.css
stickz: Yes, that is correct.

PR #2837, opened 2024-12-25
Improve s-table column resizing

jevenski:
- Resize columns in real time as user drags and moves header resizer.
- Fix column resizing separator display error.
- Disable column resizing and moving on touch devices.

Screen Recording 2024-12-25 215850


stickz:
Thanks!

PR #2839, opened 2024-12-31
Control alternate row color using CSS

jevenski:
- Control s-table alternate row color using CSS.
- Define odd row, even row, and active row background using variables.
- Define highlight row background using color filter with a brightness of 0.75.


stickz:
Could we remove the highlight row colour when hovering of rows? It's blurring text. Also, I don't think it's needed.

jevenski:
Oblivion and MaterialDesign had this highlight effect on hovering. Do you think we can remove that for consistency among different themes?

stickz:
> Oblivion and MaterialDesign had this highlight effect on hovering. Do you think we can remove that for consistency among different themes?

No, you can keep the highlighting effect on hovering on Oblivion and MaterialDesign. It looks likes fine and people have a choice to use them. It's just the other themes (especially the default one) I'm concerned about.

PR #2841, opened 2025-01-03
Control alternate row color using CSS

jevenski:
- Control s-table alternate row color using CSS.
- Define odd row, even row, and active row background using variables.
- Define highlight row background using color filter with a brightness of 0.75 for MaterialDesign theme only.

This is an amended version of https://github.com/Novik/ruTorrent/pull/2839, which I accidentally closed, and didn't find a way to reopen. Background highlight on hovering is now only enabled in MaterialDesign theme. A variable is created for every theme though, for the consistency among all themes.


stickz:
Thanks!

PR #2846, opened 2025-01-08
Check-in the npm lockfile for tests

selfisekai:
This is not a library, it should be here for reproducibility


stickz:
> This is not a library, it should be here for reproducibility

I'm happy to add any libraries that help test ruTorrent. We just want to keep dependences in core to an absolute minimum.

PR #2851, opened 2025-01-10
Improve s-table header row hover event handler

jevenski:
- Previously the modified event handler was attached separately to every single cell in s-table header. The new create method now attaches one handler to the entire header row and does the math to find out the index number of the currently targeted cell. This simplifies the creation of s-tables.
- In the process of s-table initialization, this event handler registration is also moved to the `dxSTable.assignEvent()` method.


stickz:
Thanks!

PR #2855, opened 2025-01-16
Drag and drop s-table header using DnD class

jevenski:
This is a big change that removes the `dxSTable.ColumnMove` object that basically did the same things as the `DnD` utility class, and handles resizing, moving, and sorting actions in one `DnD` object. After this commit, I think we are ready to make the final movement that merge the s-table header and the body into one single HTML table.

- Modify `DnD` class initialization method a bit, so that the object can be attached to an HTML element or an `id` identifier. This is a preparation for the following change.
- Merge header cell resizing and moving methods into one object, reusing ruTorrent's `DnD` utility class, and thus removes the `dxSTable.ColumnMove` object defined in the s-table module that did the same thing as the `DnD` class.
- Rearrange header row drag-and-drop event handling process. Resizing, moving, and sorting actions are distributed according to the position of mouse clicks and displacement of the mouse cursor.
- Remove obsolete properties from S-Table class (`dxSTable.cancelMove` and `dxSTable.cancelSort`).

### DIFF
--- a/js/objects.js
+++ b/js/objects.js
@@ -5,8 +5,18 @@
 
 // Drag & Drop object 
 class DnD {
-	constructor(id, options) {
-		this.obj = $('#' + id);
+	/**
+	 * Construct a drag-and-drop instance.
+	 * @param {string | HTMLElement} dndObj Reference to the drag-and-drop element.
+	 * Can be the HTML `id` of the element, or a reference to the HTML element.
+	 * @param {Object} options Drag and drop options to be passed.
+	 */
+	constructor(dndObj, options) {
+		if ($type(dndObj) === "string") {
+			this.obj = $('#' + dndObj);
+		} else {
+			this.obj = $(dndObj);
+		}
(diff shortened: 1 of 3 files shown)


js/objects.js
```diff
@@ -5,8 +5,18 @@
...
+	 * @param {string | HTMLElement} dndObj Reference to the drag-and-drop element.
+	 * Can be the HTML `id` of the element, or a reference to the HTML element.
+	 * @param {Object} options Drag and drop options to be passed.
+	 */
+	constructor(dndObj, options) {
+		if ($type(dndObj) === "string") {
+			this.obj = $('#' + dndObj);
+		} else {
+			this.obj = $(dndObj);
+		}
```
stickz, line 19: `this.obj = $type(dndObj) === "string" ? $('#' + dndObj) : $(dndObj);`

stickz (changes requested):

Issue #2794, opened 2024-11-14
Torrent table row header not always aligned with column


koblack:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Link provided to install script if applicable
- [X] Not using broken rtinst install script
- [X] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Web Browser: Google Chrome 130.0.6723.117
ruTorrent: master: https://github.com/Novik/ruTorrent/commit/c6caa699ebd79dc4fb5130d64c027ad0dca6dcf7
PHP: PHP 7.4.3
OS: Ubuntu 20.04.6 LTS
web server: lighttpd/1.4.55

### Tell us how you installed ruTorrent

git clone https://github.com/Novik/ruTorrent

### Describe the bug

When we scroll the torrent table all the way to the right, the header row is not aligned with the column.

![ezgif-2-e051a7cb57](https://github.com/user-attachments/assets/2f7c508c-0e5e-4100-8a1d-0d5019dcdc5e)

### Steps to reproduce

Scroll the torrent table all the way to the right.

### Expected behavior

The header row is aligned with the column.

stickz:
@jevenski once this is fixed, we can post Beta 5. And possibly the stable `v5.1` release!

jevenski:
This issue existed before v5, and is because of the internal structure of `s-table's. The table header and the table body belong to two separate HTML `table`s. I'm afraid this won't be easily fixed in with one line or two.

koblack:
Thanks for checking this. Let's leave it as is for now and move forward.

stickz:
Alright, target adjusted for `v5.2`. ruTorrent `v5.1 Beta 5` released. https://github.com/Novik/ruTorrent/releases/tag/v5.1-beta5

PR #2860, opened 2025-01-21
Merge stable head and body into one table

jevenski:
- Merge stable-head and stable-body into one HTML table, and thus removes the process to sync horizontal scroll position.
- Fix table header split away from table body when scrolled to right end (https://github.com/Novik/ruTorrent/issues/2794).
- Adjust internal styles of the s-table to fit the new table structure.

Fixes: https://github.com/Novik/ruTorrent/issues/2794

### DIFF
--- a/css/stable.css
+++ b/css/stable.css
@@ -8,6 +8,7 @@
   --row-even-bg-color: #ffffff;
   --row-active-bg-color: #cfdeef;
   --row-highlight-brightness: 1;
+  position: relative;
   height: 100%;
   background-color: #FFFFFF;
   overflow: hidden;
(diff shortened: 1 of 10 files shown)


koblack:
Can't sort table on mobile device. When I tap on header nothing happens.
Tested on jevenski/merge-stable branch with chrome/andoroid.

On desktop chrome/windows everything seems to work fine.

jevenski:
Alright... This patch is still buggy with other header dragging issues. Needs a little more tuning.

koblack:
Thanks for the quick fix, everything seems to be fine now.

PR #2875, opened 2025-03-22
Add danger color to #randomize_hash

anthonyryan1:
The torrent add dialog #tadd remembers previous settings now to help users avoid needing to set the same setting repeatedly.

Users also assume that default settings are always safe settings for general usage, and click through without reviewing them.

A user who checks #randomize_hash for one torrent (deliberately or otherwise), then is surprised that none of their torrents download anymore, because #randomize_hash is still checked on all future torrent additions.

Because users may legitimately want this setting to persist through several subequent torrent adds, we don't want to remove persistence for this field. But I think we should draw more attention to its use, because accidental usage can cause the entire client to stop functioning as expected (from a user perspective).


stickz:
Thanks, confirmed working.

randomize_hash_danger

PR #2880, opened 2025-04-03
Improve $tempDirectory handling:

rozhuk-im:
- make sure that contain trail slash
- add extra checks to system provided temp dirs detection
- add '/var/tmp' to system provided temp dirs
- add dir exist check and create for user provided and user profile temp dir
- add comments to code

This change is for be sure that setting ```$tempDirectory = '/tmp';``` will not silently break installation.


stickz:
Could this be submitted to develop? It doesn't qualify for a hotfix. We need to signal a minor version bump.

rozhuk-im:
done

rozhuk-im:
Thanks!

PR #2889, opened 2025-04-14
Diskspace option to display free bytes

TheGoblinHero:
Currently diskspace plugin shows colored meter with free % on it. Meter tooltip includes free/total bytes, but it is not convenient for me. I want to know whether I can start downloading those 200 GiB of remuxes without tooltip opening.

Here's a solution. Now you can change plugin config to display free bytes instead of %.
meter2


plugins/diskspace/conf.php
```diff
@@ -4,3 +4,4 @@
 $notifySpaceLimit = 512;	// in Mb
 $partitionDirectory = null;	// if null, then we will check rtorrent download directory (or $topDirectory if rtorrent is unavailable)
 				// otherwise, set this to the absolute path for checked partition. 
+$freeBytesInMeter = 0;		// show free bytes in meter instead of %
```
stickz, line 7: Please use true or false value here to avoid confusion.

stickz (commented):
Please create a new branch based on develop and open a new pull request. This code is outdated. It is reverting previous pull requests.

stickz:
I will definitely accept this change. It just needs to be fixed up.

PR #2892, opened 2025-04-25
Log tab saving to log file and loading 10 last messages from it if any.

ranirahn:
Sorry for messing up pull request. I dont know how to properly do these things. But what this does is loading last 10 messages from .log file that is set up in "config/config.php" file. And any message that get send to Log tab with noty() function gets this message saved to .log file. So any time you refresh the rutorrent you see last 10 message from that file.


stickz:
I like this feature. Thanks @ranirahn. Can you save the log output to a `.dat` file instead?

I would like to help you improve the implementation. I'm going to give you an example of our rCache class from the cpuload plugin. https://github.com/Novik/ruTorrent/blob/master/plugins/cpuload/cpu.php

The only required difference is to populate an array instead. This is called an instance varriable in the class. All instance variables are automatically saved as part of the rCache class. All you need to do is create and populate them.

https://github.com/Novik/ruTorrent/blob/1029e8afa40a233c93ee6813733269d397141bd7/plugins/cpuload/cpu.php#L10

ranirahn:
So this should be plugin like this CPU one? I will try to figure this out, but that takes a lot of time for me.

stickz:
> So this should be plugin like this CPU one? I will try to figure this out, but that takes a lot of time for me.

One step at a time. I would highly recommend to try the rCache implementation first. You can do that without creating a plugin.

ranirahn:
I think i did it already :D

ranirahn:
Found little mistake.

ranirahn:
Added Stopped State also with this pull.

stickz:
Could you create a new pull request instead for stopped states? And anything else unrelated to logging? 

This will get changed merged faster. It will also help to document changes if there is a problem we miss.

ranirahn:
And also added my fix for RSS feeds not loading problem. Now i have added all my fixes in at once :D

ranirahn:
> Could you create a new pull request instead for stopped states? And anything else unrelated to logging?
> 
> This will get changed merged faster. It will also help to document changes if there is a problem we miss.

Oh man. I did not know why it added everything in this one pull, but it did not really give me any option. So how do i do that one by one separately?

PR #2893, opened 2025-04-25
Adding Stopped state

ranirahn:
The new Stopped state will have torrents that
1) Are in downloading status, but nothing have arrived yet and no data moving up or down. Basically dead torrent without seeds.
2) Torrents that are half way, but no data is moving.
3) Paused torrents. Its same thing as half way done, but no traffic.

Completed state have all torrents that are 100% done. Does not matter if they are Finished or Seeding Status.

Downloading State have all torrents that have not started yet. And torrents that are started and have traffic.
The question is. Do this state have only the second type torrents? And all not started torrents will be in Inactive state? Or should there be another state for Waiting torrents? Right now this Downloading State have 1000 not started torrents and nothing else. For me it would be fine like this, but what you guys think?

### DIFF
--- a/index.html
+++ b/index.html
@@ -162,6 +162,7 @@
 							<panel-label id="pstate_all" icon="all" uilangtext="All" selected></panel-label>
 							<panel-label id="-_-_-dls-_-_-" icon="down" uilangtext="Downloading"></panel-label>
 							<panel-label id="-_-_-com-_-_-" icon="completed" uilangtext="Finished"></panel-label>
+							<panel-label id="-_-_-wfa-_-_-" icon="paused" uilangtext="Stopped"></panel-label>
 							<panel-label id="-_-_-act-_-_-" icon="up-down" uilangtext="Active"></panel-label>
 							<panel-label id="-_-_-iac-_-_-" icon="inactive" uilangtext="Inactive"></panel-label>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -162,6 +162,7 @@ <h4 class="offcanvas-title" id="offcanvas-sidepanel-label"></h4>
 							<panel-label id="pstate_all" icon="all" uilangtext="All" selected></panel-label>
 							<panel-label id="-_-_-dls-_-_-" icon="down" uilangtext="Downloading"></panel-label>
 							<panel-label id="-_-_-com-_-_-" icon="completed" uilangtext="Finished"></panel-label>
+							<panel-label id="-_-_-wfa-_-_-" icon="paused" uilangtext="Stopped"></panel-label>
```
stickz, line 165: Please update the files in `/lang` with the new phrase.

stickz (changes requested):

ranirahn:
I think its already existing in Lang

```javascript
Stopped			: "Stopped",
``` 
This does not work?

stickz:
> I think its already existing in Lang
> 
> ```js
> Stopped			: "Stopped",
> ```
> 
> This does not work?

You are correct. Thank you. Merged.

PR #2896, opened 2025-04-25
Saving Log tab into logs.dat and loading 10 last message on loading rutorrent

ranirahn:
Just cleaned up the pull request. Now it should be good to go.


gashtal:
Any change the log size could be made configurable through the ruTorrent config file?

ranirahn:
> Any change the log size could be made configurable through the ruTorrent config file?

Yes it could. Dont know what to name these in the config file. If i figure that out i can add them in config and change little log.php code and done.

ranirahn:
Now it can be changed from config.php file.

php/log.php
```diff
@@ -0,0 +1,84 @@
...
+require_once('cache.php');
+require_once('../conf/config.php');
+
+class LogHandler
+{
+    public $hash = 'logs.dat';
+    public $logs = [];
+    public $max_entries = 100;
+    public $log_count = 10;
+
```
stickz, line 11: Please add `public $modified = false;` this is required for the rCache class.

php/log.php
```diff
@@ -0,0 +1,84 @@
+<?php
+require_once('cache.php');
+require_once('../conf/config.php');
+
+class LogHandler
+{
+    public $hash = 'logs.dat';
```
stickz, line 7: Please change name of `.dat` file to something less ambiguous like `log_history.dat`.

php/log.php
```diff
@@ -0,0 +1,84 @@
...
+
+            $response = $handler->saveLog($message, $status);
+            header('Content-Type: application/json');
+            echo json_encode($response);
+        } else {
+            $response = $handler->getLatestLogs();
+            header('Content-Type: application/json');
+            echo json_encode($response);
+        }
+    }
```
stickz, line 43: Please use our `JSON::safeEncode($response);` wrapper.

Please bring these lines outside and bellow the if statement block. We don't need to repeat them.

```
header('Content-Type: application/json');
JSON::safeEncode($response);
```

js/webui.js
```diff
@@ -2546,6 +2546,7 @@ $(document).ready(function()
 	Promise.all([
 		import('./backgroundtask.js'),
 		import('./category-list.js'),
+		import('./log.js'),
```
stickz, line 2549: Please change the name of this file to something less ambiguous like `log-history.js`.

conf/config.php
```diff
@@ -29,6 +29,8 @@
 	$al_diagnostic = true;			// Diagnose auto-loader. Set to "false" to make composer plugins work.
 
 	$log_file = $_ENV['RU_LOG_FILE'] ?? '/tmp/errors.log'; // path to log file (comment or leave blank to disable logging)
+	$LogTab_max_entry = 100;		// For saving Log Tab content to logs.dat file.
+	$LogTab_count = 10;			// For loading logs to Log Tab on rutorrent start.
```
stickz, line 33: Please include option to enable the feature with true or false value as well.

php/log.php
ranirahn: changed that file also if i was changing everything else log_history then why not this one and added this modified.

php/log.php
ranirahn: done

php/log.php
ranirahn: should be done.

js/webui.js
ranirahn: done

conf/config.php
ranirahn: changed this to array and in log_history.php added little code in the beginning that if this does not exist it does nothing. so if commenting out its disabled. had to fix log_history.js also to not do anything if log_history.php is disabled. Should be working now.

PR #2897, opened 2025-04-26
Saving Log tab into logs.dat and loading 10 last message on loading rutorrent

ranirahn:
Can be disabled in config.php by commenting it out.
Should this be disabled by default?


stickz:
No, it can be enabled by default thanks!

Issue #2638, opened 2024-03-03
Add "Stopped" state to the states


gashtal:
### Is your feature request related to a problem?

Currently, torrents that are "stopped" and not complete, and torrents that are in downloading state but not active (i.e. no data is being transferred due to lack of seeders) are both categorized under the "Downloading" state on the left panel. This is counterintuitive since the "stopped" torrents are not "downloading" and should not be under the "Downloading" state. Such torrents do actually have a different state (stopped) under the "status" column, as well. The current behavior prevents the user from quickly finding all downloading torrents that have no seeds and as such, are stuck, by looking at the "Downloading" category, since such torrents get buried among all the stopped (and incomplete) torrents.

### Describe the solution you'd like

The request is to add a new "Stopped" state under the "State" section in the left panel and have stopped torrents that are **_incomplete_** moved under that section. In terms of category prioritization, any torrent that is complete, whether stopped or not, should got to the "Completed" category. Any torrent that is stopped and incomplete, should go under the "Stopped" category. All torrents that are not stopped or paused and are not complete should then go under the "Downloading" category. Torrents that are complete but have an error such as some tracker error currently go to the "Error" category; these should probably be moved to the "Completed" category as well (a tracker error is not important when the torrent has already completed downloading), which should automatically happen if the Completed category is given the highest priority.

### Additional context

I have a few thousand torrents in RuTorrent, the majority of which are in stopped state and have not been started due to lack of space. As space becomes available, I download and clear the torrents. Since, currently, all the stopped torrents are categorized under the "Downloading" category, it is very difficult to find torrents that are actually downloading but stuck due to lack of seeds, as they get buried among all the stopped torrents in that category.

ranirahn:
I would rather see "Paused" state. From my point of view, all Stopped and not completed torrents already are in one place. That is "Downloading" state. But I think there could be even more state options available. "Stopped", "Paused", "Finished", "Incomplete", and so on. I would separate them all. Then probably anyone could find a suitable option from there for any occasion.

gashtal:
Bump so that it doesn't get closed as stale.

ranirahn:
I think I will take it on when I finish what I am doing right now.

ranirahn:
Done with this
https://github.com/Novik/ruTorrent/pull/2892

ranirahn:
Now its this one
https://github.com/Novik/ruTorrent/pull/2893

PR #2899, opened 2025-04-28
Update "State" categories to match "Status" column

gashtal:
Fix for issue #2638 

Updates logic that determines "State" groups to match the states in the "Status" column of the torrent list:

- Downloading: Include all "started" torrents. Matches the "Downloading" state in the "Status" column.
- Completed: Include all "finished" torrents. Matched the "Finished" state in the "Status" column.
- Stopped: Match all torrents paused or stopped by the user, except ones that are "Completed". i.e. will includes torrents that are either not started at all, or have been partially downloaded and then stopped or paused by the user.
- Active: Only match "Downloading" torrents that have upload/download activity.
- Inactive: Only match "Downloading" torrents that do not have upload/download activity.
- Error: Match all torrents that have an error, except "Completed" ones. Completed torrents only ever have tracker errors which don't matter when the torrent has already completed downloading.

Also swap icons for "Inactive" and "Stopped" states. Now the "Stopped" state has the same icon as stopped torrents in the torrent list. The "Inactive" state now uses the "Paused" icon which is not correct, but there is no dedicated icon for "Inactive" state as of right now. Maybe the current "Active" icon with green background would work better for the "Active" state, and the same icon with red background would work for "Inactive" state.

Ideally we would want to have a dedicated "Paused" state, as well, since there is such state defined in the torrent list already. It also probably makes sense to rename the "Completed" state to "Finished", to match the state in the status column in the torrent list.

### DIFF
--- a/index.html
+++ b/index.html
@@ -162,9 +162,9 @@
 							<panel-label id="pstate_all" icon="all" uilangtext="All" selected></panel-label>
 							<panel-label id="-_-_-dls-_-_-" icon="down" uilangtext="Downloading"></panel-label>
 							<panel-label id="-_-_-com-_-_-" icon="completed" uilangtext="Finished"></panel-label>
-							<panel-label id="-_-_-wfa-_-_-" icon="paused" uilangtext="Stopped"></panel-label>
+							<panel-label id="-_-_-wfa-_-_-" icon="inactive" uilangtext="Stopped"></panel-label>
 							<panel-label id="-_-_-act-_-_-" icon="up-down" uilangtext="Active"></panel-label>
-							<panel-label id="-_-_-iac-_-_-" icon="inactive" uilangtext="Inactive"></panel-label>
+							<panel-label id="-_-_-iac-_-_-" icon="paused" uilangtext="Inactive"></panel-label>
 							<panel-label id="-_-_-err-_-_-" icon="error" uilangtext="Error"></panel-label>
 						</category-panel>
 						<category-panel id="plabel" uilangtext="Labels">
--- a/js/category-list.js
+++ b/js/category-list.js
@@ -23,23 +23,26 @@ export class CategoryList {
     this.statistic = CategoryListStatistic.from("pview", this.viewSelections, {
       pstate: [
         (_, torrent) => [
-          torrent.done >= 1000
-            ? "-_-_-com-_-_-"
-            : (torrent.state === 1 && torrent.done === 0 && torrent.dl === 0 && torrent.ul === 0)
-            ? "-_-_-wfa-_-_-"
-            : (torrent.done > 0 && torrent.dl === 0 && torrent.ul === 0)
-            ? "-_-_-wfa-_-_-"
-            : (torrent.done === 0 && torrent.dl === 0 && torrent.ul === 0)
-            ? "-_-_-dls-_-_-"
-            : (torrent.dl > 0 || torrent.ul > 0)
-            ? "-_-_-dls-_-_-"
-            : "-_-_-dls-_-_-",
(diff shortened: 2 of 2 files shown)


ranirahn:
I dont like that change at all. I add torrents all Stopped Status. Then i look over and select what to download. The stopped was meant to have torrents that have problems like its state is downloading but there is no seeds so traffic is 0 and size is 0 you will never find these torrents with your change. All torrents that you stop half way go to stopped State so you can easily find them and start again. Or torrents that have not completed but have no errors. Stopped torrents have no errors, so you will never find these if you have thousands of torrents like me.

ranirahn:
You opened the issue Year ago and did absolutelty NOTHING with it and now that someone did something with it you complain and change it to worse? Why did you not push that change yourself year ago? People would have time to test it and change it back or forward. I change it and day later you have the fix? Maybe take a week to test the change out?

gashtal:
You are taking this too personally, mate. It is up to the owner of the project whether to accept and merge this PR or not, not me. Just because I have sent a PR doesn't mean it is going to be merged. This is just a bunch of if/else conditions. There no need to "test" this to see how it works; it is evident from the logic of the code how it will work.

ranirahn:
Its not personal. I am literally defending the ORIGINAL code. active and inactive already have clear purpose but YOU want to change it. Error already have very clear purpose and you want to changet that too. If you dont like the new State it should go back to original and forget the addition. That way it have no problems. Anyway i added my PR and I let people to choose witch one they like better. Eighter way i will use my code no matter what happens because your change makes no sense at all and then could just delete the new State and nothing would change. So original was 100x better then your change.

stickz:
Thanks for the comments and screenshots explaining the pull request. I removed a few of them that didn't add any value.

@ranirahn, I'm merging this pull request. The reason is because there are less torrent states. It will be easier for the user to understand. Simplicity is the reason, not clarity. Thank you for the contributions. I will continue to review and merge your code!

ranirahn:
Well, I will never use it. I rather copy from old version back without any change at all. Literally no reason to have rutorrent figure out what torrents have error now. Zero reason. So because of this reason alone I haveto replace these 2 files on ANY new update. But thats just me. Lets see if there is anyone else in the future who dont like torrents with errors lost in the void.

GrumpyBear57:
I agree with @ranirahn, completed torrents should still show in the 'Error' tab. This is now just as useless as qBit is for finding which torrents from my private tracker have been removed from the tracker, and should be removed from the client.

Just because a torrent is completed, doesn't mean I don't care about the status of it outside of 'yup, it's complete'. There's no point in keeping a torrent in my client if it's not going to do anything.

stickz:
> I agree with @ranirahn, completed torrents should still show in the 'Error' tab. This is now just as useless as qBit is for finding which torrents from my private tracker have been removed from the tracker, and should be removed from the client.
> 
> Just because a torrent is completed, doesn't mean I don't care about the status of it outside of 'yup, it's complete'. There's no point in keeping a torrent in my client if it's not going to do anything.

I responded to the issue report. If someone would like to make this feature optional, I will accept the pull request.

ranirahn:
I will not touch that. I have my own change in the code for this so for me its all working. I dont even know how to make something optional and I have already spend too much time for this already and I regret wasting my time for this in the first place. Original was fine. I wanted to add something without changing already good parts, but now its so much worse compared to original so yeah, I will never touch that again.

Abzie:
This is such a bad change, I agree with @ranirahn and @GrumpyBear57 - People do care what happens to torrents after they're completed.

This needs to be reverted.

n1ce-0ne:
another reason to stay on the older versions imo
"Simplicity is the reason, not clarity" - id rather be clear on whats going on

Issue #2637, opened 2024-02-26
RSS plugin should use guid for history management


de666:
### Please complete the following tasks.

- [X] Web browser cache cleared
- [X] Link provided to install script if applicable
- [X] Not using broken rtinst install script
- [X] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

ruTorrent: v4.2.9

### Tell us how you installed ruTorrent

using
https://github.com/crazy-max/docker-rtorrent-rutorrent/releases/tag/4.2.9-0.9.8_2-0.13.8_2-r0

### Describe the bug

Using rss plugin with feed which sometimes changes final link maintaining the same guid (like [Jackett](https://github.com/Jackett/Jackett) does), history is not managed the correct way because it refers to final download url (link element) instead of guid which (if present and valid) shuold be used as history reference because is unique and it doesn't change.

### Steps to reproduce

1. add rss feed from jackett
2. load some torrent -> torrent are loaded and marked as loaded in the UI
3. wait a couple of hours
4. reload feed
5. torrent previous loaded not marked as loaded anymore since link element has changed (for the same torrents already loaded)

### Expected behavior

rss plgin shuold handle correctly the torrent already loaded.
This bug causes also rssmanager is not working properly because if the torrent has been already loaded, completed and removed from active torrent list, it is loaded again.

TrimmingFool:
@de666 Thanks for this!

Currently, the `<link>` is used to identify a rss item.
However, according to the spec, using the `<guid>` as the item-identifer instead, seems to be correct:
https://www.rssboard.org/rss-specification#hrelementsOfLtitemgt

I am not sure if only changing `rRSSHistory` will be enough, though: 
https://github.com/Novik/ruTorrent/blob/2d67a00b4fde90b4a652c130ed76ac2fc4383fb3/plugins/rss/rss.php#L301

If no one else wants to take this on, I think, I would, eventually.

stickz:
@TrimmingFool Version 4.3 will be released in the near future. Would you like to sneak this important change in?

ranirahn:
![ss (2024-04-26 at 01 13 18)](https://github.com/Novik/ruTorrent/assets/34799179/2ce2eb2f-13e8-4fc7-96c6-7d5b576b2504)
Is this related to this problem that same torrent gets added over and over again? This stops only when this torrent falls out of RSS feed finally.

de666:
>  Is this related to this problem that same torrent gets added over and over again? 

[...]

PR #2902, opened 2025-04-28
Added checking torrents by guid before loading it again

ranirahn:
I added check torrents by guid if the torrent was loaded before. Because some torrents have links and dates changing sometimes but file itself stay the same so no it checks first by guid and there is no guid then it check the hash like before. At least this stopped my torrents loading over and over again that i have already deleted 100 times. I also tested addins check by title, but sometimes files have same names but coming from different sources so then these same name torrents from different feed get ignored. Maybe it is the exactly same thing but maybe its not so its better to use guid and not the title. I hope it fixes also this problem https://github.com/Novik/ruTorrent/issues/2637

### DIFF
--- a/plugins/rss/rss.php
+++ b/plugins/rss/rss.php
@@ -275,7 +275,7 @@ class rRSS
 			rTorrentSettings::get()->pushEvent( "RSSFetched", array( "rss"=>&$this ) );
 			if(!$this->hasIncorrectTimes())
 				foreach( $this->items as $url=>$item )
-					$history->correct($url, $item['timestamp']);
+					$history->correct($url, $item['timestamp'], $item['guid']);
 		}
 		return(true);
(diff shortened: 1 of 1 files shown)


plugins/rss/rss.php
```diff
@@ -318,20 +318,20 @@ public function __construct()
 		$this->version = 2;
 	}
 
-	public function add( $url, $hash, $timestamp )
+	public function add( $url, $hash, $timestamp, $guid = null, $title = null )
```
stickz, line 321: Avoid default parameters. Make everything follow the new function prototype. This will address unintentional or undefined behavior.

stickz (changes requested):

stickz:
Can you submit this PR to the develop branch instead?

PR #2904, opened 2025-04-28
Check torrents by guid before loading

ranirahn:
That does not add guid to history files. Its just checking by guid if torrent was already loaded. I had many errors like that and that have fixed it for me.


stickz:
Thanks. I will review this on the weekend.

PR #2905, opened 2025-04-28
Update for log_history.php

ranirahn:
For some reason it did not take values from config file and also did not save more logs if the log was full. So I had to rewrite this a little. Now I have tested it many ways and now the log is rotating and changing values in config work after reload instantly. Sorry, that i did not notice this problem before last PR. should have added that there aswell.


stickz:
No problem. This is the develop branch.

PR #2914, opened 2025-05-04
Add OblivionBlue theme

Abzie:
I have made a faithful recreation of the OblivionBlue theme that is compatible with ruTorrent v5.x.

I'm not versed in why OblivionBlue was not added to the main repo, so apologies in advance if not allowed.

Original credit to LesBleus for the original OblivionBlue theme, this theme is based off the updated v5 Oblivion theme.

Original OblivionBlue:
image

ruTorrent v5 OblivionBlue
image


stickz:
Thanks!

PR #2916, opened 2025-05-08
Docker image (#1)

nelu:
* docker image build


nelu:
using php 8.3 and apache as base image - UI only
/data mountpoint inside the container for persistent data (downloads, rtorrent session, rutorrent share and logs)
rtorrent is present in the image but doesnt start by default

 secrets: DOCKERHUB_USER 
 and DOCKERHUB_TOKEN 
 
 must be configured on the repo actions settings for pushing to dockerhub

Issue #2908, opened 2025-05-01
torrent in error not listed in "state" side panel


Neonlinx:
### Please complete the following tasks.

- [x] Web browser cache cleared
- [x] Link provided to install script if applicable
- [x] Not using broken rtinst install script
- [x] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Web Browser: Chrome 135/136
ruTorrent: 5.2.2
php: 8.1.27
OS: Debian 11

### Tell us how you installed ruTorrent

I'm using Quickbox

### Describe the bug

torrents that (for example) the tracker has deleted (i.e. "unregistered torrent") no longer appear as "in error" in the state list on the left side panel. This is annoying because when you have thousands of files, you have to scroll through the entire list to find the ones to delete. This problem occurs for any problem reported by the tracker (not only for "unregistered torrents"). And of course, in the "status" column these torrents are classified as OK, as normally in seed.

### Expected behavior

Torrent with any problems reported by the tracker need to be classified as "in error". Or you need to create a new flag, something like "to check", or "to verify".

JMSDOnline:
@Neonlinx , this was resolved after we downgraded xmlrpc-c from the downgrade of your `0.15.1` to `0.9.8`, correct?

Same with your #2907 ?

stickz:
> 
> Same with your [#2907](https://github.com/Novik/ruTorrent/issues/2907) ?

No, this is a ruTorrent related query. We removed the error state for completed torrents in ruTorrent `v5.2.2`.

@Neonlinx Thank you for your query. I will accept making this feature optional. But it will still be enabled by default.

Neonlinx:
@JMSDOnline Hi JM, I replied on #2907 

@stickz Ok, thanks!

ranirahn:
Well well well. As I said when this update was made that it is WRONG to remove torrents that have ANY errors to be in the Errors state. How was the explanation again. "I dont care about tracker errors" or something like that. So I replaced this with my fix and everything is perfect, but everyone else. Have fun.

Abzie:
Yep, I discovered I had <200 torrents in an errored state due to #2899

ranirahn:
> reverting [caac23d](https://github.com/Novik/ruTorrent/pull/2899/commits/caac23d721b24d145f2fa41b841bd8a4a3416bcd) restores the old behaviour of showing completed torrents in an errored state.

[...]

PR #2918, opened 2025-05-11
Fix the behavior of the "Error" state to include all torrents that have an error but are not stopped

gashtal:
Addresses #2908


stickz:
Thanks!

PR #2925, opened 2025-05-16
fix for labeling new torrents

ranirahn:
For some reason $$ cant get the values. Maybe in the future someone finds better solution why its not working, but directly getting the value with getElementById works.

### DIFF
--- a/js/content.js
+++ b/js/content.js
@@ -230,15 +230,15 @@ function makeContent() {
 			req.push('not_add_path=1');
 		if($("#randomize_hash").prop("checked"))
 			req.push('randomize_hash=1');
-		var dirEle = $$("#dir_edit");
+		var dirEle = document.getElementById('dir_edit');
 		if (dirEle) {
-			var dir = dirEle.val().trim();
+			var dir = dirEle.value.trim();
 			if(dir.length)
 				req.push('dir_edit='+encodeURIComponent(dir));
 		}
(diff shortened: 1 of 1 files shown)


js/content.js
```diff
@@ -230,15 +230,15 @@ function makeContent() {
 			req.push('not_add_path=1');
 		if($("#randomize_hash").prop("checked"))
 			req.push('randomize_hash=1');
-		var dirEle = $$("#dir_edit");
```
tohenk, line 233: Just remove `#` which is not needed, it should `var dirEle = $$("dir_edit");`.

js/content.js
```diff
@@ -230,15 +230,15 @@ function makeContent() {
...
+		var dirEle = document.getElementById('dir_edit');
 		if (dirEle) {
-			var dir = dirEle.val().trim();
+			var dir = dirEle.value.trim();
 			if(dir.length)
 				req.push('dir_edit='+encodeURIComponent(dir));
 		}
-		var lblEle = $$("#tadd_label")
```
tohenk, line 239: This too.

js/content.js
ranirahn: tested and it works, but val() did not work like it was before.

js/content.js
ranirahn: also done.

stickz:
Thanks!

PR #2927, opened 2025-05-18
Supports relative path in symlink.

tohenk:
As symbolic link can be relative, this patch resolve the path before handing it for permission check.


stickz:
Thanks!

PR #2930, opened 2025-05-19
Improve diskspace usage tooltip.

tohenk:
The tooltip for diskspace is mislead.

Before:
image

After:
image

### DIFF
--- a/plugins/diskspace/init.js
+++ b/plugins/diskspace/init.js
@@ -3,14 +3,20 @@ plugin.loadMainCSS();
 
 plugin.setValue = function( full, free )
 {
-        var percent = iv(full ? (full-free)/full*100 : 0);
-        if(percent>100)
-	        percent = 100;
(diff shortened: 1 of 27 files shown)


gashtal:
The original design might have been misleading, but the new one is unnecessarily cluttering the tooltip with "used space" amount which has no significance, and puts the "free space" amount which is the most important number, and is what everyone looking at the tooltip is actually looking for, at the far end of the tooltip. Simply changing the tooltip to show "10.6 TiB free of 18.0 TiB", just like Windows Explorer does, would have eliminated the confusion, without cluttering the tooltip or changing the location of the "free space" amount inside the tooltip. I was wondering why I kept looking for the free space amount at the beginning of the tooltip text while it is at the end, after upgrading from 5.2.10 to 5.3.1, until I realized the design had been changed as a part of this PR.

image

PR #2932, opened 2025-05-22
fix for icons not getting added to torrent

ranirahn:
When adding torrents with magnet links in stopped state and starting this torrent will download meta data and change name from hash.meta to torrent name and stay in stopped state, but because icon was changed before name change the icon was added to last DOM element and now the torrent dont have icon in torrent list or in RSS list. If we do icon change after then its always there. If changing the view that this torrent is not in the list and change back where it is the icon will appear, but with this change its always there. Did not change the code, just reorder the IF statements.


stickz:
Thanks!

PR #2948, opened 2025-06-05
Allows to use known icons instead of duplicate it

tohenk:
Instead of duplicate the same icon, we now can alias it to known icons. Both label and tracker can benefit from this behavior. As an example, `example.com` tracker using this domain as the website, while the tracker itself is served from `example.net` or `example.org` domain.

```
<?php

$labelAliases = [
  'example.com' => ['example.net', 'example.org'],
];
```

Also tracker icon now updated to look for either `favicon.ico` or `favicon.png`.


stickz:
Awesome, thanks!

PR #2949, opened 2025-06-05
Allows both count and size part to be right aligned

tohenk:
This patch adds a separator between count and size with `/`, allows both to be right aligned, honor the selected state and fix Acid theme selected color to be contrast with background.


stickz:
Thanks!

PR #2960, opened 2025-06-16
Fix Chrome performance regression

anthonyryan1:
Now that dxSTable.rows is a getter, and not a variable, we need to be mindful about using it in for loops. On an ruTorrent with 20K torrents, putting it into the for loop, instead of copying it into a variable will result in 20K executions of the getter function and tank performance.

By checking the rowcount once, before we start the loop, we can keep the getter (reducing the risk of OOS bugs), without suffering the performance penalty.

Related to #2830


stickz:
Thanks @anthonyryan1. I will squash and merge. There's a button for that on GitHub.

PR #2963, opened 2025-07-09
feat(check_port): Implement dual IPv4/IPv6 port status checking

cantalupo555:
- Detect public IPv4 and IPv6 addresses using dedicated APIs (ipify.org).
- Perform port status checks for both IPv4 and IPv6 protocols independently.
- Update the backend (action.php) to return a JSON response containing status for both protocols.
- Revise the frontend to display separate status icons in the ruTorrent status bar, with each IP address shown next to its respective icon.
- Update the tooltip on the status bar pane to show detailed port status for both IPv4 and IPv6.
- Improve visual layout with proper spacing and alignment for status indicators.

### DIFF
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -1,142 +1,194 @@
 <?php
 require_once( dirname(__FILE__)."/../../php/settings.php" );
 require_once( dirname(__FILE__)."/../../php/Snoopy.class.inc" );
+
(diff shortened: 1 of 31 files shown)


cantalupo555:
Sorry, I didn't know that renaming the branch ~~deleted~~ the pull request.

Can restore the previous branch.

PR #2964, opened 2025-07-10
feat(check_port): Conditionally render protocol status display

cantalupo555:
The plugin now offers two ways to conditionally hide the status display for a specific IP protocol (IPv4 or IPv6), providing a cleaner and more adaptive interface.

1.  **Manual Hiding via Configuration:**
    * Users can set `$useWebsiteIPv4` or `$useWebsiteIPv6` to `false` in `conf.php` to completely hide the corresponding protocol status from the UI.
    * The backend now passes `use_ipv4` and `use_ipv6` flags to the frontend to control visibility based on this setting.

2.  **Automatic Hiding for Unavailable Protocols:**
    * If a protocol is enabled in the configuration but detected as unavailable on the server (e.g., a public IP cannot be found), its status is now automatically hidden.
    * This prevents the UI from showing a persistent "unknown" status for an unusable protocol.
    * More accurate interface without requiring manual configuration changes from the user.

### DIFF
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -174,6 +174,10 @@ $response = [
 	"ipv6" => "-", "ipv6_port" => (int)$port, "ipv6_status" => 0,
 ];
 
+// Add enabled flags to the response
+$response['use_ipv4'] = ($currentUseWebsiteIPv4 !== false);
+$response['use_ipv6'] = ($currentUseWebsiteIPv6 !== false);
+
 // Perform the IPv4 check if it's enabled in conf.php
 if ($currentUseWebsiteIPv4 !== false) {
 	$ipv4_result = get_and_check_ip('4', $currentUseWebsiteIPv4, $ip_glob, $port, $currentCheckPortTimeout);
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -168,6 +168,10 @@ function get_and_check_ip($ip_version, $use_website, $rtorrent_ip, $rtorrent_por
 		// Call the appropriate checking function based on the selected service
 		if ($use_website == "yougetsignal") {
 			$status = check_port_yougetsignal($ip_to_check, $rtorrent_port, $timeout);
+			// If yougetsignal fails, fall back to portchecker for IPv4
+			if ($status === 0 && $ip_version == '4') {
+				$status = check_port_portchecker($ip_to_check, $rtorrent_port, $timeout);
+			}
 		} elseif ($use_website == "portchecker") {
 			$status = check_port_portchecker($ip_to_check, $rtorrent_port, $timeout);
 		}
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -42,9 +42,21 @@ function get_public_ip($version, $timeout) {
 		if (filter_var($ip, FILTER_VALIDATE_IP, $flag)) {
 			return $ip; // Return the valid IP
 		}
-		error_log("check_port plugin: {$url} returned invalid IP: " . $ip);
+		// If looking for IPv6 and received a valid IPv4, it indicates no IPv6 is available (e.g. NAT64/DNS64).
+		// If looking for IPv4 and received a valid IPv6, it indicates no IPv4 is available.
+		// These are not error conditions, so we can return null without logging.
+		$isIPv6RequestWithIPv4Response = ($version == '6' && filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4));
+		$isIPv4RequestWithIPv6Response = ($version == '4' && filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6));
+
+		if (!$isIPv6RequestWithIPv4Response && !$isIPv4RequestWithIPv6Response) {
+			error_log("check_port plugin: {$url} returned invalid IP: " . $ip);
+		}
 	} else {
-		error_log("check_port plugin: Failed to fetch from {$url}. Status: {$snoopy->status}, Error: {$snoopy->error}");
+		// cURL error 7 ("Failed to connect to host") indicates a lack of connectivity for the requested IP version.
+		// This is not a plugin error, so we suppress the log for both IPv4 and IPv6 checks.
+		if ($snoopy->status != 7) {
+			error_log("check_port plugin: Failed to fetch from {$url}. Status: {$snoopy->status}, Error: {$snoopy->error}");
+		}
 	}
 	return null; // Return null on failure
 }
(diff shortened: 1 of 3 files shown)


cantalupo555:
As reported in the commit description, the value `false` within parameters `$useWebsiteIPv4` and `$useWebsiteIPv6` are working again as they should.

Screenshot from 2025-07-09 23-03-48
Screenshot from 2025-07-09 19-25-49


---

Here it seems to need a little tweaking.
Screenshot from 2025-07-09 23-03-48

cantalupo555:
~~New commit: https://github.com/Novik/ruTorrent/pull/2964/commits/bf238d0869b4c50467022cae1dcdcecb339f7320~~ **(git amend applied)**

The plugin now intelligently adapts the UI based on the server's network environment.

If a protocol (IPv4 or IPv6) is enabled in the configuration but is detected as unavailable on the server (i.e., a public IP cannot be retrieved), its status icon will now be automatically hidden.

This prevents the UI from showing a persistent "unknown" status for an unusable protocol and provides a cleaner, more accurate interface without requiring manual configuration changes from the user.

stickz:
I get the following error with IPV6 disabled:
[10-Jul-2025 03:22:48 UTC] check_port plugin: https://api64.ipify.org/ returned invalid IP: `xxx.xxx.xxx.xxx`.

We need to refrain from printing an error log when this happens. Otherwise, everything works as intended. Great job @cantalupo555!

cantalupo555:
> I get the following error with IPV6 disabled: [10-Jul-2025 03:22:48 UTC] check_port plugin: https://api64.ipify.org/ returned invalid IP: `xxx.xxx.xxx.xxx`.
> 
> We need to refrain from printing an error log when this happens. Otherwise, everything works as intended. Great job @cantalupo555!

The problem is in this line: `error_log("check_port plugin: {$url} returned invalid IP: " . $ip);`

I need to submit a commit/pull request to change the logic.
It would be helpful to submit this along with and reset margins. (https://github.com/Novik/ruTorrent/pull/2964#issuecomment-3055029145)
`a.separator.text("").hide().css({ "margin-left": "0", "margin-right": "0" });`

I won't have time to implement it today, but the new requests(fix) will be ready tomorrow.
Additionally, it would be great to see the error log when `curl` is not installed within the ruTorrent interface(check_port).

```
[10-Jul-2025 12:08:59 UTC] check_port plugin: Failed to fetch from https://api4.ipify.org/. Status: 127, Error: Error: cURL could not retrieve the document, error 127.
[10-Jul-2025 12:08:59 UTC] check_port plugin: Failed to fetch from https://api64.ipify.org/. Status: 127, Error: Error: cURL could not retrieve the document, error 127.
```

---

The plugin appears simple to the end user, but its overhaul was essential to modernize it and ensure compatibility with the future of IPv6.

stickz:
Awesome thanks. I'll wait for a commit be added to this pull request.

Ser4ph4:
Hello, taking advantage of the post, for about 2 days the port status icon has been unknown, the problem is that the website :https://ports.yougetsignal.com/check-port.php is offline , how to change this in plugin to use other site and get real port status, check on other site like : https://portchecker.co/check-v0
listed as open:
noname

stickz:
> Hello, taking advantage of the post, for about 2 days the port status icon has been unknown, the problem is that the website :https://ports.yougetsignal.com/check-port.php esta offline , how to change this in plugin to use other site and get real port status, check on other site like : https://portchecker.co/check-v0 listed as open: >
>

Hello, we might add a fallback in a separate pull request. I'm willing to hold off releasing `v5.3` until this is completed. It's not recommended to change the website. The other one goes offline more often. We want to use both websites instead.

cantalupo555:
Progress...
Now the borders follow the screen size...
image

cantalupo555:
New commit: https://github.com/Novik/ruTorrent/pull/2964/commits/24f8ebfaf0ad1b712bbc60ba5aae93ddb2b3b914

```
refactor(check_port): Improve component layout using modern CSS

Refactors the port status component to use a modern Flexbox layout instead of the previous inline-block and dynamic margin approach.

This change provides a more robust and maintainable method for handling vertical alignment and consistent spacing between elements using the `gap` property.

As a result, the corresponding JavaScript was simplified by removing the now-unnecessary logic for manipulating margins, leading to cleaner and more declarative code.
```

---

Another subtle change, during initial loading it now only displays 1 icon.
image
image

cantalupo555:
Yougetsignal is offline again...
Another provider I was targeting for a fallback was `port.tools`, which is also experiencing stability issues.
image

For the `v5.3` release, I believe `portchecker` should be the default until the fallback is implemented in the future.
In the meantime, using `ipify.org` as a fallback will be much smoother due to the wide variety of available APIs.

Ser4ph4:
> Yougetsignal is offline again... Another provider I was targeting for a fallback was `port.tools`, which is also experiencing stability issues. image
> 
> For the `v5.3` release, I believe `portchecker` should be the default until the fallback is implemented in the future. In the meantime, using `ipify.org` as a fallback will be much smoother due to the wide variety of available APIs.

Really OFFline again, is there any way to use another service, until they fix this in a future version?

cantalupo555:
> Really OFFline again, is there any way to use another service, until they fix this in a future version?

@Ser4ph4 
I'll submit modify to leave only `portchecker` as the default until a fallback is implemented in the future.
You just need to wait for version 5.3 to be ready.

cantalupo555:
New commit: https://github.com/Novik/ruTorrent/pull/2964/commits/ef72bd08e078a360387b47db217f7112b9d5a020

```
fix(check_port): Fix layout bug by rendering UI dynamically

Previously, if both IP versions were enabled in the configuration, the plugin would render HTML elements for both, even if the system only had connectivity for one. The unused elements were then hidden with CSS, which caused a layout bug where the Flexbox gap property created unwanted spacing.

This commit refactors init.js to build the UI completely dynamically on each update. The getPortStatus function now checks for the actual availability of each IP address at runtime before creating and appending its corresponding DOM elements (the .port-group).

This ensures that only the necessary HTML is rendered, resulting in a cleaner DOM and definitively fixing the layout spacing issue regardless of the user's network connectivity.
```

---

New commit: https://github.com/Novik/ruTorrent/pull/2964/commits/bcb97df146bb81b28fd96d547a8d4b5cd1fea1f2

```
fix(check_port): Improve IP detection and error logging

This commit enhances the check_port plugin to more intelligently handle different network configurations and to prevent unnecessary error logging.

Improved IP detection for single-stack environments:
The plugin now suppresses error logs when an IP check fails due to the system lacking either IPv4 or IPv6 connectivity.
cURL error 7 ("Failed to connect to host") is ignored, as it is an expected outcome in these scenarios.
The "invalid IP" error is also suppressed when a service like ipify returns an IP of a different version than requested (e.g., an IPv4 on an IPv6 query), which is common in NAT64/DNS64 networks.
```

---

New commit: https://github.com/Novik/ruTorrent/pull/2964/commits/3eae10102f7a82e14fb6b4cca6241c4ddd7fc53f

```
feat(check_port): Add fallback mechanism for IPv4 port check

The yougetsignal service, the default for IPv4, has become unstable and is frequently offline. This caused the port check functionality to fail for users with the default configuration.

This commit introduces a fallback mechanism to improve service reliability.

Fallback Implementation: If the primary service (yougetsignal) fails an IPv4 check (returning an "unknown" status), the plugin will automatically try the check again using the portchecker service as an alternative.

This is the first step towards a more robust implementation. Future improvements may include adding more service providers and implementing a similar fallback for the IP detection service (ipify.org).
```

cantalupo555:
@stickz Since I am working on a fork, it would be wise to avoid squashing commits.
The statistics on my GitHub profile are highly valued.

PR #2967, opened 2025-07-21
Fix: check_port with portchecker.co

dylansama:
Problem: check_port plugin fails checks when using the portchecker alternative site. Whatever their internal API was, changed paths for https://portchecker.co/

Easily checked by opening dev console and seeing the new path, with same body of `target_ip`, `port`, `selectPort`, `_csrf`. POSTing to the legacy path `/check-it` will get a 303 to the main page. Which just has default input form.


cantalupo555:
@dylansama This change has already been applied to the develop branch: https://github.com/Novik/ruTorrent/pull/2942

PR #2982, opened 2025-10-16
fix(_getdir): Remove unused info.php

allixx:
info.php seems to be a leftover of previous implementation and as far as I can see is not used anywhere.


Novik:
This file is not 'unused'. Chrome and firefox browser extenisons use it.
https://github.com/XirvikMadrid/xirvik_chrome_plugin
https://github.com/xirvik/seedbox_firefox_extension

allixx:
Alright.

Issue #1966, opened 2019-11-15
Request: Bulk Add (import), Bulk Copy (export) magnet links


a-raccoon:
The client qBittorrent allows you to paste hundreds of magnet links (or info-hashes or http links) in a single `Add Torrent(s)` operation.  It also allows you to select multiple loaded torrents to `Copy Magnet Link(s)` to clipboard for export to another client or friend.

I would really love to bulk export magnet links from qBittorrent then bulk import them into ruTorrent in one single click, instead of 4500 clicks.

GroupMaster:
Use the watch folder feature. Get the .torrent files you want imported into rutorrent and just copy them into the watch folder and let it rip. Maybe use the /load watch folder so you can make sure the torrents are set properly before they start.

a-raccoon:
@GroupMaster that would require SSH/FTP rights and privileges that joe user may not possess.  Please make this a working solution for joe users.

a-raccoon:
:'(

Novik:
> It also allows you to select multiple loaded torrents to Copy Magnet Link(s) to clipboard for export to another client or friend.

I added new plugin 'bulk_magnet'. 
In the current implementation it simply add for torrents context menu with a 3 commands
- Copy/Name
- Copy/Hash
- Copy/Magnet link.

But I have no idea - will this works for hundreds of torrents or not. 
You may check this, and if this will be useful then I will think about bulk addition.

Novik:
I added torrents import too. You may find corresponded menu item in the 'Plugins' submenu.

PR #2994, opened 2025-12-21
magnet bulk add

### LINKED ISSUE #2025: [Feature Request] Add/copy multiple magnet links and multiple peers.
I just switched to ruTorrent after being a [Transmission Web Control](https://github.com/ronggang/transmission-web-control) user for years and this is the feature that I miss most.
The current dialogues to add the links and peers only accept single line values while multi line inputs are reduced to single liners and in case of magnets only the first link is honored. Also there is no context menu option to copy magnet links like transmission-web-control and most major clients. It would be really helpful if these could be changed to multi-line dialogues, with use cases like mine where I sometimes need to manually add peers from trackers like RARBG that blacklist for example my seedbox's IP address preventing me from getting peers, and for users (including me) that mass import magnet links.

I'm aware that this issue is a (partial) duplicate of #1966 but I felt I needed to open another because this also includes the peer import dialogue.

edde746:
title, actually resolves #2025 and #1966

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -692,6 +692,14 @@ span#loadimg {padding: 20px; background: transparent url(../images/ajax-loader.g
 
 /* Custom break point settings for responsiveness */
 
+/* Multiline URL textarea in Add Torrent dialog */
+#addtorrenturl textarea#url {
+	height: 80px;
+	resize: vertical;
+	font-family: monospace;
+	font-size: 0.9em;
+}
+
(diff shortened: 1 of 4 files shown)


xirvik:
@edde746 This removal of the API 11 port aliases from content.js is unrelated to the rest of the PR, did that change sneak in? Could you rebase and remove unrelated changes please?

Issue #2977, opened 2025-09-29
parse errors xmlrpc - Bad response from server: (200 [parsererror,list])


spicemint:
### Please complete the following tasks.

- [x] Web browser cache cleared
- [x] Link provided to install script if applicable
- [x] Not using broken rtinst install script
- [x] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Web Browser: Firefox (142.0.1 (aarch64)), Edge (Version 140.0.3485.81 (Official build) (arm64)), Safari (Version 26.0.1 (20622.1.22.118.4)), Chrome (Version 140.0.7339.210 (Official Build) (arm64)
), Vivaldi Version 140.0.7339.210 (Official Build) (arm64)

rutorrent: 5.2.10
PHP: 8.4.12
OS: Arch 6.16.8-arch1-1
rtorrent/libtorrent: 16

rutorrent resides in /usr/share/webapps/rutorrent/
the conf/config.php
`	<?php
	// configuration parameters
	// for snoopy client
	$httpUserAgent = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36';
	$httpTimeOut = 30;			// in seconds
	$httpUseGzip = true;
	$httpIP = null;				// IP string. Or null for any.
	$httpProxy = array
	(
		'use'	=> false,
		'proto'	=> 'http',		// 'http' or 'https'
		'host'	=> 'PROXY_HOST_HERE',
		'port'	=> 3128
	);

	// for xmlrpc actions
	$rpcTimeOut = 10;			// in seconds
	$rpcLogCalls = false;
	$rpcLogFaults = true;

	// for php
	$phpUseGzip = false;
	$phpGzipLevel = 2;

	$schedule_rand = 10;			// rand for schedulers start, +0..X seconds

	$do_diagnostic = false;			// Diagnose ruTorrent. Recommended to keep enabled, unless otherwise required.
	$al_diagnostic = true;			// Diagnose auto-loader. Set to "false" to make composer plugins work.

	$log_file = $_ENV['RU_LOG_FILE'] ?? '/tmp/errors.log'; // path to log file (comment or leave blank to disable logging)

	$saveUploadedTorrents = true;		// Save uploaded torrents to profile/torrents directory or not
[...]

PR #2996, opened 2026-01-12
Update JSON::safeEncode to sanitize, rather than attempting convert/repair

anthonyryan1:
Right now, we've got a lot of code attempting to correct every malformed UTF8 string into something usable. This pull request is more of an alternate approach where we don't try and coax every string into something valid, but instead replace invalid sequences with a broken character encoding character.

Also ensures that on JSON encode failures we don't fail silently and return a HTTP 200.

One of two possible fixes for #2977

The other approcah is collecting every broken UTF8 name and adding unit tests for the UTF8 repair class. I have saved some strings that I've observed causing these problems in the wild, and can share them with someone motivated to write tests and fix the UTF8 fixer instead.

### DIFF
--- a/php/utility/json.php
+++ b/php/utility/json.php
@@ -3,10 +3,9 @@
 require_once('utf.php');
 
 class JSON
-{	
+{
 	public static function safeEncode($value)
 	{
-		$encoded = json_encode($value);
-		return(!function_exists('json_last_error') || json_last_error()==JSON_ERROR_NONE ? $encoded : json_encode(UTF::utf8ize($value)));
+		return json_encode($value, JSON_INVALID_UTF8_SUBSTITUTE);
 	}
 }
--- a/php/utility/utf.php
+++ b/php/utility/utf.php
@@ -107,61 +107,6 @@ class UTF
 		return($outstr);
 	}
 
-	private static function mix2utf($str, $inv = '_') 
-	{
-		$len = strlen($str);
-		for($i = 0; $i < $len; $i++)
-		{
-			$c = ord($str[$i]);
-			if($c > 128) 
-			{
-				$bytes = 0;
-				if(($c > 247)) $str[$i] = $inv;
-				elseif($c > 239) $bytes = 4;
-				elseif($c > 223) $bytes = 3;
-				elseif($c > 191) $bytes = 2;
-				else $str[$i] = $inv;
-				if($bytes)
-				{
-					if(($i + $bytes) > $len) $str[$i] = $inv;
-					else
-					{
-						$start = $i;
-						$cnt = $bytes;
-						while($bytes > 1) 
-						{
-							$i++;
-							$b = ord($str[$i]);
-							if($b < 128 || $b > 191) 
-							{
-								$str[$start] = $inv;
-								$i = $start;
-								break;
-							}
-							$bytes--;
-						}
-					}
-				}
-			}
-		}
-		return($str);
-	}
-
-	public static function utf8ize($mixed) 
-	{
-		if(is_array($mixed) || is_object($mixed)) 
-		{
-				foreach($mixed as $key => $value) 
-				{
-					$mixed[$key] = self::utf8ize($value);
-				}
-			} 
-			else 
-				if(is_string($mixed)) 
-					$mixed = self::mix2utf($mixed);
-		return($mixed);
-	}
-
 	private static function maybe_mb_convert_to_utf8($string) {
 		return function_exists('mb_convert_encoding') ? mb_convert_encoding($string, "UTF-8", "auto") : $string;
 	}
--- a/plugins/httprpc/action.php
+++ b/plugins/httprpc/action.php
@@ -496,4 +496,4 @@ if(is_null($result))
 	CachedEcho::send( (isset($req) && $req->fault) ? "Warning: XMLRPC call is failed." : "Link to XMLRPC failed. Maybe, rTorrent is down?","text/html");
 }
 else
-	CachedEcho::send(JSON::safeEncode($result),"application/json");
+	CachedEcho::send(json_encode($result, JSON_THROW_ON_ERROR|JSON_INVALID_UTF8_SUBSTITUTE), "application/json");


anthonyryan1:
Just to clarify, this problem affects *every* endpoint that uses safeEncode, not just the list in this example.

ranirahn:
If all calls to safeEncode are affected why not fix the main function in
Rutorrent/php/utility/json.php ?

anthonyryan1:
That is the long term plan, yes. This isn't ready to merge, it's a discussion starting point for how we want to try solving these UTF8 normalization failures.

ranirahn:
What do you think if the json class is like this?

```php
class JSON
{
    public static function safeEncode($value)
    {
        if (defined('JSON_THROW_ON_ERROR')) {
           try {
               return json_encode($value, JSON_THROW_ON_ERROR | JSON_INVALID_UTF8_SUBSTITUTE);
           } catch (JsonException $e) {
               throw $e;
           }
        }
        $encoded = json_encode($value);
        if ($encoded === false) {
            require_once('utf.php');
            return json_encode(UTF::utf8ize($value));
        }
        return $encoded;
    }
}
``` 

I would leave the old version just incase someone have PHP 5 still in use and then these substitude and stuff dont work. So far dont see anything break with this.

anthonyryan1:
There hasn't been much discussion on the trade-offs between attempting to repair vs sanitizing the corrupt strings.

So I've just made a call, and amended this PR to my personal preference for how to handle this issue.

------

UTF::utf8ize previously attempted to "repair" as many mangled strings
as possible, but I've observed a number of examples over the years where
it choked and broke rutorrent.

I would argue that taking arbitrary strings of bytes, and attempting
to convert them with perfect accuracy into valid UTF8 (when those strings
come to us mangled, corrupt and always with unknown encodings) isn't possible.

Here's a handful of examples of of byte sequences that choke the UTF8 repair code:

```php
require('php/utility/json.php');
var_dump(JSON::safeEncode("\xC0\x80"));
var_dump(JSON::safeEncode("\xED\xA0\x80"));
var_dump(JSON::safeEncode("\xED\xBF\xBF"));
var_dump(JSON::safeEncode("\xF5\x80\x80\x80"));
var_dump(JSON::safeEncode("\xF7\xBF\xBF\xBF"));
```

The problem is that I think this list is nowhere near comprehensive, and
even if we fix all of these, there will be many more corrupt sequences in
the wild we can't predict.

Instead of trying to repair all this corrupt data, let's just use the xFFFD
replacement character. Users who add files with corrupt strings will see the
replacement character (adding some awareness), but that becomes a content
problem and not a "ruTorrent is broken" problem.

ranirahn:
I think it breaks everyones rutorrent that still runs on PHP 5. Also why delete code in utf8 file if its not in use anymore? What I propose was that it try-s without utf8 file and if keys exist that means its newer PHP and uses that and if not then it has fall back on old system. Just to cut it off seems little too big change. Cant predict how many people coming then that rutorrent dont work at all. Maybe someone with more knowledge can tell what will happen with all the people who still use PHP5 and older. I have no idea, but i am sure its not nothing.

anthonyryan1:
You are correct that this won't work with PHP 5, it'll only work with PHP 7.2 or higher (when JSON_INVALID_UTF8_SUBSTITUTE was added).

PHP 7.2 was released in Nov 2017. The last version of PHP that wouldn't work with this change is 7.1 which hit it's security end of life on Dec 2019.

I think it's probably safe to say that anyone who hasn't installed updates in ~9 years and is running very unsafe PHP versions, probably also isn't updating ruTorrent. But if @xirvik's thinks we should preserve PHP 5 compatibility, I can rework this into feature detection.

xirvik:
> I think it breaks everyones rutorrent that still runs on PHP 5. Also why delete code in utf8 file if its not in use 

People running PHP 5 need to update. There's no way we're going to keep supporting a PHP version that the PHP maintainers themselves declared end-of-life many years ago.

If you really really can't update PHP,  then you have to stay on whatever version of ruTorrent still happens to work, which will also support an old version of rtorrent, etc.

ranirahn:
Probably just need to add a warning somewhere that if you run PHP7.1 or older rutorrent does not run. Same with if its older rtorrent. Then i think its fine to clean this up. Then rutorrent tells you what is wrong and how to fix it. Right now its just not work and there is no info about it on rutorrent itself. Need to come and find that on internet. in content.js there is check for rtorrent version. If there would be also PHP version check then that would work. Example this

```
if(theWebUI.systemInfo.rTorrent.iVersion < 0x907) {
		const title = theUILang.requiresAtLeastRtorrent.replace('{version}', 'v0.9.7');
		$($$('webui.show_open_status')).attr({ disabled: '', title });
		$($$('lbl_webui.show_open_status')).attr({ title }).addClass('disabled');
	}
``` 
rutorrent should be also disabled if it does not have new enough PHP. Not sure how to do that. Same problem was with one plugin. File was removed but plugin still required it. This time its not plugin, but dont know where this file was required more. May not be only place where this utf8 was required.

xirvik:
@anthonyryan1 
- Agreed on the core direction: sanitize, don't repair.
-  We've set PHP 7.4 as the documented minimum, and the new env_check.php reports an unsupported PHP version. So no need to check for old versions or workaround them.
- There's one change I'd ask for before merge: drop JSON_THROW_ON_ERROR, keep just JSON_INVALID_UTF8_SUBSTITUTE, and log-and-degrade on the residual error.  The substitute flag alone already fixes the reported bug, and it handles every one of your five choke examples without throwing (at least on PHP 8.4 which is what I have handy - correct me if I'm wrong on other versions).
-

anthonyryan1:
Amended.

JSON_THROW_ON_ERROR is just a personal habit for PHP. I always felt like PHP did a poor job with the default function return and JSON_THROW_ON_ERROR mostly fixes it.

I never liked that json_decode returns null on failures for example, when null is valid JSON that can also be decoded.

```
php > var_dump(json_decode('{"name": "John Doe}'));
NULL
php > var_dump(json_decode("null"));
NULL
php > var_dump(json_decode('{"name": "John}', flags: JSON_THROW_ON_ERROR));
PHP Warning:  Uncaught JsonException: Control character error, possibly incorrectly encoded in php shell code:1
Stack trace:
#0 php shell code(1): json_decode()
#1 {main}
  thrown in php shell code on line 1
```

It's not especially applicable here, because json_encode is a little better (string or false), and I've mainly used it there to give PHPStan a simpler function signature for reasoning about type hints (not applicable in this code base).

PR #2998, opened 2026-01-23
Normalize IPv4-in-IPv6 notation

anthonyryan1:
rutorrent 0.16.x returns some IPv4 peers using IPv4-in-IPv6 notation which leads to confusion and breaks things like the geoip plugin.

Normalize these IPs to their canonical IPv4 format, and fix tests.

I've only observed it returning IPv6's with surrounding square brackets but I added a test for the other notation to be robust against future changes.


anthonyryan1:
Closing this as irrelevant, one point release in rtorrent had this problem and all new ones don't.

PR #3009, opened 2026-04-09
fix(httprpc): use trusted SCGI for server-side handlers, add d.resume for 0.16.8

xirvik:
## Summary

rtorrent's XMLRPC untrusted connection security model (merged in rakshasa/rtorrent#1710) blocks commands sent with `UNTRUSTED_CONNECTION=1`. The httprpc plugin marks **all** connections as untrusted via `success(false)`, which breaks torrent management (start, stop, labels, priorities, settings, etc.) on rtorrent builds with this feature enabled.

### Changes

- **Dedicated mode handlers** (`start`, `stop`, `remove`, `setprops`, `setsettings`, etc.): change `success(false)` → `success(true)`. These handlers build XMLRPC commands server-side from validated `mode`/`hash` parameters — they are the trusted intermediary between the browser and rtorrent.
- **Default raw handler** (`send($HTTP_RAW_POST_DATA, false)`): remains **untrusted**. This path forwards arbitrary XMLRPC from the browser and is exactly what the security model should restrict.
- **Start handler**: add `d.resume` alongside `d.open`/`d.start`. In rtorrent 0.16.8, `d.start` only sets the view state; `d.resume` is needed to actually activate the download.
- **Unpause handler**: use `d.resume` instead of `d.start` (correct command for resuming a paused torrent).

### Background

With the untrusted connection patches active, rtorrent checks each XMLRPC command — including internal sub-commands — against a whitelist. Commands like `d.start`, `d.stop`, `d.custom1.set`, and their internal sub-commands are not all whitelisted — and they don't need to be, because the PHP layer already authenticates the user and constructs the commands. The trust boundary is at the PHP/HTTP-auth level, not at the SCGI level.

## Test plan

Live in production seedboxes for the past weeks.

### DIFF
--- a/plugins/httprpc/action.php
+++ b/plugins/httprpc/action.php
@@ -62,7 +62,7 @@ function makeMulticall($cmds,$hash,$add,$prefix)
 		$cmd->addParameter($prm);
 	$cnt = count($cmds)+count($add);
 	$req = new rXMLRPCRequest($cmd);
-	if($req->success(false))
+	if($req->success(true))
 	{
 	        $result = array();
 		for($i = 0; $i<count($req->val); $i+=$cnt)
@@ -78,7 +78,7 @@ function makeSimpleCall($cmds,$hash)
 	foreach($hash as $h)
 		foreach($cmds as $cmd)	
 			$req->addCommand( new rXMLRPCCommand( $cmd, $h ) );
-       	return($req->success(false) ? $req->val : false);
+       	return($req->success(true) ? $req->val : false);
 }
 
 $result = null;
@@ -102,7 +102,7 @@ switch($mode)
 			$cmd->addParameter($prm);
 		$cnt = count($cmds)+count($add);
 		$req = new rXMLRPCRequest($cmd);
-		if($req->success(false))
+		if($req->success(true))
 		{
 			$theCache = new rpcCache();
 			$dTorrents = array();
@@ -171,7 +171,7 @@ switch($mode)
 			$req->addCommand( new rXMLRPCCommand( $cmd ) );	
 		foreach( $add as $prm )
 			$req->addCommand( new rXMLRPCCommand( $prm ) );	
-		if($req->success(false))
+		if($req->success(true))
 		{
 	        	$result = array();
 			$dht_active = $req->val[0];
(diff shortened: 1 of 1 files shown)


PPN-SD:
Thanks @Novik for merging. Would it be possible to have a new release to ensure compatibility with rtorrent 0.16.9 more easily?

crazyzhc98:
It doesn't work on rtorrent `master` brunch.

xirvik:
> It doesn't work on rtorrent `master` brunch.

This was merged on develop. If you want to use master, cherry-pick.

ranirahn:
There have not been a new release since July last year and I don't expect we seeing one for a long time. So anyone who wants to use the fixes need to add these to their copy themselves. Master and develop versions both are behind and both have some broken parts for last rtorrent version. Until every fix gets merged, there is no point doing the release and there is no one who has time and can do the release, so we will be waiting for an official release for a long time I think.

crazyzhc98:
> > It doesn't work on rtorrent `master` brunch.
> 
> This was merged on develop. If you want to use master, cherry-pick.

What I mean is that even using the current `master` branch of rtorrent together with the `develop` branch of ruTorrent, I still can't control the torrent status.

The issue still persists.

xirvik:
> What I mean is that even using the current `master` branch of rtorrent together with the `develop` branch of ruTorrent, I still can't control the torrent status.

Don't know how stable the develop branch of ruTorrent is. You have some options

- Use 5.2.10 + this patch (cherry pick).
- Disable the header in ruTorrent if you want to use bleeding edge
- Use rtorrent 0.16.8 until the issue is clear

You can also open an issue here with details of what's not working so we try to repro.

JasonKhew96:
@xirvik commands in `js/rtorrent.js` are not handled with trusted SCGI, `update trackers` in context menu will get an error `Command "d.tracker_announce" is not allowed for untrusted connections.`

EDIT: it's untrusted by default in the code
https://github.com/Novik/ruTorrent/blob/87315d6fad35dc616503caaf52c06953bcfaab6a/plugins/httprpc/action.php#L494


> Default raw handler (send($HTTP_RAW_POST_DATA, false)): remains untrusted. This path forwards arbitrary XMLRPC from the browser and is exactly what the security model should restrict.

we might need a fix/refactor for some of the restricted commands

xirvik:
> @xirvik commands in `js/rtorrent.js` are not handled with trusted SCGI, `update trackers` in context menu will get an error `Command "d.tracker_announce" is not allowed for untrusted connections.`
> 
Thanks - that's a bug. Fixed in develop branch, commit ec32347.

PR #3010, opened 2026-04-11
Update README Development version download link

henning-schild:


henning-schild:
It seems that "master" is recently used for stable and "develop" is the ref to follow for the most recent dev version.

If merged this patch should also be backported to "master" so people clicking the link in the default github view actually get the newest code.

PR #3011, opened 2026-04-13
Fix PHP 8.2+ deprecation warnings

xirvik:
## Summary

  - **cache.php**: Replace dynamic `$modified` property injection with a static `$modifiedTimes` array and suppress `filemtime()` warnings on non-existent cache files
  - **Torrent.php**: Add `#[AllowDynamicProperties]` attribute (required by PHP 8.2+)
  - **rss/rss.php**: Declare `$isValid` property to avoid dynamic property deprecation
  - **trafic/stat.php + update.php**: Pass explicit `$escape` parameter to `fgetcsv()`/`fputcsv()` (the default backslash escape was deprecated in PHP 8.4)

 ## Context

 These warnings are noisy on PHP 8.2+ and will become errors in PHP 9. The cache.php change is the most significant — it replaces per-object `$modified` property injection with a class-level static array keyed by class+hash, which is cleaner and avoids the dynamic property entirely while maintaining backward compatibility with plugins that declare their own `$modified`.

 ## Test plan

Tested in a live seedbox, no errors.

### DIFF
--- a/php/Torrent.php
+++ b/php/Torrent.php
@@ -11,6 +11,7 @@
 
 require_once( 'util.php' );
 
+#[\AllowDynamicProperties]
 class Torrent
 {
 	protected $errors = array();
--- a/php/cache.php
+++ b/php/cache.php
@@ -115,7 +115,7 @@ class rCache
 	}
 	public function getModified( $obj = null )
 	{
-		return(filemtime( is_null($obj) ? $this->dir : 
+		return(@filemtime( is_null($obj) ? $this->dir : 
 			(is_object($obj) ? $this->getName($obj) : $this->dir."/".$obj) ));
 			
 	}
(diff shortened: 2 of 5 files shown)


usr20210909:
The changes to **trafic/stat.php + update.php** are causing many warnings in the log when using PHP 7.0:

`fgetcsv(): escape must be character in /REDACTED/rutorrent/plugins/trafic/stat.php on line 25`
`fgetcsv(): escape must be character in /REDACTED/rutorrent/plugins/trafic/stat.php on line 26`
`fgetcsv(): escape must be character in /REDACTED/rutorrent/plugins/trafic/stat.php on line 27`
`fgetcsv(): escape must be character in /REDACTED/rutorrent/plugins/trafic/stat.php on line 28`
`fgetcsv(): escape must be character in /REDACTED/rutorrent/plugins/trafic/stat.php on line 29`
...

PR #3013, opened 2026-04-13
Rewrite geoip plugin for GeoIP2/MaxMindDB

xirvik:
The old GeoIP1 PHP extension (`geoip_country_code_by_name()`) was removed from PECL and is unavailable on PHP 8.x. The GeoIP2 composer library works with MaxMindDB files (`.mmdb`) which are the current standard from MaxMind. The lookup now uses `\GeoIp2\Database\Reader` which is already available via composer in most ruTorrent installations.

### DIFF
--- a/plugins/geoip/conf.php
+++ b/plugins/geoip/conf.php
@@ -9,3 +9,8 @@
 
 	$dnsResolver = '1.1.1.1';	// use gethostbyaddr, if null
 	$dnsResolverTimeout = 1;	// timeout in seconds
+
+	// GeoIP2 MaxMindDB configuration
+	$geoip2Autoloader = '/usr/local/share/GeoIP/vendor/autoload.php';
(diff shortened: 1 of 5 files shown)


usr20210909:
What should seedbox users do, who are limited to PHP 7.0 and can't install GeoIP2/MaxMindDB - stay on ruTorrent v5.2.10?

xirvik:
> What should seedbox users do, who are limited to PHP 7.0 and can't install GeoIP2/MaxMindDB - stay on ruTorrent v5.2.10?

Honestly - upgrade to a supported PHP version or stay with a ruTorrent version (and many other things) that works on PHP 7. PHP 7.0 has been EOF for over 7 years now.

If you can't because of your seedbox provider, consider switching to one that doesn't force you to be on hopelessly obsolete software.

PR #3018, opened 2026-04-13
Fix check_port: gracefully handle missing IPv6 instead of showing unknown status

xirvik:
On servers without IPv6, the check_port plugin showed an "unknown" status icon for IPv6, which looked like an error. Now it cleanly hides the IPv6 indicator and shows "Not available on this server" in the tooltip.

### DIFF
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -160,7 +160,7 @@ function get_and_check_ip($ip_version, $use_website, $rtorrent_ip, $rtorrent_por
 		return ["ip" => $ip_to_check, "status" => $status];
 	}
 	// Return a default "not available" state if no IP could be determined
-	return ["ip" => "-", "status" => 0];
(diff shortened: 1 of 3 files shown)


usr20210909:
IPv6 indicator gets hidden but an error is logged every time ruTorrent is opened in browser:

`2026/04/19 20:58:00 [error] 9193#9193: *515747 FastCGI sent in stderr: "PHP message: check_port plugin: https://api64.ipify.org/ returned invalid IP: RE.DA.CT.ED" while reading response header from upstream, client: RE.DA.CT.ED, server: REDACTED.com, request: "POST /rutorrent/plugins/check_port/action.php?init HTTP/1.1", upstream: "fastcgi://unix:/REDACTED/.nginx/php/socket:", host: "REDACTED.com", referrer: "https://REDACTED.com/REDACTED/rutorrent/"`

PR #3027, opened 2026-04-20
[check_port] Properly hide IPv6 icon if the protocol is disabled

TheGoblinHero:
I have in conf.php:
`$useWebsiteIPv4 = "yougetsignal"; 
$useWebsiteIPv6 = false;`

But the IPv6 icon is still shown with the status text "Unknown status" instead of "Not availabale". Here's my fix to solve this problem.


TheGoblinHero:
And second small fix here. [erasedata] now requires httprpc. We should add it to its dependencies.

xirvik:
erasedata doesn't depend on httprpc - it needs that "something" provides  getCommon("removewithdata") . It might be httprpc but it could be a different plugin. I know that vanilla ruTorrent doesn't have other plugin (at the moment) that provides that, but they do exist. 

Let's remove that dependency please.

Also, can you target develop and not master?

TheGoblinHero:
Rebased to develop. About deps. In the current state [erasedata] needs exactly httprpc:
https://github.com/Novik/ruTorrent/commit/e08b97c42d66775246de83b5c750ace2e795f442
and won't work without it. I've spend 15 minutes finding why after the update [erasedata] does not work anymore. Dependency warning is a good thing for users like me, who use only needed plugins. But if you still don't like deps change, I'm ok with it.

xirvik:
> Rebased to develop. About deps. In the current state [erasedata] needs exactly httprpc: [e08b97c](https://github.com/Novik/ruTorrent/commit/e08b97c42d66775246de83b5c750ace2e795f442) and won't work without it. I've spend 15 minutes finding why after the update [erasedata] does not work anymore. Dependency warning is a good thing for users like me, who use only needed plugins. But if you still don't like deps change, I'm ok with it.

It's not needed if you have another plugin that overrides that method; but if it didn't have that dependency before and now it does (not that plugin, but some plugin), that's a bad change.

I'm curious though, what do you use instead of httprpc?

TheGoblinHero:
Reverted. 
I'm using nginx SCGI interface for years.

koblack:
> Reverted. I'm using nginx SCGI interface for years.

I have similar setup. I always disable httprpc plugin. Is the httprpc plugin now mandatory?

TheGoblinHero:
[erasedata] now won't work without it. May be something else.

xirvik:
> > Reverted. I'm using nginx SCGI interface for years.
> 
> I have similar setup. I always disable httprpc plugin. Is the httprpc plugin now mandatory?

Apparently yes until fixed. But I'm looking into it now, I don't like these regressions.

PR #3030, opened 2026-04-21
Fixes and dialog window to view and clear full log in log_history plugin.

ranirahn:
Fixing the clear button in log tab. Also made new dialog window to see all saved logs in the .dat file. Can open it from Plugins->Log History. I am not coder, so maybe someone can fix my mess before merging it.

### DIFF
--- a/plugins/log_history/init.js
+++ b/plugins/log_history/init.js
@@ -143,7 +143,6 @@ theWebUI.showLogHistory = function () {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


xirvik:
@ranirahn I fixed the bug (the clear button). 
You will need to rebase your PR and remove that part.
Also, please try to keep separate things in separate PR - not have one PR with bugfixes and new features, it makes testing a lot harder, and it's possible we just want to merge one thing but not the other, etc.

PR #3044, opened 2026-04-23
Refactor logging functionality in init.js

ranirahn:
Changing noty() to log(). All messages go trough log function, but only some use noty function, so better to use log function then. That means status info is missing. Also using noty break the popup noty messages

### DIFF
--- a/plugins/log_history/init.js
+++ b/plugins/log_history/init.js
@@ -84,13 +84,13 @@ function sendLogToServer(msg, status) {
 plugin._replaying = false;
 
 plugin.init = function () {
-    plugin._originalNoty = window.noty;
-    const originalNoty = window.noty;
+    plugin._originalLog = window.log;
+    const originalLog = window.log;
 
-    window.noty = function(msg, status, noTime) {
-        originalNoty(msg, status, noTime);
+    window.log = function(text, noTime, divClass, force) {
+        originalLog(text, noTime, divClass, force);
         if (!plugin._replaying) {
-            sendLogToServer(msg, status || 'info');
+            sendLogToServer(text, status || 'info');
         }
     };
 


plugins/log_history/init.js
```diff
@@ -84,13 +84,13 @@ function sendLogToServer(msg, status) {
...
+    plugin._originalLog = window.log;
+    const originalLog = window.log;
 
-    window.noty = function(msg, status, noTime) {
-        originalNoty(msg, status, noTime);
+    window.log = function(text, noTime, divClass, force) {
+        originalLog(text, noTime, divClass, force);
         if (!plugin._replaying) {
-            sendLogToServer(msg, status || 'info');
+            sendLogToServer(text, status || 'info');
```
xirvik, line 93: status is now undefined, isn't it?

xirvik (changes requested):
I don't think we want this TBH. We're using noty() because it captures user-visible messages (errors, notifications), while  log() captures everything including internal debug messages - that can make history very noisy.

ranirahn:
Well, the point was to log this log tab. If its logging only the popup window then its different thing. But I dont care. I use the log function to see everything that was in the log tab. If plugin shows only half the log tab then i dont know what is the point of this plugin at all. Status with the log function is undefined, but my original plugin show every message in popup window, but that also wanted to rewrite same message again on every reload. Now messages dont get written to log multiple times, but also the popup window does not exist. So I dont know what is the best to do, but broken popup noty window is 100% bad and thats why i would use log function and leave noty function alone. noty always calls log function, but log never calls noty.
Why i would leave the status in there is maybe should rewrite the log function itself that uses the status from noty when it calls the log funktion it also gives the status to log function, but I dont know.

PR #3047, opened 2026-04-24
Whitelist load.* methods as trusted in httprpc raw XMLRPC pass-through

### LINKED ISSUE #3046: httprpc: incomplete rTorrent 0.16.9 trust fix — `default` case still blocks `load.start`

- [x] Web browser cache cleared
- [x] Link provided to install script if applicable
- [x] Not using broken rtinst install script
- [x] Web browser, ruTorrent, PHP and OS version provided

https://github.com/crazy-max/docker-rtorrent-rutorrent
Browser not relevant

https://github.com/crazy-max/docker-rtorrent-rutorrent

## Description
The v5.3.0 fix for rTorrent 0.16.9 compatibility ([commit 87315d6](https://github.com/Novik/ruTorrent/commit/87315d6fad35dc616503caaf52c06953bcfaab6a)) is incomplete.

External clients that use `plugins/httprpc/action.php` as their RPC endpoint (Prowlarr, Sonarr,
Radarr, Transdrone, NZB360, etc.) still receive the following error when trying to add torrents:

```
rTorrent returned error code 0: Command "load.start" is not allowed for untrusted connections
```

### Steps to reproduce

1. configure and run https://github.com/crazy-max/docker-rtorrent-rutorrent
2. configure and run prowlarr 
3. configure rutorrent as a client inside prowlarr through httprcp
4. try to add any torrent from prowlarr

### Expected behavior

Torrent is added correctly

## Background: rTorrent 0.16.9 trust model

rTorrent 0.16.9 introduced a trusted/untrusted XMLRPC connection security model based on the
`UNTRUSTED_CONNECTION` SCGI header flag. When set to `1`, rTorrent restricts the connection to
a whitelist of allowed commands, blocking anything not explicitly listed — including `load.start`,
`load.raw.start`, and others commonly used by external clients.

The flag is set in `php/xmlrpc.php` (line ~109):

```php
$reqheader = "CONTENT_LENGTH\x0".$contentlength."\x0"."CONTENT_TYPE\x0"."text/xml\x0"
           . "SCGI\x0"."1\x0UNTRUSTED_CONNECTION\x0".($trusted ? "0" : "1")."\x0";
```

The `$trusted` parameter propagates from `rXMLRPCRequest::send($data, $trusted)`,
`rXMLRPCRequest::run($trusted)`, and `rXMLRPCRequest::success($trusted)`.

## What the v5.3.0 fix addressed

The fix correctly changed all `->success(false)` calls to `->success(true)` throughout
`action.php`, covering all the named modes (`list`, `start`, `stop`, `pause`, `remove`, etc.)
used by the ruTorrent UI itself.

## What was missed

There is one remaining call that still passes `false` — the `default` case, which handles
**raw XMLRPC pass-through**:

```php
default:
{
    if(isset($HTTP_RAW_POST_DATA))
    {
        $result = rXMLRPCRequest::send($HTTP_RAW_POST_DATA, false); // ← still untrusted
```

This is exactly the code path used by external clients like Prowlarr, Sonarr, and Radarr.
These clients do not use the named modes — they send raw XMLRPC payloads directly, which are
handled exclusively by this `default` case. As a result, commands like `load.start` are still
rejected by rTorrent 0.16.9 even after upgrading to ruTorrent v5.3.0.

## How I found this

The error first appeared when Prowlarr (configured to use `plugins/httprpc/action.php` as the
RPC endpoint) stopped being able to add torrents after upgrading to rTorrent 0.16.9.

Tracing the request path:
- Prowlarr sends a raw XMLRPC payload to `plugins/httprpc/action.php`
- Since the payload does not match any named `$mode`, it falls through to the `default` case
- `rXMLRPCRequest::send($HTTP_RAW_POST_DATA, false)` is called, setting `UNTRUSTED_CONNECTION: 1`
- rTorrent 0.16.9 rejects `load.start` as it is not on the untrusted whitelist

The named modes (`start`, `stop`, etc.) work correctly after the v5.3.0 fix because they all
use `->success(true)`. Only the raw pass-through path remains broken.

## Why marking this as trusted is correct and safe

The `default` case is only reachable by clients that have already passed through nginx
authentication (HTTP basic auth or similar) configured by the image. The trusted/untrusted
distinction in rTorrent 0.16.9 is designed to separate unauthenticated external network
connections from trusted local ones — not to filter commands within an already-authenticated
PHP proxy. Marking this connection as untrusted is therefore inconsistent with how all other
modes in the same file are treated.

## Fix

One-line change in `plugins/httprpc/action.php`:

```diff
- $result = rXMLRPCRequest::send($HTTP_RAW_POST_DATA, false);
+ $result = rXMLRPCRequest::send($HTTP_RAW_POST_DATA, true);
```

## Affected clients

Any client configured to use `/plugins/httprpc/action.php` as the RPC endpoint and sending
raw XMLRPC payloads, including:

- Prowlarr
- Sonarr (with rTorrent integration)
- Radarr (with rTorrent integration)
- Transdrone / Transdroid
- NZB360
- Any other *arr or automation tool using raw XMLRPC

## Workaround (until fixed)

Add the following to a custom init script (mapped into `/etc/cont-init.d/` for
crazy-max/docker-rtorrent-rutorrent, or equivalent for other setups):

```sh
ACTION_PHP="/var/www/rutorrent/plugins/httprpc/action.php"

if grep -qF 'success(false)' "$ACTION_PHP" || grep -qF 'send($HTTP_RAW_POST_DATA,false)' "$ACTION_PHP"; then
    sed -i 's/success(false)/success(true)/g' "$ACTION_PHP"
    sed -i 's/send(\$HTTP_RAW_POST_DATA,false)/send(\$HTTP_RAW_POST_DATA,true)/g' "$ACTION_PHP"
    echo "httprpc action.php patched"
else
    echo "httprpc action.php already patched, skipping"
fi
```

## References

- rTorrent 0.16.9 release: https://github.com/rakshasa/rtorrent/releases/tag/v0.16.9
- ruTorrent v5.3.0 partial fix: https://github.com/Novik/ruTorrent/commit/87315d6fad35dc616503caaf52c06953bcfaab6a

xirvik:
External clients (Prowlarr, Sonarr, Radarr, Transdrone, NZB360, etc.) send raw XMLRPC to `plugins/httprpc/action.php`, which falls through to the `default` case. This case sent all raw XMLRPC as untrusted, blocking `load.start` on rtorrent 0.16.9+ with:

 Command "load.start" is not allowed for untrusted connections

 Instead of making the entire default case trusted (which would allow dangerous commands like `execute`), parse the XMLRPC method name using `simplexml_load_string` and whitelist only load-related methods:

  - `load.start`, `load.raw_start`, `load.raw`, `load.normal`
  - Legacy equivalents: `load_start`, `load_raw_start`, `load_raw`

All other methods remain untrusted. Untrusted raw XMLRPC method names are logged via `FileUtil::toLog()` to help diagnose missing whitelist entries.

Includes 10 unit tests covering trusted methods, untrusted methods (including `system.multicall` with nested `execute`), invalid XML, and parameter injection attempts.

Fixes #3046

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,27 +21,27 @@
 		<meta name="msapplication-TileColor" content="#FFFFFF" />
 		<meta name="msapplication-TileImage" content="./images/mstile-144x144.png" />
 
-		<link href="./css/bootstrap.min.css?v=530" rel="stylesheet" type="text/css" />
-		<link href="./css/style.css?v=530" rel="stylesheet" type="text/css" />
-		<link href="./css/stable.css?v=530" rel="stylesheet" type="text/css" />
-		<link rel="preload" as="style" href="./css/category-panel.css?v=530" />
-		<link rel="preload" as="style" href="./css/panel-label.css?v=530" />
-		<script type="text/javascript" src="./js/jquery.js?v=530"></script>
-		<script type="text/javascript" src="./js/jquery.flot.js?v=530"></script>
-		<script type="text/javascript" src="./js/sanitize.js?v=530"></script>
-		<script type="text/javascript" src="./js/sanitize.config.js?v=530"></script>
-		<script type="text/javascript" src="./js/custom-elements.js?v=530"></script>
-		<script type="text/javascript" src="./lang/langs.js?v=530"></script>
-		<link rel="preload" as="script" href="./js/browser.js?v=530" />
-		<link rel="preload" as="script" href="./js/common.js?v=530" />
-		<link rel="preload" as="script" href="./js/objects.js?v=530" />
-		<link rel="preload" as="script" href="./js/options.js?v=530" />
-		<link rel="preload" as="script" href="./js/content.js?v=530" />
-		<link rel="preload" as="script" href="./js/stable.js?v=530" />
-		<link rel="preload" as="script" href="./js/graph.js?v=530" />
-		<link rel="preload" as="script" href="./js/plugins.js?v=530" />
-		<link rel="preload" as="script" href="./js/rtorrent.js?v=530" />
-		<link rel="preload" as="script" href="./js/webui.js?v=530" />
+		<link href="./css/bootstrap.min.css?v=531" rel="stylesheet" type="text/css" />
+		<link href="./css/style.css?v=531" rel="stylesheet" type="text/css" />
+		<link href="./css/stable.css?v=531" rel="stylesheet" type="text/css" />
+		<link rel="preload" as="style" href="./css/category-panel.css?v=531" />
+		<link rel="preload" as="style" href="./css/panel-label.css?v=531" />
+		<script type="text/javascript" src="./js/jquery.js?v=531"></script>
+		<script type="text/javascript" src="./js/jquery.flot.js?v=531"></script>
+		<script type="text/javascript" src="./js/sanitize.js?v=531"></script>
+		<script type="text/javascript" src="./js/sanitize.config.js?v=531"></script>
+		<script type="text/javascript" src="./js/custom-elements.js?v=531"></script>
+		<script type="text/javascript" src="./lang/langs.js?v=531"></script>
+		<link rel="preload" as="script" href="./js/browser.js?v=531" />
+		<link rel="preload" as="script" href="./js/common.js?v=531" />
+		<link rel="preload" as="script" href="./js/objects.js?v=531" />
+		<link rel="preload" as="script" href="./js/options.js?v=531" />
+		<link rel="preload" as="script" href="./js/content.js?v=531" />
+		<link rel="preload" as="script" href="./js/stable.js?v=531" />
+		<link rel="preload" as="script" href="./js/graph.js?v=531" />
+		<link rel="preload" as="script" href="./js/plugins.js?v=531" />
+		<link rel="preload" as="script" href="./js/rtorrent.js?v=531" />
+		<link rel="preload" as="script" href="./js/webui.js?v=531" />
 
 		<script>
 			loadUILang(() => {
@@ -57,10 +57,10 @@
 			});
 		</script>
 
-		<script type="module" src="./js/backgroundtask.js?v=530"></script>
-		<script type="module" src="./js/category-list.js?v=530"></script>
-		<script type="module" src="./js/panel.js?v=530"></script>
-		<script type="text/javascript" src="./js/category-list-elements.js?v=530" defer></script>
+		<script type="module" src="./js/backgroundtask.js?v=531"></script>
+		<script type="module" src="./js/category-list.js?v=531"></script>
+		<script type="module" src="./js/panel.js?v=531"></script>
+		<script type="text/javascript" src="./js/category-list-elements.js?v=531" defer></script>
 	</head>
 	<body>
 		<div id="preload" class="d-none"></div>
@@ -139,7 +139,7 @@
 				</div>
 				<div class="p-1 p-md-0 offcanvas-body">
 					<template id="panel-label-template">
-						<link href="./css/panel-label.css?v=530" rel="stylesheet" type="text/css" />
+						<link href="./css/panel-label.css?v=531" rel="stylesheet" type="text/css" />
 						<div part="prefix" hidden></div>
 						<div part="icon"></div>
 						<div part="text"></div>
@@ -148,7 +148,7 @@
 						<div part="size" style="display: none;"></div>
 					</template>
 					<template id="category-panel-template">
-						<link href="./css/category-panel.css?v=530" rel="stylesheet" type="text/css" />
+						<link href="./css/category-panel.css?v=531" rel="stylesheet" type="text/css" />
 						<div part="heading">
 							<span class="text"></span><slot name="decorator" class="decorator"></slot>
 						</div>
@@ -304,6 +304,6 @@
 		<div class="graph_tab_legend"></div>
 		<div id="dialog-container"></div>
 		<div id="dir-container"></div>
-		<script type="text/javascript" src="./js/bootstrap.bundle.min.js?v=530"></script>
+		<script type="text/javascript" src="./js/bootstrap.bundle.min.js?v=531"></script>
 	</body>
 </html>
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -1,5 +1,10 @@
 // this function is obsolete
 
+function cacheBust(url) {
+	var sep = (url.indexOf('?') === -1) ? '?' : '&';
+	return url + sep + 'v=' + (typeof theWebUI !== 'undefined' ? theWebUI.version.replace(/\./g, '') : Date.now());
+}
+
 function injectScript(fname,initFunc)
 {
 	var h = document.getElementsByTagName("head").item(0);
@@ -9,9 +14,9 @@ function injectScript(fname,initFunc)
 		s.onload = initFunc;
 	}
 	if(s.setAttribute)
-		s.setAttribute('src', fname);
+		s.setAttribute('src', cacheBust(fname));
 	else
-		s.src = fname;
+		s.src = cacheBust(fname);
 	s.type = "text/javascript";
 	void (h.appendChild(s));
 }
@@ -20,7 +25,7 @@ function injectCSS(fname, onLoadFunc)
 {
 	var newSS=document.createElement('link');
 	newSS.rel='stylesheet';
-	newSS.href=fname;
+	newSS.href=cacheBust(fname);
 	newSS.onload = onLoadFunc;
 	var h = document.getElementsByTagName("head").item(0);
 	void (h.appendChild(newSS));
@@ -166,7 +171,7 @@ rPlugin.prototype.loadLangPrim = function(lang,template,sendNotify)
 	var self = this;
 	$.ajax(
 	{
-		url: template.replace('{lang}',lang), // this is because plugin.path may be changed during call
+		url: cacheBust(template.replace('{lang}',lang)),
 		dataType: "script",
 		cache: true
 	}).done( function()
--- a/js/webui.js
+++ b/js/webui.js
@@ -4,7 +4,7 @@
  */
 
 var theWebUI = {
-	version: "5.3.0",
+	version: "5.3.1",
 	tables: {
 		trt: {
 			obj: new dxSTable(),
--- a/lang/langs.js
+++ b/lang/langs.js
@@ -110,7 +110,7 @@ function loadUILang(onLoadFunc)
 			translateDOM();
 		}
 	};
-	langScript.src = `./lang/${lang}.js?v=530`;
+	langScript.src = `./lang/${lang}.js?v=531`;
 	document.head.appendChild(langScript);
 }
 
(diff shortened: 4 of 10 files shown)


fffe:
permitting the load commands breaks the whitelist

```
test@downloads:/# curl -kX POST https://localhost/torrent/plugins/httprpc/action.php \
  --data '<?xml version="1.0"?><methodCall><methodName>load.start</methodName><params>
    <param><value><string></string></value></param>
    <param><value><string>/tmp/file.torrent</string></value></param>
    <param><value><string>execute.nothrow={"sh","-c","echo hello world > /tmp/test.txt"}</string></value></param>
  </params></methodCall>'
<?xml version="1.0"?><methodResponse><params><param><value><i8>0</i8></value></param></params></methodResponse>
test@downloads:/# cat /tmp/test.txt
hello world
```

xirvik:
> permitting the load commands breaks the whitelist
> 
> ```
> test@downloads:/# curl -kX POST https://localhost/torrent/plugins/httprpc/action.php \
>   --data '<?xml version="1.0"?><methodCall><methodName>load.start</methodName><params>
>     <param><value><string></string></value></param>
>     <param><value><string>/tmp/file.torrent</string></value></param>
>     <param><value><string>execute.nothrow={"sh","-c","echo hello world > /tmp/test.txt"}</string></value></param>
>   </params></methodCall>'
> <?xml version="1.0"?><methodResponse><params><param><value><i8>0</i8></value></param></params></methodResponse>
> test@downloads:/# cat /tmp/test.txt
> hello world
> ```

Excellent, thanks for reporting!
So this is not a solution clearly....  I don't have a great idea TBH. We could strip that param, but that'd break other things...

fffe:
I don't think you can avoid this being a breaking change, at least not for anything that assumes you're an xmlrpc proxy.

Transdroid uses another half-dozen commands that aren't marked safe. The others probably use more.

xirvik:
> I don't think you can avoid this being a breaking change, at least not for anything that assumes you're an xmlrpc proxy.

Let's try. Can you take this for a spin? https://github.com/xirvik/ruTorrent/commit/114eec69e94d142ab31be178d7d325ddfad5f59f
 
> Transdroid uses another half-dozen commands that aren't marked safe. The others probably use more.

@erickok thoughts?

fffe:
This fixes the whitelist bypass, but it's a breaking change to just cut off the remainder of the command.

[This person hasn't noticed](https://github.com/Novik/ruTorrent/issues/3046#issuecomment-4321662800), but prowlarr/nzbdrone won't be able to set labels, priorities, or directories when loading torrents.

The most recent release (v2.3.6.5351, apparently) sets those by [tacking them onto the load.start command](https://github.com/Prowlarr/Prowlarr/blob/v2.3.6.5351/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs#L92), but even if it were polling it's ultimately using [.set commands that aren't marked safe](https://github.com/Prowlarr/Prowlarr/blob/v2.3.6.5351/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs#L166).

fffe:
You might want to take a step back and think about how you want this to work for ruTorrent, and what (if any) guarantees you want to make for downstream apps who just use you as an xmlrpc proxy.

Personally, I would treat the RPC socket as a trusted/internal interface and scrap exposing it to users entirely.

At a glance, the repos I looked at (transdroid, prowlarr) were using the same commands to accomplish the same goals: a torrent-adding function will load a torrent, set a directory, set a priority, set tags in one of the custom fields... which is pretty close to what addtorrent.php already does.

I'd try to extend the REST API to meet those needs, and maybe send patches downstream to add an 'rutorrent' target hitting the REST API instead.

xirvik:
> 
> Personally, I would treat the RPC socket as a trusted/internal interface and scrap exposing it to users entirely.

You mean just no proxy at all, either you have directly access to rtorrent or you don't have it at all?
> 
> At a glance, the repos I looked at (transdroid, prowlarr) were using the same commands to accomplish the same goals: a torrent-adding function will load a torrent, set a directory, set a priority, set tags in one of the custom fields... which is pretty close to what addtorrent.php already does.

Yes. But let's think about what's realistic here - don't want to permanently break setups. I think things being in bad shape for a few days while we work out the kinks is totally fine, people who needs stability can stay behind for a bit and use whatever works for them while power users help us test and polish. 

We can't dictate what other projects do of course, so we can't just say "use addtorrent.php from now on instead of building your own payload". 

> I'd try to extend the REST API to meet those needs, and maybe send patches downstream to add an 'rutorrent' target hitting the REST API instead.

That's really out of scope here. To the point that it feels like it should be a separate new project that doesn't depend on ruTorrent at all.

fffe:
> We can't dictate what other projects do of course, so we can't just say "use addtorrent.php from now on instead of building your own payload".

You can deprecate and remove features on whatever timeline suits. They can choose to support the replacement or not.

Should you? Dunno! I'm partial to secure-by-design, but it's not my project.

> That's really out of scope here. To the point that it feels like it should be a separate new project that doesn't depend on ruTorrent at all.

That's my argument too: split it off, make it someone else's problem.

The API isn't designed for untrusted users, and making it safe should be out of scope.

The [current patch](https://github.com/xirvik/ruTorrent/commit/cd05700dcf68d172f51a396087764ed1ceda07b7#diff-aba3b1ff95036a6da488553346f8926af3137cebebaf044fcb7615478b71724fR28-R29) allows `d.directory.set` and `d.directory_base.set`, which means arbitrary file write as rtorrent... and I guess code exec for a lot of setups? (overwrite a cron job or something)

I'd also add that if you end up hardcoding the specific commands these downstream agents use as permitted, it's not really an xmlrpc interface anyway, it's just a crappy REST API.

xirvik:
> You can deprecate and remove features on whatever timeline suits. They can choose to support the replacement or not.

It's not just other projects, it's also users - they may choose to stay on an old version forever. So there's a trade-off here.
> 
> Should you? Dunno! I'm partial to secure-by-design, but it's not my project.

Me too, but secure + useful. 

> The API isn't designed for untrusted users, and making it safe should be out of scope.

Well, depends on how you look at it. Users can be trusted but the stuff they run not be. Plus realistically, shared servers is a very common use case, so boundaries at several layers sounds like a good idea.
> 
> I'd also add that if you end up hardcoding the specific commands these downstream agents use as permitted, it's not really an xmlrpc interface anyway, it's just a crappy REST API.

That's a good point and I agree; just happens to be compatible with what already exists.

fffe:
> It's not just other projects, it's also users - they may choose to stay on an old version forever. So there's a trade-off here.

They might, but I don't think there's a huge risk:
- You're always going to have some users who do that, for any number of reasons... including benign ones like "if it ain't broke, don't fix it."
- Users can still expose the RPC socket directly using their webserver. This is still a bad idea, but now it's not your problem.
- For users that can't (eg no control over the httpd) that portion of the httprpc plugin could be split off, although that would change the url path... either way, an open XMLRPC proxy doesn't seem like a huge dev effort for anyone inclined.

> Me too, but secure + useful.

Sorry, that wasn't meant to sound catty. Apologies if it did. :)

Useful to whom, and useful at what cost?

As far as I can tell, the plugins don't need this interface, they can include xmlrpc.php and make direct calls themselves. The javascript uses the 'defined' interfaces like trkall, setsettings, etc. I'm probably missing something, but I don't see any internal consumers... which leaves external consumers.

A number of other applications have settled on httprpc/action.php as a de facto standard, but that doesn't mean you need to support the API indefinitely: when rtorrent finally removed deprecated APIs, ruTorrent adapted.

The other thing is: I'm not using ruTorrent right now (the tl;dr is that Chainguard's PHP packages suck), but with how I had it configured, my other related applications weren't using ruTorrent as an XMLRPC proxy. They would just drop torrents into a watch directory for rTorrent to pick up, and post-processing was done/initiated by scripts run by rtorrent on completion.

For a setup like mine, having the bare XMLRPC API exposed externally at all offers no benefit, it's _just_ a vulnerability.

If you're going to keep the whitelist approach, would you also consider adding a kill switch option for httprpc to enable/disable the raw XMLRPC case?

> Well, depends on how you look at it. Users can be trusted but the stuff they run not be. Plus realistically, shared servers is a very common use case, so boundaries at several layers sounds like a good idea.

I could probably have explained myself a little better.

There is a security boundary between ruTorrent + rtorrent in most configurations:
  - They may be running as different users
  - They may be running in different containers
  - There is a general expectation that access to a webapp does not mean command execution or arbitrary file access in the context of backend services

I call the XMLRPC API a trusted interface because it's a config and control API, and you generally don't expose those directly to users. It's not designed to be safe, and it's unsafe in non-obvious ways (like load.* taking additional commands to execute as varargs...)

When I talk about trusted users, I mean things like a local user (or the rtorrent process) with access to the actual RPC socket invoking rtxmlrpc or similar; someone who's demonstrably already operating at the same privilege level as rtorrent itself. I'd consider ruTorrent the application a trusted user of the XMLRPC socket because it's functionally at the same privilege level. I _do not_ consider other users of httprpc/action.php as trusted in that sense, because they're not.

This matters _more_ in the shared server case, but it matters in a single-user setup too.

If it's just a download client I don't need authentication on my LAN, but I do if it's RCE-as-a-Service.

> That's a good point and I agree; just happens to be compatible with what already exists.

Yeah, and easier too, which is fair.

I'll leave it with a reminder that what already exists isn't safe (the d.directory*.set bits previously mentioned)

xirvik:
>     * Users can still expose the RPC socket directly using their webserver. This is still a bad idea, but now it's not your problem.

Depends. So there's users that just set everything up themselves and users that just pay seedbox providers to set things up in a sane way that minimizes headaches (us being just one of such providers, but we are active in the general ecosystem and care about everybody to shooting themselves in the foot).

> Sorry, that wasn't meant to sound catty. Apologies if it did. :)

We're just talking :-) 

> A number of other applications have settled on httprpc/action.php as a de facto standard, but that doesn't mean you need to support the API indefinitely: when rtorrent finally removed deprecated APIs, ruTorrent adapted.

Not indefinitely, but there's usually some transition period.

> For a setup like mine, having the bare XMLRPC API exposed externally at all offers no benefit, it's _just_ a vulnerability.

It is for most people; but that's the current state of things. 

> If you're going to keep the whitelist approach, would you also consider adding a kill switch option for httprpc to enable/disable the raw XMLRPC case?

That's what the current PR has - you can completely disable it, or leave it open, or sanitize the input (where we do a best effort filtering - probably it can still be bypassed though which is the same as nothing, which is why it's still not merged).

> There is a security boundary between ruTorrent + rtorrent in most configurations:

Correct; but ultimately you do need to share a bunch of things between ruTorrent, rtorrent, the FTP server and many more containers. So even if you try go have as much sanity as possible, there's no way, or I haven't found it, to have a clear separation that guarantees that one single bad tool won't break a lot more than it should. Ah, and we also have the reality of shared IP address, not just between containers but between users...

> I call the XMLRPC API a trusted interface because it's a config and control API, and you generally don't expose those directly to users. It's not designed to be safe, and it's unsafe in non-obvious ways (like load.* taking additional commands to execute as varargs...)

If key apps in the ecosystem rely on it, you either expose it or tell users they can't have this or that...

PR #3052, opened 2026-05-07
Adding option to have multiple mountpoints in diskspace plugin.

ranirahn:
Adds option to see more mount points same time. Added example to conf.php file to easy understand how it works. On statusbar it will rotate through all mountpoints you added and hovering on the status bar shows all the mountpoints diskusage. Also saving data to rCache file for mutliple reasons. Only thing changing for peole not using this option is tooltip have default download path shown maybe.


xirvik:
@ranirahn can you target develop please?

PR #3055, opened 2026-05-20
Fix/simplify ipad plugin

adenwuts:
iOS Safari 26.5 seems to change the interactions between this plugin and
and touch events, resulting in double-clicks on touch which break any
'toggleable' menus (e.g. add torrent). Unfortunately without this plugin
we lose a working right-click on long-presses and functionality in the
side-bar.

This commit attempts to strip down the plugin to a more minimal set of
features, and rely on iOS to handle more of the work. There may be a
more elegant fix in the ruTorrent internals that would make this plugin
redundant, but this appears to fix the issues on modern iOS versions.

### DIFF
--- a/plugins/ipad/init.js
+++ b/plugins/ipad/init.js
@@ -1,116 +1,82 @@
-$.extend($.support, 
+$.extend($.support,
 {
 	touchable: 'createTouch' in document
 });
 
-plugin.holdMouse = { x:0, y: 0 };
-plugin.curMouse = null;
+let target = null;
+let startPos = null;
+let longPressTimer = null;
+let suppressNextClick = false;
+let suppressSafetyTimer = null;
 
-plugin.emulateRightClick = function()
-{
-	if(( (Math.abs(plugin.rightClick.screenX - plugin.holdMouse.x)<8) &&
-		(Math.abs(plugin.rightClick.screenY - plugin.holdMouse.y)<8)))
-	{
-		var mouseEvent = document.createEvent("MouseEvent");
-		mouseEvent.initMouseEvent("contextmenu", true, true, window, 1, 
-			plugin.rightClick.screenX + 20, plugin.rightClick.screenY + 5, 
-			plugin.rightClick.clientX + 20, plugin.rightClick.clientY + 5,
-			false, false, false, false, 2, null);
-		plugin.rightClick.target.dispatchEvent(mouseEvent);
-		plugin.cancelMouseUp = true;
+const clearLongPress = () => {
+	if (longPressTimer) {
+		clearTimeout(longPressTimer);
+		longPressTimer = null;
 	}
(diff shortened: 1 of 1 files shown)


adenwuts:
@xirvik any blockers on merging this?

xirvik:
> @xirvik any blockers on merging this?

Only that I haven't tested it personally.
Can you rebase it on top of master? I'm done with the develop branch BS - now that we're up to date with all the useful stuff it had, let's work on master.

adenwuts:
> > @xirvik any blockers on merging this?
> 
> Only that I haven't tested it personally. Can you rebase it on top of master? I'm done with the develop branch BS - now that we're up to date with all the useful stuff it had, let's work on master.

Done, thanks @xirvik

xirvik:
@adenwuts I'm not an expert on ipad stuff so I asked a few LLMs to review the PR, take a look:


  1. Missing touchcancel handler (bug). If iOS cancels the touch (edge swipe, notification, app switcher), touchend never fires, so the 600ms timer still goes off — spurious context menu at stale coordinates, plus armSuppress() eats the next real click. Fix is one line:

  document.addEventListener("touchcancel", onTouchEnd, { passive: true });

  2. Suppression window is armed at the wrong moment (likely UX bug). armSuppress() runs when the menu appears (600ms into the hold, finger still down), and the 500ms safety timer starts then. But the compatibility mouse events arrive at release. If the user holds more than ~500ms after the menu appears before lifting (easy to do), the flag has expired, the compat mouseup passes through, and theContextMenu's document mouseup handler hides the menu 50ms later — menu flashes and vanishes. Suggest clearing the flag on the next touchstart instead of (or in addition
  to) a wall-clock timer: the compat events for a gesture always arrive before the next gesture's touchstart, so that's deterministic. It also fixes the inverse case, where iOS sends no compat events and a fast tap on a menu item within the 500ms window gets swallowed.

  3. Form-control exclusion was dropped. The old code skipped select/input/button/label. Now a long-press inside a text input (add-torrent URL field, search box) dispatches a synthetic contextmenu and arms suppression, which can interfere with iOS text selection/paste. Cheap to restore with an early return in onTouchStart when the target matches select, input, textarea, button, label.

  4. Question — native dblclick. The old plugin synthesized dblclick on a second tap; now double-tap actions (e.g. double-tap a torrent row → details) depend on Safari firing native dblclick. With our viewport (maximum-scale=1.0) it should, but did you confirm double-tap actions still work on your device?

  Nits: the e.touches.length !== 1 branch clears the timer but leaves target/startPos stale (harmless, just tidy it); the old code offset the menu +20/+5px so it didn't open directly under the fingertip; and plugin.info is still 5.1 — worth a bump for a full rewrite.

adenwuts:
I'll close this as I'm unlikely to get the time to clean it up in the near future.

PR #3071, opened 2026-06-18
Ratio plugin: gate group.*.set call shape by rtorrent version

xirvik:
#3066 fixed the ratio plugin on rtorrent 0.16.x by always sending (target_string, value) to group.NAME.ratio.{min,max,upload}.set, which is what the strict 0.16 signature requires. On older rtorrent the same set commands are the opposite — strictly a bare value, and the (target, value) shape is rejected with "Wrong object type".

Encapsulate the call-shape adaptation in
rTorrentSettings::getRatioGroupCommand so callers (ratio plugin flush(), but also any future consumer) do not have to know which signature applies. The helper:

* always uses the canonical group.* prefix. group2.* was a transient alias name: stubbed and non-functional on 0.15.x and removed in 0.16. The old "if iVersion >= 0x904 and cmd in ratioCmds use group2." branch fell into both broken cases and produces fewer successful calls than the unconditional group.* path on every rtorrent version verified.

* prepends an empty-string target to *.set commands only when iVersion >= 0x1000, so 0.15.x callers continue to receive the single-value shape they require and 0.16+ callers receive the (target, value) shape they require.

ratio.php is reverted to passing the bare value to the helper for the .set commands; the helper does the version-aware wrapping.

The unused $ratioCmds whitelist is dropped — only the group2.* branch consulted it, and that branch is gone.

Reported by @ml-1 in #3065 after upgrading to v5.3.3 with rtorrent 0.15.7.

### DIFF
--- a/php/settings.php
+++ b/php/settings.php
@@ -32,18 +32,6 @@ class rTorrentSettings
 
 	static private $theSettings = null;
 
-	private $ratioCmds = array
-	(
-		"view",          
-		"view.set",
-		"ratio.min",
-		"ratio.min.set",
-		"ratio.max",
-		"ratio.max.set",
-		"ratio.upload",
-		"ratio.upload.set",
-	);
-
 	private function __construct()
     	{
     		if( array_key_exists("browser_timezone",$_COOKIE) )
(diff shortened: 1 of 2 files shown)


xombiemp:
I tried these changes and I'm still getting ratio: Plugin failed to start. on 0.15.6.

xirvik:
> I tried these changes and I'm still getting ratio: Plugin failed to start. on 0.15.6.

I need complete logs here in order to figure it out - just the error message is not enough.

PR #3085, opened 2026-07-09
Update deprecated rTorrent commands for v0.16.16+

drrako:
## Summary

rTorrent v0.16.16 removed several commands in favor of new canonical names. This PR adds a new `php/methods-0.16.16.php` alias file loaded at `iVersion >= 0x1010` (rTorrent 0.16.16+) with matching JS aliases in `js/content.js`.

| Old | New |
|---|---|
| `network.http.proxy_address` / `.set` | `network.proxy.http` / `.set` |
| `network.proxy_address` / `.set` | `network.proxy.global` / `.set` |
| `d.multicall2` | `d.multicall` |
| `network.open_sockets` | `system.sockets.size` |
| `network.max_open_sockets` | `system.sockets.max_size` |
| `network.http.max_total_connections.set` | `system.sockets.http.max_alloc.set` |
| `network.max_open_files.set` | `system.sockets.files.max_alloc.set` |

Older rTorrent versions continue to use the original aliases from `methods-0.9.4.php` / `methods-0.16.0.php`. The new file only takes effect on 0.16.16+.

### DIFF
--- a/js/content.js
+++ b/js/content.js
@@ -896,6 +896,27 @@ function correctContent()
 			"ratio.upload.set"  : { name: "group.seeding.ratio.upload.set",        prm: 1 },
 		});
 	}
+	if(theWebUI.systemInfo.rTorrent.iVersion>=0x1010)
+	{
+		// rtorrent >= 0.16.17: override aliases for commands that were
+		// removed or now log deprecation warnings. Mirrors php/methods-0.16.16.php.
+		$.extend(theRequestManager.aliases,
+		{
+			"get_max_open_http"      : { name: "system.sockets.http.min_alloc",          prm: 0 },
+			"set_max_open_http"      : { name: "system.sockets.http.min_alloc.set",      prm: 1 },
+			"set_max_open_files"     : { name: "system.sockets.files.min_alloc.set",     prm: 1 },
+			"d.multicall"            : { name: "d.multicall",                             prm: 1 },
+			"d.multicall2"           : { name: "d.multicall",                             prm: 1 },
+			"get_http_proxy"         : { name: "network.proxy.http",                     prm: 0 },
+			"set_http_proxy"         : { name: "network.proxy.http.set",                 prm: 1 },
+			"http_proxy"             : { name: "network.proxy.http",                     prm: 0 },
+			"get_proxy_address"      : { name: "network.proxy.global",                    prm: 0 },
+			"set_proxy_address"      : { name: "network.proxy.global.set",                prm: 1 },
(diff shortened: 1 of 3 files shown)


xirvik:
Thanks. I've been looking into this myself. 

Notes  about these three:

"get_max_open_http"  => "system.sockets.http.min_alloc"       // was network.http.max_total_connections
"set_max_open_http"  => "system.sockets.http.min_alloc.set"
"set_max_open_files" => "system.sockets.files.min_alloc.set"  // was network.max_open_files.set

min_alloc isn't the equivalent of the old cap — it's a guaranteed floor the socket manager reserves. All the per category min_alloc floors (plus a reserve) must fit under open_max — i.e. the soft RLIMIT_NOFILE captured at Poll::create(). If the sum exceeds it, libtorrent throws in set_max_size_and_adjust():

 torrent::internal_error: set_max_size_and_adjust: total + min_generic + 8 reserve exceeds max_open : 10944 + 512 + 8 > 1024

That's an uncaught throw on a worker thread → terminate() → SIGABRT, not a graceful RPC fault. So a user whose "max open files" is any sizeable number now pushes that value into files.min_alloc, and:  - on a stock box at the default ulimit of 1024 it crashes rtorrent on config load, and - because ruTorrent pushes the setter when settings are saved, it can abort an already-running instance the moment the user hits save.

We hit this exact set_max_size_and_adjust abort on rtorrent 0.16.17 (Debian, threaded-resolver libcurl) — the 10944 + 512 + 8 > 1024 above is a real capture. It only stays hidden if you test at a raised ulimit.

Proposed fix — map to the ceiling (max_alloc), not the floor:

"get_max_open_http"  => "system.sockets.http.max_alloc"
"set_max_open_http"  => "system.sockets.http.max_alloc.set"
"set_max_open_files" => "system.sockets.files.max_alloc.set"  (and the same in js/content.js)

This preserves the original cap semantics (max → max), and a max_alloc ceiling — default 1000000 — never forces a reservation that can exceed open_max, so it can't trigger the abort. Setting a ceiling below current usage just declines new sockets; it never crashes.

These two olds still exist (only warn) on 0.16.16/0.16.17, so there's no urgency forcing a risky remap — the max_alloc swap is safe today. Alternatively, since the socket manager self-sizes within open_max, dropping the two setters entirely is also defensible; but max_alloc keeps the UI knob working. The renames for proxy / d.multicall2 / socket getters all look correct.

drrako:
Thanks @xirvik , I've updated PR.

PR #3093, opened 2026-07-12
ignore config.local.php for deployment-specific overrides

cpugeniusmv:


xirvik:
You should also add plugins/*/conf.local.php if you want to exclude these files (I'm fine with it).
Also, typo in the title, it's config, not conf 
@cpugeniusmv

cpugeniusmv:
updated based on your feedback @xirvik

PR #3099, opened 2026-07-13
Add alias finished and started

ranirahn:
I looked sins what version of rtorrent these were in rtorrent and added these. Example timestamp.finished is exactly the same thing then seedingtime in seedingtime plugin. Seems like should have used that for this from the beginning not add new custom value into rtorrent session file. After that is added could rewrite seedingtime plugin to use timestamp.finished instead of custom_seedingtime and remove adding this seedingtime value in the session file. Sadly did not find a replacement for addtime so that we need to keep.

### DIFF
--- a/js/content.js
+++ b/js/content.js
@@ -618,6 +618,8 @@ function correctContent()
 			"d.get_skip_total"	:	{ name: "d.skip.total", prm: 0 },
 			"d.get_state"		:	{ name: "d.state", prm: 0 },
 			"d.get_state_changed"	:	{ name: "d.state_changed", prm: 0 },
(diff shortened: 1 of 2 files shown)


xirvik:
I'll wait for your rewrite on that plugin before merging.

ranirahn:
I separated it just because both of these can be used in any plugin. Not sure if I even want to rewrite the seedingtime plugin. This may be too complicated for me. But if these dont exist, there is no way someone can rewrite the plugin before these are added. I can add the workaround that I have right now, but that still uses the seedingtime value if it exists. I am just not sure why these 2 was just left out in the first place.

PR #3119, opened 2026-07-16
Adding option to have multiple mountpoints in diskspace plugin

ranirahn:
I hope i get the develop base now.


xirvik:
@ranirahn Right base is master, forget develop.
The reason for needing the rebase is that most files have been properly formatted so all PRs wouldn't merge cleanly.

PR #3134, opened 2026-07-24
mobile: fix the empty torrent list on the direct XML-RPC mount, plus a round of fixes

xombiemp:
## Summary

Follow-up to #3128. The headline fix is the empty torrent list reported by @Abzie in the #3128 discussion, which affects installs using the direct XML-RPC mount (no httprpc):

* The mobile tracker fetch sent `?action=getalltrackers` with **no hashes**. httprpc happens to treat an empty hash list as "all torrents", but the direct mount's stub (`rTorrentStub.prototype.getalltrackers`) builds one `t.multicall` per *given* hash — with none given it produces zero commands, and `rTorrentStub` silently drops command-less requests without invoking the callback. The mobile list render runs inside that callback, so the list stayed empty forever while settings/add pages worked. The fetch now names every hash explicitly, exactly like the desktop's `getAllTrackers`, and short-circuits the zero-torrent case (the render also removes deleted torrents' rows, so it must still run).

Also in this round:

* Torrents that appear between full renders (e.g. added from another session) are inserted at their sorted position instead of being appended to the end of the list
* Torrents added from elsewhere (another session, rss, automation) show a toast, like the existing deletion toasts; a local add keeps its upload receipt and scroll-to-torrent behavior instead
* The on-screen keyboard is dismissed on scroll for all text fields, not just the quick search (iOS anchors position:fixed elements to the layout viewport while it's open)
* A new-label edit commits only via its check button or the return key; tapping or scrolling away cancels instead of saving

## Test plan

* On a direct XML-RPC mount (no backend plugin): open the mobile UI — the torrent list renders (regression: also verify on httprpc, where explicit hashes >50 collapse back to the all-torrents form server-side)
* Two sessions: add a torrent in one — the other inserts it at its sorted position and toasts "Added: name"
* Add locally: receipt toast, return to list, scroll+highlight to the new torrent, no "Added" toast
* Edit a label via "New label": check button and return save; tapping/scrolling away reverts
* Focus any add-form field, scroll — keyboard dismisses, fixed bars stay put

### DIFF
--- a/plugins/mobile/init.js
+++ b/plugins/mobile/init.js
@@ -1983,12 +1983,21 @@ plugin.processTorrents = function(torrents, singleUpdate) {
     // the rows of deleted torrents)
     process({}, false);
   } else {
-    // Name every hash explicitly, like the desktop's getAllTrackers:
-    // the direct XML-RPC mount builds one t.multicall per given hash
-    // and silently sends nothing when none are given, which left the
-    // list empty forever (httprpc happens to treat "no hashes" as
-    // "all torrents", masking this)
-    mobile.request('?action=getalltrackers' + $.map(hashes, function(h) {return '&hash=' + h;}).join(''), function(data) {
+    // The PHP transports (httprpc and its out-of-tree siblings) treat a
+    // hashless getalltrackers as "all torrents" — exactly what's wanted
+    // here, and naming thousands of hashes can overflow transport
+    // limits. Only the direct XML-RPC mount can't take that form: the
+    // base stub builds one t.multicall per given hash and silently
+    // sends nothing when none are given, which left the list empty
+    // forever. Probe how the active transport handles the hashless
+    // form and enumerate only where it's needed; the base stub packs
+    // the hashes into POSTed XML, so there's no size problem there
+    var query = '?action=getalltrackers';
+    var probe = new rTorrentStub(query);
+    if (probe.content == null && !probe.commands.length) {
+      query += $.map(hashes, function(h) {return '&hash=' + h;}).join('');
+    }
+    mobile.request(query, function(data) {
       plugin.trackersCache = {key: trackersKey, data: data};
       process(data, true);
     });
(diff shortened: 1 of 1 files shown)


xombiemp:
@Abzie Please try this along with https://github.com/Novik/ruTorrent/pull/3133 and let us know if it fixes the issue.

Abzie:
@xombiemp I have cherry-picked both commits and the torrent list is populating now... It is looking good :)

xirvik:
Tested this on a large torrent list and it regresses into the same blank screen it fixes. 

Claude explains it better, sorry for the slop:

getalltrackers is a GET and this names every hash in the query string; on a ~950-torrent view that's a ~46 KB URL, which nginx rejects with 414 Request-URI Too Large before it reaches PHP — the request fails, the callback never runs, list stays empty. The wall is nginx's default large_client_header_buffers (8 KB) ≈ ~150 hashes (measured: 105 hashes/5 KB reaches PHP, 950/46 KB → 414). Small lists test clean, which is why the reporter's did, but a few hundred+ torrents gets the
  blank list back — on any backend, since the mobile fetch bypasses the multirpc stub's >50→all collapse. The >50 collapse server-side note doesn't help: the URL is rejected before the server can collapse it.

  Options:
  - POST the hashes instead of GET (no URL-length limit), or
  - chunk into ≤50-per-request batches and merge, or
  - keep the empty-hash "all" form where the backend supports it and enumerate only on the direct XML-RPC mount (the case that needs it).

In any case, test it (well, everything) against a rtorrent/ruTorrent with around 5k torrents loaded. It's more common than you'd think.

xombiemp:
@xirvik Not sure I'm following this one completely.
What is multirpc? Is that an out of tree backend?

AI seems to think the in repo backends are not affected:

On stock httprpc, the wire request for a large list is byte-for-byte identical before and after our PR. On the direct mount, the hashes travel as an XML POST body — no URL, no 414. The regression needs a transport that (a) puts the query in a GET URL and (b) does its >50 collapse somewhere our fetch doesn't pass through.

So is there a backend I'm not aware of?

xombiemp:
@xirvik I implemented your suggestion 3. Can you try it out on your system where it failed previously?

xirvik:
> @xirvik I implemented your suggestion 3. Can you try it out on your system where it failed previously?

I'm not totally sure the fail wasn't due to my own tests - sorry about this, I suspect I made you waste a bit of time on this one.

PR #3135, opened 2026-07-24
mobile: log swallowed request failures to the console

xombiemp:
## Summary

`plugin.request` drops timeouts and errors by design (the next poll retries), and an exception thrown inside a response handler died invisibly in the ajax machinery. Each of the recent blank-torrent-list bugs was hidden by exactly this silence: the disabled-backend 500s from the #3128 review, the label-panel TypeError (#3133), and the direct-mount tracker fetch (#3134) all produced a blank list with nothing to go on.

This logs all three cases to the console, tagged with the request that owned them:

* `mobile: handler for <url> failed: <error>` (handler exceptions, with stack)
* `mobile: request timed out: <url>`
* `mobile: request failed: <url> (<status>: <text>)`

Console-only diagnostic breadcrumbs — no notifications, no behavior change (failed requests are still retried by the next poll cycle).

## Test plan

* Normal use: console stays quiet
* Point a request at a bogus action or stop the backend: each poll logs a warning naming the failing request
* Throw inside a response handler: logged with stack, poll loop keeps running

### DIFF
--- a/plugins/mobile/init.js
+++ b/plugins/mobile/init.js
@@ -181,7 +181,27 @@ plugin.backListener = function() {
 };
 
 plugin.request = function(url, func) {
-  theWebUI.requestWithTimeout(url, function(d){if (func != undefined) func(d);}, function(){}, function(){});
+  theWebUI.requestWithTimeout(url,
+    function(d) {
+      if (func == undefined) {
+        return;
+      }
+      try {
+        func(d);
+      } catch (e) {
+        // Log which request's handler blew up, without user-facing
+        // noise; silent failures here have hidden real bugs
+        console.error('mobile: handler for ' + url + ' failed:', e);
+      }
+    },
+    function() {
+      // Failures are swallowed by design — the next poll retries — but
+      // leave a trace in the console
+      console.warn('mobile: request timed out: ' + url);
+    },
+    function(status, text) {
+      console.warn('mobile: request failed: ' + url + ' (' + status + (text ? ': ' + text : '') + ')');
+    });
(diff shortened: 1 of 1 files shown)


xirvik:
As written this won't actually log the #3133 case you mention. That TypeError is thrown inside processTorrents(), which is only reached from plugin.update()'s direct theWebUI.requestWithTimeout(...) call and from the theWebUI.addTorrents hook — neither goes through plugin.request, so the new try/catch never sees it. The #3134 tracker fetch is caught (it uses mobile.request), but the initial-render throw isn't.

Two small additions cover it:

Route the list fetch through plugin.request so its handler gets wrapped:

```
  plugin.update = function(singleUpdate) {
    plugin.request('?list=1&getmsg=1', function(data) {
      plugin.processTorrents(data.torrents, singleUpdate);
    });
  };

```

And guard the addTorrents hook (not a request, so it needs its own):
```

  theWebUI.addTorrents = function(listData) {
    plugin.desktopAddTorrents.call(this, listData);
    if (listData && listData.torrents) {
      try { plugin.processTorrents(listData.torrents); }
      catch (e) { console.error('mobile: processTorrents (addTorrents hook) failed:', e); }
    }
  };
```

  With those, a future processTorrents regression would actually leave a breadcrumb instead of a silent blank list.

xombiemp:
Good catch. Both suggestions applied: the list fetch now goes through plugin.request, and the addTorrents hook got its own guard. Every path into processTorrents leaves a named breadcrumb now.

PR #3139, opened 2026-07-28
Feat: Convert URLs in torrent comments into clickable links

jakobbg:
## Description
This PR updates the torrent description view to automatically identify plaintext URLs (`http` or `https`) within the torrent comments and convert them into clickable HTML hyperlinks.

## Changes
- Created a safe, type-hinted helper function `getClickableComment()` to handle URL regex replacement.
- Sanitizes raw user input with `htmlspecialchars()` before parsing links to prevent XSS vulnerabilities.
- Applied secure link attributes (`target="_blank"` and `rel="noopener noreferrer"`) so links safely open in a new tab without navigating away from the application.

## Testing Instructions
1. View a torrent containing a plain URL in its comment (e.g., `Trump: https://torrent.site/torrent.php?id=12345`).
2. Verify that the URL turns into a clickable link.
3. Confirm that clicking the link opens the target page in a new browser tab.
4. Verify that normal text in the comment remains unaffected and safe.

### DIFF
--- a/plugins/feeds/action.php
+++ b/plugins/feeds/action.php
@@ -105,8 +105,15 @@ function formatItemDescription($torrent)
 	if(!empty($torrent["mesage"]))
 		$desc.='<fieldset><legend>'.$theUILang["Track_status"].'</legend>'.htmlspecialchars($torrent["mesage"],ENT_COMPAT,"UTF-8").'</fieldset>';
 	if(!empty($torrent["comment"]))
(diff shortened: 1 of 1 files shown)


xirvik:
Please target master - develop is no longer in use. @jakobbg

xirvik:
Two things:

- The title says "torrent comments", but `formatItemDescription()` builds  `<item><description>` for the outbound RSS feed — this affects what a feed reader shows, not comments in the ruTorrent UI. If the UI is what you were after, this is the wrong place. Otherwise please reword.

The regex includes `ftp` but the description says http/https. Browsers dropped FTP, so those links are dead — drop it or say so.

 Minor: `[^\s<>"]+` swallows trailing punctuation, so `see http://x/y.` links the final  `.` (same for `)` and `,`).

PR #3148, opened 2026-08-03
chore: ignore .htaccess files in git repository

jakobbg:
... because I have a use for it myself :-)

Thanks!


xirvik:
This can't be merged - we do have .htaccess files committed, with very good reason -  and those are already tracker so would survive this, but we might need to add more.

You can just add the exclusion to your local clone:

```
 echo '.htaccess' >> .git/info/exclude
```

PR #3149, opened 2026-08-03
Fix stale time deltas after browser sleep/tab suspension

jakobbg:
deltaTime and serverDeltaTime are computed once on the first AJAX response and cached for the lifetime of the page. When the browser sleeps or the tab is suspended, the cached values become stale, causing all displayed timestamps to be offset by the sleep duration.

Reset both values to 0 on visibilitychange so Ajax_UpdateTime recalculates them from the next fresh response.

Co-authored-by: AI (Gemini via Antigravity IDE)

### DIFF
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -1376,6 +1376,16 @@ function Ajax_UpdateTime(jqXHR)
 	jqXHR = null; // Cleanup memory leak
 }
 
+// Recalculate time deltas after browser sleep/tab suspension
+document.addEventListener("visibilitychange", function()
+{
+	if(!document.hidden)
+	{
+		theWebUI.deltaTime = 0;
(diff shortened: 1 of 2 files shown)


xirvik:
One issue: the test leaks its mock.

 `beforeEach` saves the descriptor from `Document.prototype`, but the tests define `hidden` on the instance (`document`), and  `afterEach` restores only the prototype — the own property is never removed. The  second test happens to redefine it, which is why nothing fails today, but a third  test that expects real `document.hidden` would inherit the stub. Adding   `delete document.hidden;` to `afterEach` fixes it.

jakobbg:
Thanks for catching that! The last commit (8043ca1) cleans up properly. Appreciate the review, @xirvik!

PR #3164, opened 2026-08-15
lint: use the config filename editorconfig-checker 5 actually reads

IvanShift:
**Problem.** `.github/workflows/lint.yml` runs `npx editorconfig-checker@5`, and v5
reads `.editorconfig-checker.json`; `.ecrc` is the deprecated v2 name. The repo's
Exclude list (minified assets, vendored jquery/bootstrap/codemirror, test fixtures)
is therefore ignored: the job warns about the filename and reports findings the
project explicitly excluded.

**Fix.** `git mv .ecrc .editorconfig-checker.json` — content byte-identical.

**How to verify.** Before: `npx --yes editorconfig-checker@5` from the repo root
prints the deprecated-filename warning and flags vendored files. After: exclusions
honoured, warning gone.


xirvik:
Taking this — the deprecation warning is real and renaming ahead of the removal is worth doing.
 
 The description overstates it though. npx --yes editorconfig-checker@5 installs checker v3.11.1, which still reads .ecrc and applies it; it only warns. Checked here: with .ecrc in place the run is 0 findings (including a trailing-whitespace line I added to an excluded file to be sure the Exclude list was live), and it only starts reporting excluded paths once .ecrc is moved away. So the exclusions were being honoured, and the before/after in "How to verify" differs only by the warning line.

PR #3165, opened 2026-08-15
Snoopy: resolve a protocol-relative Location against its own host

IvanShift:
**Problem.** A `Location: //example.org/login.php` is a network-path reference
(RFC 3986 §4.2) carrying its own authority. Snoopy tests for `://`, treats it as a
bare path, and requests `https://original-host//example.org/login.php` — looping
until `maxredirs`, so the caller sees a redirect loop instead of the page the server
pointed at. Kinozal serves exactly this shape of redirect on its login flow, which is
where it was found.

**Fix.** Resolve the authority from the Location itself, in both transports (curl
and the plain-socket path). Also renames the PHP 8.5-deprecated
`socket_set_timeout` to `stream_set_timeout`.

**How to verify.** `tests/php/SnoopyTest.php` grows 3 → 5 tests, one per transport
(fake-curl double; a real socket pair for the plain path). Revert the
`Snoopy.class.inc` hunks and the two new tests fail with the looping URL; restore and
they pass.

### DIFF
--- a/php/Snoopy.class.inc
+++ b/php/Snoopy.class.inc
@@ -474,7 +474,9 @@ class Snoopy
 
 		// set the read timeout if needed
 		if ($this->read_timeout > 0)
-			socket_set_timeout($fp, $this->read_timeout);
+			// socket_set_timeout() is the same function under a name PHP 8.5
+			// deprecates; stream_set_timeout() has existed since PHP 4.3.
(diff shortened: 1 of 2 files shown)


xirvik:
@IvanShift can you rebase please?

IvanShift:
> @IvanShift can you rebase please?

Done, check please.

PR #3170, opened 2026-08-15
rss: normalize legacy XML document encodings before parsing

IvanShift:
**Problem.** Feeds still commonly declare windows-1251/koi8-r/iso-8859-x. A clean
such document parses fine — but real aggregators emit whitespace before the XML
declaration, and libxml then reports "XML declaration allowed only at the start",
recovers, and reads the cp1251 bytes as UTF-8: every non-ASCII title is mojibake
(`json_encode` of the result returns `false` outright).

**Fix.** Convert the document to UTF-8 up front and rewrite the declaration to match
the bytes handed to the parser — but only when the bytes are not already valid UTF-8:
feeds whose declaration *lies* (says windows-1251, bytes already UTF-8) must not be
converted twice. Detection falls back `mb_check_encoding` → strict `iconv`
round-trip → PCRE `//u`. The regex reader's per-item path shares the same helper
instead of its ad-hoc iconv/mb cascade. Unconvertible input is returned unchanged.

**How to verify.** Three fixtures in `tests/plugins/rss/` (all with the leading
blank line): declared-and-actual cp1251 — mojibake before the fix, correct
`Русский торрент канал` after; pure UTF-8 — byte-identical pass-through both ways;
declared cp1251 but actually UTF-8 — parsed correctly before (by libxml accident),
mojibake with a naive converter, correct with this guard.

### DIFF
--- a/plugins/rss/rss_reader.php
+++ b/plugins/rss/rss_reader.php
@@ -1,6 +1,77 @@
 <?php
 require_once(dirname(__FILE__) . "/../../php/util.php");
 
+function rssIsValidUTF8($data)
+{
+	if (function_exists('mb_check_encoding')) {
+		return mb_check_encoding($data, 'UTF-8');
+	}
+	if (function_exists('iconv')) {
+		return @iconv('UTF-8', 'UTF-8', $data) === $data;
+	}
+	// PCRE is always available: the /u modifier rejects invalid UTF-8.
+	return preg_match('//u', $data) === 1;
+}
+
+function rssConvertToUTF8($data, $encoding)
+{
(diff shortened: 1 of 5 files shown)


xirvik:
A bit out of my depth, I suppose this is related to rutracker... 

Merged it based on claude's review:

 One correction for the record: the symptom isn't a hard failure. The plugin answers through JSON::safeEncode, which sanitises invalid bytes to U+FFFD since #2996, so the feed loads and the titles come out as ��� — and an RSS filter matching on the title then stops matching with nothing in the log. Also worth knowing the trigger is narrower than the description implies: libxml honours a legacy declaration on its own, and tolerates a BOM. I measured all five shapes; only whitespace before <?xml breaks, and the four healthy ones are byte-identical before and after this patch.

PR #3179, opened 2026-08-16
feat:webui: give a finished partial download its own status 

IvanShift:
**Problem.** rTorrent calls a download complete only when *every* chunk is on
disk, so a torrent whose deselected files own at least one chunk outright never
qualifies — those chunks are never fetched. Once everything the user selected
has arrived, ruTorrent still says **Downloading** forever: zero speed, nothing
left to fetch, and the row never leaves the downloading group or reaches
Completed. The status was derived from the percentage alone, and the percentage
counts the whole torrent.

**Fix.** The daemon already answers the other question — `d.is_partially_done`
reports whether every *selected* chunk is on disk — ruTorrent just never asked.
It is requested as the last field of the list multicall on both sides of the
mirror and exposed as `torrent.partially_done`. A started torrent whose selected
files are complete gets the seeding icon, a status of its own (**Partial seed**,
the term BEP 21 uses for this peer) and membership in the completed category; a
stopped one is treated as a stopped completed torrent, an errored one gets the
upload error icon. The scheduler compares status text against Seeding and
Downloading to decide whether a throttle change must stop and restart a torrent,
so it gets the new term too. The displayed percentage is deliberately untouched:
it keeps counting the whole torrent, so the progress column and Remaining still
show what is missing.

`d.is_partially_done` arrived in 0.9.0, and an unknown command faults the whole
`d.multicall` rather than one field, so `php/methods-pre-0.9.0.php` routes it to
the no-op `cat` on older daemons — the field keeps its position and reads as 0.

**How to verify.** Add a multi-file torrent, deselect some files, let the
selected ones finish.
*Before:* the row shows **Downloading** with 0 B/s forever, stays in the
Downloading group and never appears under Completed.
*After:* it shows **Partial seed** with the seeding icon and sits with the
completed torrents; the percentage still reads below 100 %.

Test-side, each claim is pinned and was verified by mutation — every one of
these turns exactly one test red:

| mutation | result |
|---|---|
| derive the flag from the percentage (`done < 1000 ? 0 : …`) — i.e. make the feature dead code | 1 failed |
| drop `partially_done` from the mobile row snapshot | 1 failed |
| stop the alias gate covering the `iVersion 0` sentinel | 1 failed |

**Tested.** PHP suite 70 `ok` on 8.1 and 8.5; jest 17 suites / 178 tests, all
green. The 397-torrent observation in the commit message comes from a live
rTorrent 0.16.20.

**Note for review.** The new `php/methods-pre-0.9.0.php` is the repository's
first *ceiling* gate (`if($this->iVersion<0x900)`); `RtorrentCompatibilityTest`'s
gate parser was extended to read both comparison directions so the file is
replayed like any other.

### DIFF
--- a/js/category-list.js
+++ b/js/category-list.js
@@ -24,7 +24,10 @@ export class CategoryList {
       pstate: [
         (_, torrent) => [
           // Main group (Completed / Downloading / Stopped)
-          (torrent.done >= 1000)
+          // A torrent whose deselected files own at least one chunk outright
+          // never reaches done >= 1000 -- those chunks are never fetched -- so
+          // rTorrent's d.is_partially_done counts as completed here as well.
+          (torrent.done >= 1000 || torrent.partially_done == 1)
             ? "-_-_-com-_-_-" // Completed (even if error)
             : (torrent.state & this.dStatus.paused)
             ? "-_-_-wfa-_-_-" // Truly paused (Stopped)
--- a/js/content.js
+++ b/js/content.js
@@ -567,6 +567,16 @@ function correctContent()
 			"d.set_connection_seed"		: { name: "d.connection_seed.set", prm: 0 }
 		});
 	}
+	// rTorrent gained d.is_partially_done in 0.9.0. An unknown command faults the
+	// whole d.multicall rather than one field, so on older daemons route it to
+	// the harmless no-op "cat": it answers with an empty string, which keeps the
+	// field in place and reads as "not partially done". Firing at iVersion 0
+	// (daemon unreachable at load) is deliberate: cat is understood by every
+	// daemon generation, so an unknown version degrades safely.
+	if(theWebUI.systemInfo.rTorrent.iVersion<0x900)
+	{
+		theRequestManager.aliases["d.is_partially_done"] = { name: "cat", prm: 0 };
+	}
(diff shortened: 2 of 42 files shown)


xirvik:
Good stuff @IvanShift , thanks! Just tested it on my personal instance

IvanShift:
> Good stuff @IvanShift , thanks! Just tested it on my personal instance

Thanks for testing it! A few more PRs are coming next days.

The next one is big: RuTracker shut down the API this plugin used, so the handler is rebuilt around the forum pages. It's about a dozen production files — the rest of the diff is tests and 26 locale files.

Happy to split it if you'd rather review it in pieces.

xirvik:
> The next one is big: RuTracker shut down the API this plugin used, so the handler is rebuilt around the forum pages. It's about a dozen production files — the rest of the diff is tests and 26 locale files.

I can't test ruTracker very well since I don't speak Russian (I did create an account to test another PR you submitted and claude helped a bit with the translation) but I'll do what I can. 
> 
> Happy to split it if you'd rather review it in pieces.

No need to split, since this one will be a test live (best effort) plus claude+codex+Kimi reviews.

Let's say you own the RuTracker bit and I just sanity check :-)

PR #3195, opened 2026-08-26
Escape the download directory in the load command string

### LINKED ISSUE #3194: Torrent added into a path containing a double quote gets no directory set

- [x] Web browser cache cleared
- [x] Link provided to install script if applicable
- [x] Not using broken rtinst install script
- [x] Web browser, ruTorrent, PHP and OS version provided

Web Browser: Google Chrome Version 153.0.8010.5 (Official Build) beta (x86_64)
ruTorrent: v5.3.12
PHP: PHP 8.4.21 (cli) (built: May  7 2026 16:01:59) (NTS)
OS: Linux 7.1.8+deb14.1-amd64

This Docker image: https://github.com/crazy-max/docker-rtorrent-rutorrent

### Describe the bug

Adding a torrent with a destination path that contains a double quote leaves the download with no directory set. The torrent appears in the list with an empty save path and fails to write its data.

Setting the very same path afterwards via right click -> `Save to...` (DataDir plugin) succeeds, so the problem looks intermittent — the second attempt always works — but it is deterministic: the first attempt never works for such a path.

A path containing a backslash is affected the same way.

### Steps to reproduce

1. Add a torrent through the web UI (file or magnet).
2. As the destination, choose or type a path containing a double quote, for example `/some/path/Richard "Popcorn" Wylie`.
3. Add the torrent.
4. See the torrent appear with an empty save path and fail to save data.
5. Right click the torrent -> `Save to...` -> pick the same path again. It now works and the download starts.

### Expected behavior

The torrent is added with its directory set to the given path, on the first attempt, exactly as `Save to...` manages to do afterwards.

### Additional context

All ruTorrent references below are permalinks to master as of [515a738](https://github.com/Novik/ruTorrent/commit/515a738eaa88183186ad09ac71c29a4fce0f2de5), the current tip at the time of writing.

The download directory is not sent to rtorrent as a typed XMLRPC parameter. It is embedded in a command string that rtorrent parses itself, and `php/rtorrent.php` builds that string by concatenating the raw path between two literal quotes:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/php/rtorrent.php#L75

For the path above this produces:

```
d.directory.set="/some/path/Richard "Popcorn" Wylie"
```

rtorrent's command parser ends a quoted argument at the first *unescaped* `"`, and treats `\` as an escape character — [`src/rpc/parse.cc` L37-L50 @ rakshasa/rtorrent ac1b268](https://github.com/rakshasa/rtorrent/blob/ac1b2685d5ed224b36c2ded94694c1ff8350ca1d/src/rpc/parse.cc#L37-L50):

```cpp
  if (parse_is_quote(*first)) {
    first++;

    while (first != last) {
      if (parse_is_quote(*first))
        return ++first;

      if (parse_is_escape(*first) && ++first == last)
        throw torrent::input_error("Escape character at end of input.");

      dest->push_back(*first++);
    }

    throw torrent::input_error("Missing closing quote.");
```

So the argument ends right after `Richard `, and the remainder of the command is left unparsable. The directory is never applied, which is the empty save path.

`sendMagnet()` builds the very same string, so adding a magnet link into such a path fails identically:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/php/rtorrent.php#L127

**Why `Save to...` is not affected:** it takes a different route, passing the path as its own XMLRPC parameter, so rtorrent's command parser never sees it:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/plugins/datadir/util_setdir.php#L221-L223

The `execute mkdir -p` call in `sendTorrent()` also passes the path as separate parameters, which is why the directory itself does get created even when the failure happens:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/php/rtorrent.php#L74

**ruTorrent already escapes this correctly elsewhere.** `XMLRPCProxy::rebuildSafeLoadParam()` escapes `\` and `"` before quoting a command argument:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/php/xmlrpc_proxy.php#L648

That behaviour is pinned by the contract fixture entry ["quotes and backslashes in a value are escaped, not dropped"](https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/tests/php/XMLRPCProxyContractFixture.php#L282-L296). The two call sites above simply do not use it.

`rtAddTorrent()` carries the same unescaped concatenation, but it has no callers anywhere in the repository, so it seems out of scope for a fix:

https://github.com/Novik/ruTorrent/blob/515a738eaa88183186ad09ac71c29a4fce0f2de5/plugins/datadir/util_rt.php#L390-L393

I have a fix with unit tests and will open a PR referencing this issue.

noctuum:
Fixes #3194.

`sendTorrent()` and `sendMagnet()` embed the download directory in a
command string that rtorrent parses itself, wrapping the raw path in
literal quotes without escaping it. A path containing `"` therefore ends
the argument early and the directory is never applied — the torrent is
added with an empty save path and fails to save its data. A path
containing `\` fails the same way. The issue has the full analysis with
permalinks.

The escaping already exists in this codebase:
`XMLRPCProxy::rebuildSafeLoadParam()` escapes `\` and `"` before quoting
a command argument, pinned by the contract fixture entry "quotes and
backslashes in a value are escaped, not dropped". This adds
`rTorrent::quoteCommandArg()` doing the same, and routes both call sites
through it, which also removes the duplicated construction between the
two.

The replacement is a single `str_replace()` pass rather than two
sequential ones: escaping the quote first and the backslash afterwards
would double the backslash the quote escaping just introduced, leaving a
live quote behind it and reopening the same defect.

`tests/php/RtorrentCommandArgTest.php` covers the quote, the backslash,
that ordering trap, a comma staying inside the argument, and guards both
call sites against the raw concatenation returning. Full PHP suite:
1402 passed, 0 failed.

Not fixed here: `rtAddTorrent()` in `plugins/datadir/util_rt.php` carries
the same concatenation, but it has no callers anywhere in the
repository, so it seemed out of scope.

### DIFF
--- a/php/rtorrent.php
+++ b/php/rtorrent.php
@@ -236,9 +236,11 @@ class rTorrent
 	 */
 	static public function quoteCommandArg($value)
 	{
-		// One str_replace() pass, not two: escaping the quote first and the
-		// backslash afterwards would double the backslash the quote escaping
-		// just added, leaving a live quote behind it.
+		// The order of the pairs is what makes this correct, and it must not be
+		// changed: str_replace() works through the arrays left to right and can
+		// revisit a value an earlier pair inserted. Escaping the quote first
+		// would double the backslash that escaping just added, leaving a live
+		// quote behind it -- so the backslash pair has to come first.
 		return('"'.str_replace(array('\\', '"'), array('\\\\', '\\"'), $value).'"');
 	}
 
--- a/php/rtorrent.php
+++ b/php/rtorrent.php
@@ -72,7 +72,7 @@ class rTorrent
 				if(!rTorrentSettings::get()->correctDirectory($directory))
 					return(false);
 				$req->addCommand( new rXMLRPCCommand( 'execute', array('mkdir','-p',$directory) ) );
-				$cmd->addParameter( ($isAddPath ? getCmd("d.set_directory=")."\"" : getCmd("d.set_directory_base=")."\"").$directory."\"" );
+				$cmd->addParameter( getCmd($isAddPath ? "d.set_directory=" : "d.set_directory_base=").self::quoteCommandArg($directory) );
 			}
 			$comment = $torrent->comment();
 			if($comment)
@@ -124,7 +124,7 @@ class rTorrent
 				{
 					if(!rTorrentSettings::get()->correctDirectory($directory))
 						return(false);
-					$cmd->addParameter( ($isAddPath ? getCmd("d.set_directory=")."\"" : getCmd("d.set_directory_base=")."\"").$directory."\"" );
+					$cmd->addParameter( getCmd($isAddPath ? "d.set_directory=" : "d.set_directory_base=").self::quoteCommandArg($directory) );
 					$req->addCommand( new rXMLRPCCommand( 'execute', array('mkdir','-p',$directory) ) );
 				}
 				if($label && (strlen($label)>0))
@@ -219,6 +219,27 @@ class rTorrent
 			return($torrent);
 		}
 		return(false);
+	}
+
+	/**
+	 * Quote a value for use inside an rtorrent command string.
+	 *
+	 * The commands handed to load/load_raw are not typed parameters: rtorrent
+	 * parses them itself (src/rpc/parse.cc), ending a quoted argument at the
+	 * first unescaped '"' and reading '\' as an escape. A value wrapped in
+	 * quotes without escaping therefore ends the argument early -- a download
+	 * directory holding a quote silently loses everything from that quote on,
+	 * and the command around it stops parsing.
+	 *
+	 * Mirrors the escaping XMLRPCProxy::rebuildSafeLoadParam() applies to the
+	 * command strings it rebuilds.
+	 */
+	static public function quoteCommandArg($value)
+	{
+		// One str_replace() pass, not two: escaping the quote first and the
+		// backslash afterwards would double the backslash the quote escaping
+		// just added, leaving a live quote behind it.
+		return('"'.str_replace(array('\\', '"'), array('\\\\', '\\"'), $value).'"');
 	}
 
 	static protected function parseDirectory($directory, $isAddPath)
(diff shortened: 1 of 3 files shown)


xirvik:
Two changes needed, I'll let claude explain

 1. The comment says one str_replace() pass rather than two, but the array form is sequential — PHP's manual warns it "might replace a previously inserted value". What makes this correct is the order within the arrays: backslash first, quote second. Reversed, \" becomes \\\\" — an escaped backslash and a live quote. Please reword the comment (and the test name)  to say the order is what matters, otherwise a future cleanup may reorder the array believing it's safe.
 2. testNoCallSiteConcatenatesTheRawDirectory greps the source for the old shape, so renaming $directory or using single quotes reintroduces the bug with the test still passing. Since rXMLRPCRequest is already stubbed elsewhere in the suite, could this assert on the parameter sendTorrent() actually emits for a path containing a quote?
 
  rtAddTorrent() having no callers is a fair reason to leave it, though it's a one-liner if you'd rather not leave the pattern in the tree.

noctuum:
Thanks for the review. Fixed both.

You're right that the array form is sequential; I had the reasoning backwards. The comment now says the order of the pairs is what makes it safe, and the test name says the same. Swapping them does fail the tests, so a reorder can't land quietly.

The source grep is gone. The stub in XMLRPCProxyTest wasn't reusable here, but run() hands its commands to the settings object before it opens a socket, so a subclass over the singleton captures what sendTorrent() really emits — the same reflection trick the existing tests use. Both directory branches are covered now, plus the mkdir argument that has to stay unquoted.

Took the rtAddTorrent offer as well, so the pattern is out of the tree.

PR #3198, opened 2026-08-27
loginmgr: distinguish unavailable responses from expired sessions

IvanShift:
`KinozalTVAccount::isOK()` decides a session is still alive by *failing* to find
the guest markers in `$client->results`: no ` name="password"` and no
`/signup.php`. An HTTP 5xx page, a redirect Snoopy did not follow, and an empty
body contain neither marker either — so all three were read as a live,
authenticated session.

`isOK()` is also what `commonAccount::isOKPostFetch()` applies to a fetched
result, so the same blind spot accepted a failed download as a successful one:
loginmgr handed the empty body back to the caller as the downloaded page or
torrent and kept the cached cookies, so it never re-authenticated and every
following fetch failed the same way.

This requires a 200 and a non-empty string body before looking for the markers.

Two consequences worth naming up front:

* `status !== 200` is stricter than the `200..399` the framework applies around
  this call, so a 30x that Snoopy did not follow now forces a re-login for this
  tracker rather than being accepted as a live page. That is the intent, but it
  is a behaviour change on Kinozal.
* `login()` now returns `$this->isOK($client)` instead of a bare `true`. This
  changes no outcome — `commonAccount::fetch()` and `::check()` already apply
  `status >= 200 && status < 400` together with `isOK()` to that same response —
  it only makes `login()` honest about the result it reports. The surrounding
  indentation in `login()` is normalised to tabs while the return is changed.

### How to verify

`tests/plugins/loginmgr/KinozalAccountTest.php` covers the three answers that
used to pass (500, 302, empty body) and pins `login()`'s verdict on both a
`takelogin.php` reply that comes back as the guest form and one that comes back
authenticated.

    cd tests && bash php-test.sh

### DIFF
--- a/plugins/loginmgr/accounts.php
+++ b/plugins/loginmgr/accounts.php
@@ -136,26 +136,112 @@ abstract class commonAccount
 		return(true);
 	}
 
+	// Three answers, not two. They call for three different repairs, and this
+	// class used to make only two of them:
+	//
+	//   ANSWER_LIVE   the tracker answered from behind the login wall.
+	//   ANSWER_GUEST  the tracker answered, and answered as a guest. The
+	//                 session is dead and logging in again is the repair.
+	//   ANSWER_NONE   nothing usable arrived. This says NOTHING about the
+	//                 session, so logging in again is not a repair: it spends
+	//                 a credential POST on a tracker that is already failing
+	//                 and then throws away cookies never shown to be stale.
+	//
+	// Every isOK() under accounts/ tells the first two apart by looking for a
+	// guest marker -- a login form, a registration link, a "you must log in"
+	// line -- and none of them can answer the third, because a page that never
+	// arrived carries no marker either. That is why all three used to collapse
+	// into "the session is live", and why a 5xx error page went back to the
+	// caller as the page or the torrent it had asked for.
+	const ANSWER_NONE = 0;
+	const ANSWER_GUEST = 1;
+	const ANSWER_LIVE = 2;
+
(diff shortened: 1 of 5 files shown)


xirvik:
@IvanShift this only fixes Kinozal, but looks like the bug affects everybody (minus MyAnonamouse and  YggTorrent) . What do you think about a non-empty body guard live in commonAccount::isOKPostFetch(), so every tracker gets it, with those two overriding?

IvanShift:
Thanks, agreed. I moved the response guard into `commonAccount::isOKPostFetch()`, so it now applies to all accounts by default instead of being Kinozal-specific.

The base implementation distinguishes three outcomes:

- a valid authenticated response;
- an explicit guest/login-wall response;
- no usable response because the request failed or returned an indeterminate result.

Only an explicit guest response triggers re-login. A transport failure, empty response, unfinished redirect, or server error now returns failure while preserving the cached session, so a temporary outage cannot cause repeated credential POSTs or delete valid cookies. `304 Not Modified` remains valid for authenticated RSS requests.

Kinozal now keeps only its tracker-specific payload check. LostFilm’s `isOKPostFetch()` override preserves the base verdict while handling its recovery request; the other account-specific `isOK()` overrides continue to work behind the common guard.

The branch is rebased onto current `master` as one commit. The full PHP test harness passes on PHP 7.4, 8.1, and 8.5; the focused account suites, lint checks, and adversarial mutations are also green.

Thanks for pointing this out.

PR #3200, opened 2026-08-27
Up/small UI

IvanShift:
Three unrelated fixes in the torrent list and the details pane, kept as three separate commits so any subset can be taken on its own. They touch different regions of different functions — the two `js/webui.js` hunks sit ~880 lines apart — so cherry-picking one without the others applies cleanly.

## 1. Keep the peer table when a peer response is for another torrent

`addPeers()` ended with `if (this.dID == hash) table.updateRows(...) else table.clearRows()`, so a `getpeers` answer that is **not** for the torrent whose details are open blanked the peer table that *is* open. Open details on torrent A, switch to B, and A's in-flight response arrives after B's rows were drawn: B's peer list goes empty until the next poll paints it again.

Clearing is not this function's decision to make. `getPeers()` already clears the table before a fresh open, and `clearDetails()` clears it when the details block closes or the selected torrent disappears — so dropping the `else` branch leaves every path that *should* empty the table still doing so.

The same function also wrote `this.peers[hash]` for any hash at all, while the poll's cleanup loop only walks `this.torrents`. A response for a torrent that was removed mid-poll therefore parked its peer list in the cache with nothing left to release it, for the life of the page. It now returns early for a hash that is no longer in the list.

## 2. Give each joined tracker message its own line

`d.message` belongs to the download, not to one tracker row. When more than one tracker row is failing, rTorrent joins their messages into that single string with `" /// "` between them, so the General tab renders

```
Tracker: [Could not connect to server /// Could not resolve hostname]
```

on one line, which reads as one garbled sentence rather than one answer per tracker. This is observed behaviour rather than a citation into rTorrent's source, but it is not a one-off: on a live 400-torrent instance, sampled repeatedly, **every** non-empty `d.message` was such a join — between 36 and 122 torrents at a time depending on where the announce cycle stood, always exactly two parts, and never once a message without the separator. The two distinct texts seen were `Tracker: [Could not resolve hostname /// Could not resolve hostname]` and `Tracker: [Could not connect to server /// Could not resolve hostname]`.

`getClickableTrackerStatus()` now replaces that separator with `<br>` **after** the links have been built. Both the order and the surrounding spaces matter:

- a URL cannot contain a space, so `" /// "` can never fall inside a link that was just built;
- a path such as `https://tracker.example/a///b` is left whole;
- a bare `a///b` with nothing around it is not a join, and is left alone too.

## 3. Let a column show its untruncated value as a cell tooltip

The torrent table is `table-layout:fixed` with `white-space:nowrap` and `overflow:hidden`, so a value wider than its column is cut at the column edge with no ellipsis and no other sign that there is more to read. A save path longer than the 200px Save-path column is simply gone, and hovering the cell shows the *row* tooltip — the torrent's name — rather than the path.

Wrapping is not available here: the table body is virtualized and derives every row's height from the first one (`TR_HEIGHT`), so a single taller row desynchronizes the scroll for the whole list. A tooltip is what is left.

A column definition can now carry `titled: true`:

- `createRow()` sets the cell's own `title`;
- `syncDOM()` keeps it current — rows are built once and refreshed in place on every poll, so a title written only at creation would freeze at whatever the value was when the row first appeared;
- the title is written only when there is something to show, and removed when the value goes empty, because `title=""` on a cell shadows the row tooltip while displaying nothing in its place.

Two columns opt in: **Save path**, whose values are unbounded, and the **tracker status** column of the `trackerstatus` plugin, where `d.message` is routinely far longer than its 200px — on the instance above every one of those messages ran 60–69 characters, roughly two and a half times what a 200px column shows. Every other column is untouched.

## How to verify

Each commit carries its own cases, driving the real functions against jsdom rather than asserting on a reimplementation:

- `tests/js/webui-stale-details.spec.js` — a response for an unknown hash touches neither the cache nor the table; one for the shown torrent updates both; one for a known but unselected torrent updates the cache and leaves the table alone.
- `tests/js/trackerstatus.spec.js` — the join is split, a URL containing slashes survives, a bare `///` is left alone.
- `tests/js/stable-cell-tooltip.spec.js` (new) — drives the real `createRow()` and `syncDOM()` and is column-agnostic: an opted-in column gets the full value as its cell title, a column that did not opt in keeps no title of its own and the row tooltip still shows through, an update moves the title with the value, and a value going empty removes the attribute rather than leaving `title=""`.

```
cd tests && npx jest
```

**193 tests in 19 suites** on this branch, against **184 in 18** on `master` — the 9 added cases above, no existing case changed.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -263,13 +263,21 @@ function getClickableTrackerStatus(text)
 {
 	if (!text) return '';
 	var sanitized = escapeHTML(text);
-	return sanitized.replace(/https?:\/\/[^\s<>"']+/gi, function(match) {
+	var linked = sanitized.replace(/https?:\/\/[^\s<>"']+/gi, function(match) {
 		var entityMatch = match.match(/&(?:quot|#039|#39|lt|gt);/i);
 		var truncated = entityMatch ? match.substring(0, entityMatch.index) : match;
 		var cleanUrl = truncated.replace(/[.,)!?:;\\\]}]+$/g, '');
 		var trailing = match.substring(cleanUrl.length);
 		return '<a href="' + cleanUrl + '" target="_blank" rel="noopener noreferrer">' + cleanUrl + '</a>' + trailing;
 	});
+	// d.message belongs to the whole download, and rTorrent JOINS the message
+	// of every failing tracker row into it with ' /// ' between them -- e.g.
+	// "Tracker: [Could not connect to server /// Could not resolve hostname]"
+	// from two different rows. Run together on one line that reads as a single
+	// garbled sentence, so each row's answer gets its own line. Done AFTER the
+	// linkification: a URL cannot contain a space, so this separator can never
+	// fall inside one of the links just built.
+	return linked.replace(/ \/\/\/ /g, '<br>');
 }
 
 
--- a/js/stable.js
+++ b/js/stable.js
@@ -981,6 +981,17 @@ dxSTable.prototype.createRow = function(cols, sId, icon, attr) {
 			td.append(
 				$("<div>").text(celldata || " "),
 			);
+			// The table is table-layout:fixed with white-space:nowrap and
+			// overflow:hidden, so a value wider than its column is cut off at
+			// the edge with no ellipsis and no other sign that anything is
+			// missing. Wrapping instead is not an option: the body is
+			// virtualized and derives every row's height from the first one
+			// (TR_HEIGHT), so one taller row desynchronizes the scroll for the
+			// whole list. A column that opts in therefore carries the full
+			// value as the cell's own tooltip.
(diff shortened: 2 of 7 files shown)


xirvik:
One correction on the second fix. The separator isn't rTorrent joining multiple tracker rows — it's libtorrent, 

src/tracker/tracker_http.cc:502
 m_slot_failure(msg + " /// " + m_last_error_message)

 in receive_failed(), reached when send_next_family() is exhausted and the other family already failed. So the two parts are IPv4 and IPv6 for the same tracker, which is why you always saw exactly two and never a bare message. Rendering them on separate lines is still right IMHO. I was going go ask for to edit the comment but I'll do it myself to save a roundtrip for a trivial thing.

PR #3202, opened 2026-08-28
filedrop: add paste-to-add for magnet links and torrent URLs

lebowitz:
## Summary

Pasting a magnet link or torrent-file URL anywhere on the page (outside a
text field) now adds it immediately — no dialog, no explicit "Add torrent"
step. Extends `filedrop` rather than adding a new plugin: its drop handler
already does exactly what paste needs (extract magnet/URL text, chunk it,
rate-limit calls to `addtorrent.php`), so paste reuses that logic instead
of a parallel implementation.

## Why filedrop and not a standalone plugin

I considered a separate plugin first. But `filedrop`'s existing drop
handler already extracts `magnet:`/`http(s)://` URLs from dropped text,
chunks them via `queuefiles`, and rate-limits the calls to
`addtorrent.php` (200ms between chunks). A standalone plugin would either
duplicate that or skip the chunking protection — so a multi-magnet paste
could fire dozens of simultaneous requests. `filedrop`'s name is
drop-specific, but the underlying job (add a torrent from a gesture
anywhere on the page) is the same one paste needs, so I folded it in
rather than reimplementing it next to it.

## Safety scoping

- Skips `INPUT`, `TEXTAREA`, and `contenteditable` targets, so normal
  paste into any existing field (search, rename, settings, the
  `bulk_magnet` dialog) is untouched.
- Only intercepts (`preventDefault`) when the pasted text actually
  contains a `magnet:` or `http(s)://` URL; otherwise it's a no-op and
  default paste behavior proceeds normally.
- Known tradeoff: since this is a document-wide listener, a magnet
  unrelated to ruTorrent that happens to be on your clipboard, pasted
  while focus is on a non-editable part of the page, will get added. This
  is the same page-wide-listener tradeoff `filedrop` already accepts for
  drag-and-drop.

## Testing

- Full Jest suite passes (202/202), including 9 new tests covering: paste
  ignored in editable/contenteditable targets, paste ignored when no
  magnet/URL is present, single- and multi-line paste, the chunking/rate-limit
  behavior shared with drop, and failure-path `noty` handling.
- `php -l`, eslint, and editorconfig-checker all clean on the changed files.
- Ran end-to-end against a real rTorrent 0.10.0 + ruTorrent instance on my
  own seedbox: confirmed the paste correctly calls `addtorrent.php`, the
  editable-field guard holds under real interaction, and a multi-magnet
  paste adds every link from one paste.

Built and tested with AI assistance (Claude Code) — I reviewed the diff
and ran it myself against my own instance before opening this, per the
contributing guidelines.

### DIFF
--- a/plugins/filedrop/init.js
+++ b/plugins/filedrop/init.js
@@ -3,8 +3,89 @@ if(window.FileReader)
 else
 	plugin.disable();
 
+plugin.isEditableTarget = function(el)
+{
+	if (!el)
+		return false;
+	const tag = el.tagName;
+	return (tag == 'INPUT') || (tag == 'TEXTAREA') || el.isContentEditable;
+}
+
+plugin.addUrls = async function(urls)
+{
+	// break URL list into chunks based on plugin configuration
+	const urlChunks = [];
+	if (plugin.queuefiles) {
+		for (let i = 0; i < urls.length; i += plugin.queuefiles)
+			urlChunks.push(urls.slice(i, i + plugin.queuefiles));
+	}
+	else {
+		if (plugin.maxfiles && urls.length > plugin.maxfiles) {
+			noty(theUILang.tooManyFiles + plugin.maxfiles, "error");
+			return;
+		}
+		urlChunks.push(urls);  // all URLs in one chunk
+	}
+
+	// send API calls to rTorrent one chunk at a time with
+	// delay between chunks
+	for (const [i, urls] of urlChunks.entries()) {
+		if (i != 0)
+			await new Promise(r => setTimeout(r, 200));
+
+		await Promise.all(urls.map(async url => {
+			const data = await $.ajax({
+				url: plugin.path + '../../php/addtorrent.php',
+				method: "POST",
+				data: { url, json: 1 },
+				dataType: "json",
+			}).catch((_xhr, status, error) => {
+				console.error(
+					"filedrop: rTorrent API call error: " +
+					"url = %s | response status = %s | error = %o",
+					url, status, error);
+				return {};
+			});
+
+			const result = (typeof data === "object")
+				? (data.result ?? "Failed")
+				: "Failed";
+			noty(
+				`${url} : ${theUILang['addTorrent' + result]}`,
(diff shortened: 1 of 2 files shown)


xirvik:
I've been looking into this a bit (I now is a draft).

I ran the handler against normal clipboard content:
 
  "https://example.com/news/article-about-cats"  → paste swallowed, addtorrent.php called
  "look at this https://example.com/page and…"   → paste swallowed, addtorrent.php called
  "just some copied words"                       → no-op, correct
  
The guard itself is fine, it's the URL rule that's too wide. It came from drop, where it makes sense — dragging a link is deliberate. Ctrl+V isn't, and most of the page isn't a text field, so text copied out of a chat or a forum post ends up as a POST and an error toast.
  
 Can you only intercept when the whole paste is URLs? Every token has to match, otherwise let the paste through. Deliberate link pasting keeps working and the copied-prose case goes away.

 Don't filter on .torrent though — none of the trackers loginmgr supports use it:
  
 RUTracker   /forum/dl.php?t=
 Kinozal     /download.php?id=
 LostFilm    /download.php
 YggTorrent  /engine/download_torrent?id=
 NNMClub     /forum/download.php?id=
 
 That would break the links people actually paste.
 
  Add a test for whichever way it lands — a non-torrent URL being added or not is a decision now, not an accident. The rest is covered already.

lebowitz:
Pushed two commits addressing the review:

Narrow the URL rule to whole-paste-only. handlePaste now requires every token in the pasted text to look like a magnet/URL -- if any token doesn't match, the paste passes through untouched. Fixes the "prose with an embedded link" false positive while deliberate multi-link pastes still work. Drop is unchanged, since dragging a link is a deliberate act.

Also split comma/semicolon-separated magnet lists (magnet:aaa,magnet:bbb), which some export tools produce and which failed the new whole-paste check -- split only fires on a comma/semicolon immediately followed by another "magnet:", so a normal link's comma-joined tracker list inside tr= stays intact (regression test included).

5 new/updated tests, full suite still green (207/207).

PR #3204, opened 2026-08-28
mobile plugin: localize hard-coded strings

koblack:


xirvik (changes requested):
init.js now reads four keys, but only three were added to the lang files:

  $('#openStatus td:first').text(theUILang.Open_label);   // ← Open_label defined nowhere

koblack:
The '"Open_label" key is already defined by "lang/en.js"

https://github.com/Novik/ruTorrent/blob/lang/en.js#L163

xirvik:
> The '"Open_label" key is already defined by "lang/en.js"
> 
Ah yes, sorry. Bad grep.

PR #3209, opened 2026-08-28
httprpc: keep client-quoted load.* arguments instead of dropping them

xombiemp:
## Summary

The XMLRPC proxy sanitizer (#3181) rebuilds `load.*` command parameters and dropped any argument that started with `"`, on the assumption that clients send values raw and this side quotes them.

That does not match what rtorrent clients actually send. **cross-seed** injects with:

```
load.raw_start
  d.directory_base.set="/data/.../cross-seed/PassThePopcorn"
  d.custom1.set="cross-seed"
  d.custom.set=addtime,<timestamp>
```

Those quoted setters were stripped. The torrent was still added (`load.raw_start`), but into the default download directory with no label. Hardlinked files were already in the cross-seed path, so the client sat in **Pausing** at 0%. Unquoted clients (Sonarr/Radarr `d.custom1.set=tv/sonarr`) were unaffected.

This change unquotes a client-quoted argument (commas inside quotes stay one argument), then re-quotes it the same way as an unquoted value. A leading `$` is still dropped after unquoting, so `d.custom1.set="$execute.capture=..."` cannot sneak through.

## Test plan

- [x] `tests/php-test.sh` (full PHP suite)
- [ ] Inject a cross-seed match through httprpc: torrent should get label `cross-seed` and `directory_base` under the link dir, then start/seed
- [ ] Confirm a Sonarr/Radarr add with an unquoted label still sets the label
- [ ] Confirm `d.custom1.set="$execute.capture=/bin/hostname"` is still stripped

### DIFF
--- a/php/xmlrpc_proxy.php
+++ b/php/xmlrpc_proxy.php
@@ -592,6 +592,77 @@ class XMLRPCProxy
 	}
 
 	/**
+	 * Split command arguments the way rtorrent does: commas separate them, and
+	 * a double-quoted string is one argument even when it contains commas.
+	 * Unquoted arguments are trimmed; quoted ones are not. An unclosed quote,
+	 * or text after a quoted argument that is not a comma, is malformed and
+	 * returns null so the whole parameter is dropped.
+	 *
+	 * Clients such as cross-seed quote every value (d.custom1.set="cross-seed").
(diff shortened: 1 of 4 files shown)


xirvik (approved):
Found a couple of things it doesn't handle but it does fix some cases. I'll send a follow-up PR to save a round-trip.

PR #3224, opened 2026-08-30
Restore PHP 7.4 Torrent metadata compatibility

IvanShift:
## Summary

ruTorrent documents PHP 7.4 or newer as supported, but the native `mixed`
property declarations in `Torrent` require PHP 8.0 and prevent the file from
being parsed on PHP 7.4.

- Replace the seven native `mixed` declarations with PHPDoc-annotated public
  properties, preserving declared-property behavior and null defaults.
- Normalize metadata keys to strings before `setMeta()` dispatch. This prevents
  numeric bencode dictionary keys from matching named switch cases through
  PHP 7.4 loose comparison.
- Configure PHPStan for the documented PHP 7.4 floor.
- Run the PHP CI job on both PHP 7.4 and PHP 8.1.

## Verification

- Linted all 385 PHP/INC files on PHP 7.4 and PHP 8.1.
- Ran the complete 47-file PHP test harness on both runtimes.
- Ran PHPStan 2.2.9 with `phpVersion: 70400`: no errors.
- Verified the existing numeric-key dictionary regression on both runtimes.
- Confirmed that restoring native `mixed` breaks PHP 7.4 lint/PHPStan.
- Confirmed that removing key normalization breaks the numeric-key test on
  PHP 7.4 while PHP 8.1 remains green.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,11 +8,14 @@ on:
 jobs:
   php:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php: ['7.4', '8.1']
     steps:
       - uses: actions/checkout@v4
       - uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.1'
+          php-version: ${{ matrix.php }}
           extensions: posix, dom, xml, mbstring, iconv, json, curl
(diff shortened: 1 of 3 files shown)


xirvik:
Oops, thanks!

PR #3227, opened 2026-08-31
Make settings saves recover socket allocation failures

IvanShift:
## Summary

- Treat the five “Other Limiting” fields as numeric rTorrent settings.
- Treat a cleared field as “leave unchanged” while preserving an explicitly entered `0`.
- Keep values rejected by local socket preflight out of the client-side settings model.
- Snapshot active socket allocation bounds before direct XML-RPC writes and restore them if a settings batch fails.
- Re-read effective settings after definitive failures.
- Serialize Save, rollback, reconciliation, and deferred WebUI persistence so repeated saves cannot overlap.
- Keep indeterminate saves locked and show persistent localized recovery guidance.

## Why

rTorrent XML-RPC multicalls are not transactional. A later command can fail after earlier socket-bound writes have already succeeded, leaving partially changed or staged allocation values.

The previous browser flow could also allow another Save while rollback, settings reconciliation, or WebUI persistence was still running. In addition, a locally rejected socket value could enter the client-side model even though it was never sent to rTorrent, causing a later cleared field to reuse the rejected value and block an otherwise valid sibling change.

This PR restores the original socket bounds after definitive direct XML-RPC failures, reconciles the UI with the effective daemon state, and keeps the Save operation serialized through its terminal result. A timeout or status-zero result remains fail-closed because the client cannot know whether the remote write completed.

## Verification

- Focused settings review: 60/60 passed.
- Full Jest suite: 276/276 passed.
- ESLint, JavaScript syntax checks, and `git diff --check` passed.
- Direct XML-RPC success, failure, and restore were verified with disposable rTorrent 0.16.21.
- Enabled httprpc success and server-side rollback were verified in the same container runtime.
- The branch merges cleanly with current upstream `master`.
- Independent full-range review: APPROVED, no findings.

### DIFF
--- a/js/options.js
+++ b/js/options.js
@@ -190,7 +190,7 @@ const theOptionsWindow = {
 						$("<label>").attr({for: id}).text(label + ": "),
 					),
 					$("<div>").addClass("col-md-4").append(
-						$("<input>").attr({type: "number", id: id}).prop("maxlength", 6),
+						$("<input>").attr({type: "number", id: id}).addClass("num").prop("maxlength", 6),
 					),
 				)),
 				$("<div>").addClass("row").attr({id: "socket_alloc_budget_row"}).hide().append(
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -946,8 +946,8 @@ rTorrentStub.prototype.getResponse = function(data)
 					this.faultString.push("XMLRPC Error: "+message+" ["+this.action+"]");
 				}
 				else
-					for(var i=0; i<memberFaults.length; i++)
-						this.faultString.push("XMLRPC Error: "+memberFaults[i]+" ["+this.action+"]");
+					for(var j=0; j<memberFaults.length; j++)
+						this.faultString.push("XMLRPC Error: "+memberFaults[j]+" ["+this.action+"]");
 			}
 		}
 		if(!this.xmlRPCFault)
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -383,13 +383,7 @@ rTorrentStub.prototype.setsettings = function()
 		var prmType = "string";
 		var prm = this.vs[i];
 		if(this.ss[i].charAt(0)=='n')
-		{
 			prmType = "i8";
-			// Match the httprpc door: an empty numeric field is a valid zero,
-			// not an undecodable empty XML-RPC integer that rejects the batch.
-			if(prm==="")
-				prm = 0;
-		}
 		var cmd = null;
 		var socketAlloc = theRequestManager.getSocketAllocCategory(this.ss[i]);
 		if(this.ss[i]=="ndht")
--- a/js/rtorrent.js
+++ b/js/rtorrent.js
@@ -377,13 +377,19 @@ rTorrentStub.prototype.recheck = function()
 
 rTorrentStub.prototype.setsettings = function()
 {
-	var adjustAlloc = false;
+	var socketCategories = [];
 	for(var i=0; i<this.vs.length; i++)
 	{
 		var prmType = "string";
-		if(this.ss[i].charAt(0)=='n')
-			prmType = "i8";
 		var prm = this.vs[i];
+		if(this.ss[i].charAt(0)=='n')
+		{
+			prmType = "i8";
+			// Match the httprpc door: an empty numeric field is a valid zero,
+			// not an undecodable empty XML-RPC integer that rejects the batch.
+			if(prm==="")
+				prm = 0;
+		}
 		var cmd = null;
 		var socketAlloc = theRequestManager.getSocketAllocCategory(this.ss[i]);
 		if(this.ss[i]=="ndht")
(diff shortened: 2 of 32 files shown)


xirvik:
Claude review, but I did review the review :-) 
I agree with the hardcoded message obviously, this needs to go in the theUILang .

About replacing blanks silently... not 100% sure, but I soft-agree as well.


**An empty numeric field now saves `0` for five settings.** Adding `.num` in `options.js` makes `max_uploads_global`, `max_downloads_global`, `max_memory_usage`, `max_open_files` and `max_open_http` serialise as `i8`, and `setsettings` coerces `""` to `0`. Before, an empty field went as a string, rTorrent rejected the batch, and nothing changed. `0` is unlimited for the slot counts, so those are fine — but a user who clears **max open files** and presses Save now gets `0` applied silently, where previously the save failed loudly. Is `0` the intended meaning for each of the five, or should an empty field be treated as "leave this setting alone"? Failing is unhelpful, but applying a value the user did not type is worse.

**The indeterminate lock has no way out except a page reload.** `finishSettingsSaveIndeterminate()` leaves `settingsSavePending` set and the buttons disabled, and the only instruction is a `noty` that disappears. Closing and reopening the dialog does not clear it, since the flag lives on `theWebUI`. Locking is the right call — I would not weaken it — but the recovery needs to survive longer than a toast: a persistent line in the settings dialog, or re-enabling with a confirmation, so a network blip does not leave someone with a dead Save button and no memory of why.

**That message is hardcoded English.** `"Settings outcome is unknown. Save remains locked; reload manually only after rTorrent responds."` should be a `theUILang` key. There are three raw strings in the core JS already, but they are developer diagnostics — a deprecation warning, a JS error, a bad-response dump. This one is operational guidance a non-English user will hit during a normal failure, and `tests/js/lang.spec.js` cannot see it because it only checks keys that exist.

xirvik (changes requested):
I think Claude is right on this one.

Requesting changes on one point: the empty-numeric-field coercion creates the
partial write this PR exists to recover from.

Adding `.num` in `options.js` makes `max_open_files` and `max_open_http`
serialise as `i8`, and `setsettings` turns a cleared field into `0`. For the
socket-allocation categories the two bound writes are then accepted and only
`adjust_alloc` fails, so rTorrent is left with staged bounds it will not use.

Reproduced against rtorrent 0.16.20 over SCGI, sending exactly what the UI
sends for a cleared "max open files" field.

Baseline:

    network.max_open_files              -> 10240
    system.sockets.files.min_alloc      -> 10240
    system.sockets.files.max_alloc      -> 1000000

master today (cleared field goes as an empty string):

    system.sockets.files.min_alloc.set "" ""   -> FAULT: Not a value.

    network.max_open_files              -> 10240
    system.sockets.files.min_alloc      -> 10240
    system.sockets.files.max_alloc      -> 1000000

Nothing is staged: the first command is refused and the batch stops.

This branch (cleared field goes as 0):

    system.sockets.files.min_alloc.set "" 0    -> 0
    system.sockets.files.max_alloc.set "" 0    -> 0
    system.sockets.adjust_alloc                -> FAULT: adjust_allocation:
                                                  allocation below the category
                                                  minimum : 0 < 4

    network.max_open_files              -> 10240
    system.sockets.files.min_alloc      -> 0
    system.sockets.files.max_alloc      -> 0

Both bound writes land. The recompute is what fails, so the effective value is
still 10240 while the bounds behind it are 0/0.

The rollback does clean this up -- replaying the snapshot restore
(`min_alloc.set 10240`, `max_alloc.set 1000000`, `adjust_alloc` -> 0) put the
daemon back exactly as it was. So with this branch the outcome of clearing the
box is the same as master's, reached through a write, a fault, a restore and a
read-back rather than a single refusal. When that round trip does not complete
-- the timeout case this PR itself treats as indeterminate -- the daemon keeps
0/0 staged and Save stays locked.

Suggested change: for the socket-allocation fields, treat an empty box as
"leave this setting alone" and omit it from the batch, rather than sending 0.
That removes the partial write instead of recovering from it. `0` stays correct
for `max_uploads_global` and `max_downloads_global`, where it means unlimited.

Note this is 0.16.20 -- the write path is live from 0x1013 but the budget probe
only runs from 0x1015, so there is no pre-check to catch the 0 before it is
sent. That is the configuration where it matters.

Two things not checked: whether the `min_alloc`/`max_alloc` reads this branch
prepends exist on 0.16.19 (the gate is 0x1013; I only have a 0.16.20 daemon),
and anything Windows-specific.

IvanShift:
Thanks for the review. I used Codex Sol Ultra instead of Claude ( waiting for limits reset)
I also changed PR-description.


 I addressed all three points:
- Clearing any of the five rTorrent numeric limits now means “leave this setting unchanged”; an explicitly entered `0` is still sent as zero.
- An indeterminate Save now leaves persistent recovery guidance in the Settings dialog, using a `theUILang` key available in every core language.
- A full branch re-review found one related state bug: a socket value rejected by local preflight was being written into the client-side settings model. That rejected value could then be reused when the field was cleared and block a valid sibling change. Rejected rTorrent values now remain pending and enter the model only after preflight succeeds and the request is actually sent.

The new regressions fail on the previous code and pass with the fixes.

PR #3240, opened 2026-09-01
erasedata: queue erase requests so mass erases cannot flood rtorrent

adenwuts:
Introduces a queue to erasedata alongside an upper bound of failed attempts, so that large amounts of erasure requests don't flood the rtorrent RPC endpoint and leave it unresponsive to everything else.

Previously, large quantities of erasure requests at the erasedata plugin would trigger a substantial number of simultaneous RPC calls, which could (and would) hit RPC timeouts. This would leave the erasures as failed, and they would be re-attempted ad nauseam, continuing to hit the RPC timeout, and causing lockups.

This is fixed by the introduction of a queue, which acts as a shim between the erasure requests and the RPC calls. Each request records a marker file (which can't fail due to a busy rtorrent), and exactly one of the firing processes then drains the whole queue at rtorrent's pace. Success is defined by the .list file the pre-existing garbage collector consumes; a request that keeps failing is retried on later sweeps and abandoned after $erasePendingMaxAttempts, leaving the torrent and data in place.

This keeps the existing (and unfortunate) pattern where a torrent disappears from ruTorrent/rtorrent before its data is deleted: d.erase is only sent after the .list file is recorded, and the garbage collector deletes the data on its next pass (up to $garbageCheckInterval later). Failures before that handoff leave the torrent and its data in place to be retried, but failures after it (a GC unlink error, a lost .list) can still orphan data with the torrent already gone. This patch doesn't introduce that issue, but also doesn't fix it.

### DIFF
--- a/plugins/erasedata/conf.php
+++ b/plugins/erasedata/conf.php
@@ -6,6 +6,12 @@
 
 	$erasedebug_enabled = false;
 
+  // The maximum amount of times a download is attempted to be marked for erasure before it is skipped,
+  // Setting a limit here prevents a problematic case being retried indefinitely and blocking other RPC calls.
+  //
+  // 0 disables the limit and will retry indefinitely.
+	$erasePendingMaxAttempts = 10;
+
 	// Replaces "remove" option in torrent menu to always delete with data
 	// Refrains from showing other options on web interface
 	$replaceRemoveTorrent = false;
--- a/plugins/erasedata/erase.php
+++ b/plugins/erasedata/erase.php
@@ -1,8 +1,8 @@
 <?php
 
 // CLI entry point for "erase this download and delete its data", for callers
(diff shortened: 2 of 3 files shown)


xirvik:
@adenwuts Seems to work OK but can you add some unit tests there? I'm trying to get into the habit of having in PR come with some...

adenwuts:
Hi @xirvik,

I've been testing and iterating locally at the integration/e2e layer for this patch in particular given its a solution to a load problem rather than any internal that I want to pin future developers to. I'm happy to try and merge a similar work-flow into ruTorrent's CI, but until that exists I worry about enforcing any of this via unit-tests.

The most valuable test I'd want from this is "does this fix erasedata at load", which is not something our current testing infrastructure can answer.

Issue #3077, opened 2026-06-29
Material Design: Torrent_URL_multiline	is almost invisible


ChriZathens:
### Please complete the following tasks.

- [x] Web browser cache cleared
- [x] Link provided to install script if applicable
- [x] Not using broken rtinst install script
- [x] Web browser, ruTorrent, PHP and OS version provided

### Tell us about your environment

Chrome, rutorrent 5.3.4

### Tell us how you installed ruTorrent

Git Pull

### Describe the bug

As per the title:
The "explanation" text regarding Torrent_URL_multiline (Torrent URLs (one per line)) is almost invisible when using Material Design Theme.

xirvik:
@ChriZathens can you add a screenshot please?

xirvik:
Closing due to no answer.

ChriZathens:
Apologies for not responding earlier, my whole installation was screwed, then vacation came... But finally back and running 
Running 5.3.12 version ATM (latest git)
The problem is not only with the aformentioned string, but also with the "Directory" string.
The only browser where those are visible is Firefox:

<img width="600" height="472" alt="Image" src="https://github.com/user-attachments/assets/a06d9255-fcb8-4ada-8eb5-f808a6018a5e" />

On Edge and Chrome those strings indicated by the red arrow are almost invisible
Edge:

<img width="596" height="472" alt="Image" src="https://github.com/user-attachments/assets/86fb1945-a253-4189-b4f3-f76144054d65" />

Chrome:

<img width="604" height="471" alt="Image" src="https://github.com/user-attachments/assets/ffe070ce-31cf-4603-9355-e81a2403e892" />

xirvik:
@ChriZathens Does this solve the problem for you? https://github.com/Novik/ruTorrent/pull/3244

PR #3244, opened 2026-09-01
Material Design: make placeholder text visible

xirvik:
The add-torrent dialog's "Torrent URLs (one per line)" and "Directory" hints are unreadable in Chrome and Edge (#3077).

Both strings are placeholders rather than labels -- js/content.js sets them on the #url textarea and the #dir_edit input -- and nothing in this theme styled ::placeholder. Bootstrap's rule is scoped to .form-control and those two fields carry flex-grow-1, so no author rule applied at all and the colour came from the browser. Chrome and Edge use #757575; this theme paints the fields #6F6F6F. Measured in Chrome, that is a contrast ratio of 1.09:1 -- the hint is the same grey as the box it sits in.

Firefox derives its placeholder colour from the element's own colour, which the rules above set to #fff, so it stayed readable there. That is why the report named only the other two browsers.

White at .8 gives 3.88:1 against the same background. It still falls short of the 4.5:1 WCAG AA wants for normal text -- #6F6F6F is light enough that only pure white clears that -- but the hint is legible, and darkening the field is a wider change to this theme's palette than a bug fix should make.

The other themes are unaffected: Oblivion and OblivionBlue give these fields a white background with black text, and Dark and DarkBetter do not restyle them, so the browser default reads correctly in all of them.

### DIFF
--- a/plugins/theme/themes/MaterialDesign/style.css
+++ b/plugins/theme/themes/MaterialDesign/style.css
@@ -705,6 +705,22 @@ textarea {
 	border-radius: 2px;
 }
 
+/* Placeholder text. Nothing styled it before: Bootstrap's rule is scoped to
+   .form-control and these fields carry flex-grow-1 instead, so the colour came
+   from the browser. Chrome and Edge use #757575, which against the #6F6F6F
+   these fields are painted is a contrast ratio of 1.09:1 -- the hint is the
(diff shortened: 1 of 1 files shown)


ChriZathens (commented):
Thanks xirvik.
I have manually made these changes and i can confirm that both texts are now visible:

Image

PR #3245, opened 2026-09-01
check_port: add provider failover and Globalping

chiphead2332:
## Summary

Adds configurable provider failover to the `check_port` plugin and adds
Globalping as a third port-check provider.

The primary provider remains unchanged by default:

- IPv4: YouGetSignal
- IPv6: PortChecker

Default fallbacks are:

- IPv4: PortChecker → Globalping
- IPv6: Globalping

Provider implementations are moved out of `action.php` into separate files
and registered through a shared provider registry. The dispatcher is
provider-agnostic: providers return `0` (unknown), `1` (closed), or `2` (open),
and the configured chain continues until a definitive result is available.

Globalping is used as the final fallback when earlier providers fail or return
an inconclusive result.

## Testing

Tested on a live ruTorrent/rTorrent installation.

- Verified IPv4 and IPv6 port checks.
- Tested all three providers and provider rotation.
- Tested failover by forcing the primary provider to fail and confirming Globalping supplied the result.
- All changed PHP files pass `php -l`.
- Unchanged frontend and localization files remain identical to upstream.

### DIFF
--- a/plugins/check_port/action.php
+++ b/plugins/check_port/action.php
@@ -2,6 +2,7 @@
 require_once(dirname(__FILE__) . "/../../php/settings.php");
 require_once(dirname(__FILE__) . "/../../php/Snoopy.class.inc");
 require_once(dirname(__FILE__) . "/parse.php");
+require_once(dirname(__FILE__) . "/providers.php");
 
 // Load the plugin's configuration settings from conf.php
 eval(FileUtil::getPluginConf('check_port'));
@@ -10,6 +11,14 @@ eval(FileUtil::getPluginConf('check_port'));
 $currentCheckPortTimeout = isset($checkPortTimeout) ? (int)$checkPortTimeout : 15;
 $currentUseWebsiteIPv4 = isset($useWebsiteIPv4) ? $useWebsiteIPv4 : "yougetsignal";
 $currentUseWebsiteIPv6 = isset($useWebsiteIPv6) ? $useWebsiteIPv6 : "portchecker";
+
+$currentFailoverProvidersIPv4 = isset($failoverProvidersIPv4)
+	? $failoverProvidersIPv4
+	: array("portchecker", "globalping");
+
+$currentFailoverProvidersIPv6 = isset($failoverProvidersIPv6)
+	? $failoverProvidersIPv6
+	: array("globalping");
 
 /**
  * Gets the public IP address (IPv4 or IPv6) from ipify.org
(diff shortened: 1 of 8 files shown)


xirvik:
Tested it. TL;DR it reports all ports as open. 

Long explanation from Claude:

Globalping's rawOutput uses `tcp_conn=N` as a per-attempt counter, and it
appears on the "No reply" lines as well as the reply lines. The open test is
`preg_match('/tcp_conn=\d+/i', $raw)`, so it matches both. The closed test,
`/connection refused|tcp_conn=.*refused/i`, never fires, because the wording is
"No reply", not "connection refused".

Measured against the live API, sending exactly the request shape in this branch:

    8.8.8.8:53   open       rawOutput: "PING dns.google (8.8.8.8) on port 53.
                            Reply from dns.google (8.8.8.8) on port 53:
                            tcp_conn=1 time=1.01 ms ..."
                            -> verdict 2 OPEN     correct

    8.8.8.8:81   filtered   rawOutput: "PING dns.google (8.8.8.8) on port 81.
                            No reply from dns.google (8.8.8.8) on port 81:
                            tcp_conn=1 ..."
                            -> verdict 2 OPEN     wrong

    1.1.1.1:81   filtered   rawOutput: "PING one.one.one.one (1.1.1.1) on port
                            81. No reply from one.one.one.one (1.1.1.1) on port
                            81: tcp_conn=1 ..."
                            -> verdict 2 OPEN     wrong

Every outcome is "open". Because Globalping is the final fallback, this fires
exactly when the other providers could not answer -- so a user whose port is
genuinely unreachable, on a day when the primary provider is down, is told the
port is open. That is worse than the "unknown" state they get today: it sends
them looking anywhere except at the actual problem, which is the one thing this
indicator exists to prevent.

The discriminator is "Reply from" versus "No reply from":

    open    /^Reply from .*tcp_conn=\d+/mi
    closed  /^No reply from/mi   (or the "100% packet loss" summary line)

Worth a test over both rawOutput shapes. The parsing is pure, so it needs no
network -- tests/php/YouGetSignalParseTest.php is the pattern.

---

Second: the shipped default and the code default disagree.

    conf.php     $failoverProvidersIPv4 = ["portchecker", "globalping"]
    action.php   fallback  array("globalping", "portchecker")

conf.php is user-editable and survives upgrades, so an existing install takes
the action.php order and gets the new provider first, while the description and
a fresh install say portchecker leads. Whichever order is intended, the two
should agree.

---

Third: worst-case latency.

With $checkPortTimeout = 15, a failing primary plus a failing secondary plus
Globalping's own 5-10s window and up to three 2s polls is roughly 40s for one
interface action, where today it is capped near 15. Worth bounding the chain, or
budgeting the timeout across the providers rather than per provider.

---

Smaller notes:

- api.globalping.io is rate limited to 250 requests/hour per source IP when
  unauthenticated (`x-ratelimit-limit: 250`). Each install spends its own
  budget, so it is worth a line in conf.php beside the provider list.

- The new files are space-indented in a tab-indented tree: 77 lines in
  providers/globalping.php, 15 in providers.php, 3 in conf.php.

---

What is right, for the record: the refactor keeps the Origin header and the
check_port_parse_yougetsignal() call intact, action.php still requires
parse.php so nothing dangles, the request shape sent to Globalping is valid
(202 with a measurement id), and the branch is level with master -- php -l
clean on every changed file, PHP suite rc=0 with 2132 assertions, PHPStan 0.

Not checked: IPv6 end to end, and the failover chain driven through action.php
rather than by calling the providers directly. The wrong verdict is
demonstrable at the provider level, so I stopped there.

chiphead2332:
Fixed the Globalping verdict parsing and pushed the changes.

- `Reply from ... tcp_conn=N` → OPEN
- `No reply from ... tcp_conn=N` → UNKNOWN
- explicit connection refusal → CLOSED
- added regression coverage for these cases
- made the provider chain use a shared total timeout budget
- aligned the IPv4 fallback defaults between `conf.php` and `action.php`

Also verified the corrected Globalping provider against known-open and filtered ports: 3/3 live checks passed.

One potentially interesting avenue for the future: Globalping has probe "pingers" distributed around the world, so selecting an appropriate probe/location could potentially reduce the latency of port checks substantially compared with an arbitrary probe.

PR #3262, opened 2026-09-06
theme: MaterialDesign-QNAPforumPL — MaterialDesign with SVG left-pane…

silasmariusz:
## MaterialDesign-QNAPforumPL theme

A MaterialDesign variant maintained for the ruTorrent builds shipped to QNAP NAS users at forum.qnap.net.pl (Rtorrent 16 package). It is MaterialDesign with the things a 5.3.x install still needed on top, all inside one self-contained theme directory:

- the left panel's view / state / label icons as SVG (`panel-label[icon=…]::part(icon)`), and the category-panel chevron as an inline SVG (upstream rotates the same image when a panel is closed)
- `category-panel` headings, the File Manager's navigation controls (`#flm-nav-refresh`, `#flm-navpath`) and the settings dialog's list (`--bs-list-group-bg`) on the dialog's dark background
- context menus with readable item colours — upstream's `--menu-color` reads as "disabled" on the dark ground — and a flat 1 px border instead of the 2 px ridge
- SVG glyphs for the tracklabels rows the package creates (no label, movie, tv, music, game, software, kids, adult, book, other); other labels keep the plugin's PNG

Tested on ruTorrent 5.3.14 with the bundled plugins (filemanager, tracklabels, ratiocolor, geoip, mobile) in Chrome and Safari, light and dark system themes. No change outside `plugins/theme/themes/MaterialDesign-QNAPforumPL/`.

Screenshots: left panel, context menu, File Manager and settings dialog are in the theme's `images/` preview set.


xirvik (changes requested):
Issues:

## 1. The sort icon is the pre-#3255 one, so columns sort with inverted arrows

`stable.css:9` sets `--header-asc-icon` to a base64 GIF that is byte-identical to the
icon MaterialDesign shipped *before* #3255 (sha256 `289c1d97c791`; MaterialDesign
carries `9ee38921db19` since that fix). `css/stable.css:68` draws that icon for
`td.asc` and rotates it 180 degrees for `td.desc`, so a down-pointing icon inverts
both directions.

Rendered from `css/stable.css` plus this theme's `stable.css` in headless Chrome, an
ascending column shows a down arrow and a descending column an up arrow. Nothing in
the theme overrides the rule or the rotation, so that is what a user gets.

Copying the current value from `plugins/theme/themes/MaterialDesign/stable.css` fixes
it in one line.

## 2. 21 of the 27 bundled font files are never referenced

`plugins.css` declares only the six Ubuntu 400 faces. The Roboto set (`KFOl…`,
`KFOm…`) is referenced by no CSS in the theme, and no rule asks for Roboto by family —
the only families used are Ubuntu, Arial, monospace and `inherit`. MaterialDesign
references all 27 of its font files, so this looks like the directory was copied while
the `@font-face` blocks that used half of it were not.

That is 209,952 bytes of a 1,078,525-byte theme, carried by every install. Nine of the
54 SVGs and four of the 55 images are also unreferenced, which matters less but is the
same tidy-up.

## 3. Seven editorconfig violations

```
init.js:6                              trailing whitespace
plugins.css:72                         trailing whitespace
stable.css:252-253, 257-263, 269-275   trailing whitespace
svg/_read_me.txt                       wrong line endings / no final newline
```

The `editorconfig` job is advisory today and master passes it, so these are the only
findings it reports on the branch.

PR #3263, opened 2026-09-06
Chore/per cs formatting

adenwuts:
Initialises a conservative php-cs-fixer configuration to enable consistent code-styling in the code-base, applies that styling, and adds the commit hash to .git-blame-ignore-revs so that git blames aren't completely annihilated.


adenwuts:
@xirvik, this obviously needs some more work from my end, but the bones are there for configuration and ignoring the formatting commits.

What's the rationale behind maintaining tests and compatibility for PHP 7.4? Its been end-of-life for almost 4 years. Happy to make another issue for that if you'd like.

edit: just opened #3264 to make this process less painful.

adenwuts:
I think I'm happy to let this sit as a draft until legacy PHP support is dropped - I'm not sure the complexity/benefit ratio is there while we're still supporting older PHP versions.

xirvik:
> I think I'm happy to let this sit as a draft until legacy PHP support is dropped - I'm not sure the complexity/benefit ratio is there while we're still supporting older PHP versions.

The problem with this approach is that .git-blame-ignore-revs is not automatically honored by git. You need blame.ignoreRevsFile explicitly enabled.

I was considering enabling it in CI so it's mandatory for new PRs at least, which would eventually lead to something decent with a real blame, but it would take quite a while for everything to be decent.

adenwuts:
@xirvik I'm not sure I see that much of an issue with the changed git blames, but that might just be a result of my own workflow. From my perspective it seems like one extra blame to jump through, though I can see how that might be annoying.

Happy to defer to you on this, both options seem reasonable to me.

PR #3274, opened 2026-09-13
Telegram Notification Plugin

drrako:
# Telegram Notification Plugin

Adds per-user Telegram Bot API notifications to ruTorrent.

## Setup

Create a bot with [BotFather](https://telegram.me/BotFather), then configure the bot token and user chat ID in the Telegram settings (private chat or channel chat). Save with **OK** before sending a test message.

## Changes

- Added `plugins/telegram/` with:
  - Server-side bot-token storage.
  - Asynchronous notifications.
  - Added, finished, and removed torrent events.
  - Configurable message templates and English/Russian translations.
  - Telegram test-message action.
- Added common template:

  ```text
  Torrent {STATE}: {TORRENT}
  {LINK}
  ```

- Added HTTP(S) comment links using ruTorrent’s encoded `d.get_custom2` comment.
- Added PHP tests for configuration, rendering, links, API errors, secrets, and background argument handling.

## Validation

- Tested and verified manually.


xirvik (changes requested):
Hey, thanks, this will be useful/popular!

I've asked Claude to review it. Below it what is said. All seems reasonable to me, so I stand by the agent this time :-) 

Once that's fixed and Claude is happy, I'll do a human review and actually run it on my my test seedbox.

Claude
----------
The design is sound and the security thinking is better than most plugin PRs, but
as submitted the plugin never sends a notification on either rtorrent version it
claims to support, and it writes the bot token world-readable.

## 1. Blocking: the event handlers are inert

notificationCommand() passes torrent data as "$d.name=" / "$d.get_hash=" /
"$d.get_custom2=" inside the execute.nothrow={...} argument list. Stored as an
event key over XMLRPC -- which is how ruTorrent registers it -- that substitution
form does nothing, and execute.nothrow swallows the error.

I drove real rtorrents with the plugin's own generated string (interpreter
swapped for a probe that dumps argv), with a second handler on the same event as
a control:

    rtorrent 0.16.20
      plugin as-is (quoted $)     downloads=1 control_fired=True argv=''
      substitution unquoted       downloads=1 control_fired=True argv=''
      (cat,(d.name)) form         downloads=1 control_fired=True
                                  argv='[added] [a";touch PWN_SEMI;$(touch PWN_SUBST)`touch PWN_TICK`b] [BE7430BD...] [] []'

    rtorrent 0.9.8   -- identical results

The control fires every time, so the event itself is reached and the download is
inserted; the plugin's command produces no execute at all (empty log.execute).
The same holds for the simpler shape without the sh -c wrapper, so it is the
substitution, not the quoting: execute.nothrow={sh,probe.sh,$d.get_hash=,x}
produces nothing, while {sh,probe.sh,(cat,(d.hash)),x} produces
'[BE7430BD...] [x]'.

Fix is the (cat,(...)) form, through the existing command map so it keeps working
across versions:

    '(cat,('.getCmd('d.get_name').'))'   // d.name
    '(cat,('.getCmd('d.get_hash').'))'   // d.hash
    '(cat,('.getCmd('d.get_custom2').'))'

Worth saying what rtorrent the manual testing was done on -- if it delivered
there, that version accepts a form these two do not.

The good news from the same experiment: the argv design does exactly what the
comment claims. With the working substitution form, a torrent named
a";touch PWN_SEMI;$(touch PWN_SUBST)`touch PWN_TICK`b arrives as one argv
element, byte-intact, and none of the three injection markers was ever created.
Keep that structure.

## 2. The bot token is stored world-readable

rCache::set() ends with @chmod($name, $profileMask & 0666), and $profileMask
defaults to 0777 (conf/config.php:97). Demonstrated on this branch:

    file: share/settings/telegram.dat
    666 share/settings/telegram.dat        # and the token is in it, in plaintext

On a shared host every local account can read a credential that controls the bot.
A @chmod(..., 0600) after store() costs nothing. (Pre-existing files like
loginmgr.dat have the same problem; that is an argument for not adding another,
not for matching it.)

## 3. The CSRF comment promises more than the default delivers

action.php says the origin check happens before it is reached.
Requests::makeCSRFCheck() is indeed called from php/util.php -- but it is a no-op
unless $enableCSRFCheck is on, and that ships false. So on a default install a
cross-site POST can set the token, chat ID and events; the consequence here is
new in kind, because the attacker's own bot and chat then receive every torrent
name and comment link until someone notices. Worth rewording the comment to say
the check is opt-in, and telling users to enable it in the plugin's setup notes.

## 4. Three of the tests cannot fail

Mutation run against the new suite -- two caught, three survived:

    export the token in get()                                 caught
    drop the token redaction in sanitizeError()               caught
    while(false) instead of the UTF-8 trim loop               survived
    let an empty submitted token overwrite the stored one     survived
    accept any comment as a link (drop the https?:// guard)   survived

- testMessageLimitDoesNotSplitUtf8 uses str_repeat('я', 5000); 'я' is two bytes
  and 4096 is even, so the cut lands on a boundary and the loop never runs.
  Prefix one ASCII byte -- 'a'.str_repeat('я', 5000) -- and it bites
  (truncateMessage() then returns 4095 valid bytes; the raw substr() is invalid
  UTF-8).
- testAnEmptyTokenPreservesTheExistingSecret passes no token key at all, so
  isset() short-circuits and the documented "empty token keeps the secret" branch
  is never exercised. Add 'token' => ''.
- Nothing covers the link guard. VRS24mrkerjavascript%3Aalert(1) correctly
  renders as "Torrent finished: N" with no link today -- worth pinning.

## 5. Smaller things

- parse_mode => 'Markdown' is Telegram's legacy mode, where backslash escaping is
  not part of the format. A name that still trips the parser returns HTTP 400 and
  the notification is dropped with only an error_log line. MarkdownV2 with full
  escaping, or no parse_mode at all, fails less quietly.
- markdownLink() escapes \ and ) but not (: https://ex.org/a(b)c renders as
  [link](https://ex.org/a(b\)c). Percent-encoding the parens avoids the question.
- getCmd('d.name') is the post-rename name, while its neighbours use the
  pre-rename d.get_hash / d.get_custom2 that the version map understands. Use the
  mapped spelling for all three.
- No plugin.help. 47 of the 51 plugins in the tree have one; a wiki page plus that
  line would match.

## Gates

On the PR head: tests/php-test.sh 2264 assertions, rc=0 on PHP 8.4 and on
php:7.4-cli; PHPStan [OK] No errors; eslint 9 and editorconfig-checker clean;
Jest unaffected (no JS in the suite's scope changed).

Not checked: actual delivery to Telegram (no bot token), the settings page in a
browser, and FreeBSD. The findings above come from running the code, not from
reading it.

drrako:
Hi @xirvik, thanks for the feedback!

I let my agent address your agent feedback - oh, this new strange world 😅

Also, I’m curious about the

> 1. Blocking: the event handlers are inert

point. It sounds like it wasn’t supposed to work at all, although it did work for me on rTorrent 0.16.22?

